PR #150, opened 2026-04-12
fix(web): preserve session selection when slug changes

mgabor3141:
## Problem

When a session's slug changed (from `/new` creating a new session file, or the first user message arriving after attribution), the URL still contained the old slug. The view resolver couldn't match the old slug to the session, so `resolveViewFromPath` fell back to the project hub view, deselecting the session.

## Fix

In `upsertSession`, detect when the currently-selected session's slug changes and use `batch()` to atomically update both `sessions` and `urlPath`. This ensures the `view` computed never sees an inconsistent state (new session data + old URL).

`navigate()` is called after the batch to sync the browser URL bar via preact-iso.

## Tests

- **slug change on selected session rewrites URL atomically**: verifies `urlPath` updates and `selectedId` stays the same
- **slug change on non-selected session does not affect URL**: verifies no side effects on other sessions
- All 262 frontend tests pass


mgabor3141:
@greptile

PR #158, opened 2026-04-17
feat(cli): manage sessions with --no-attach, --list, --attach, --tail, --kill

mgabor3141:
`gmux` gains session management flags without becoming a subcommand-driven tool. The identity stays "one verb: run" — flags before the command apply to gmux itself; everything after the first positional is the command, verbatim. `--` handles commands that start with a dash.

## New shapes

```
gmux [--no-attach] <cmd> [args]   run a command in a new session
gmux --list                       list known sessions
gmux --attach <id>                reattach to an existing session
gmux --tail <N> <id>              print the last N lines of a session
gmux --kill <id>                  terminate a session
```

Short forms: `-l`, `-a`, `-t`, `-k`. A single `--session`-style selector isn't needed because each action takes its id positionally, same as `screen -r name` / `ssh host`.

## Shape decisions

- **No subcommands.** Adding `gmux tail` would collide with `gmux tail file.log`. Runner-plus-flags is the POSIX convention shared by `env`, `sudo`, `nohup`, `time`, `screen`, `ssh`, `systemd-run` — screen in particular solves the same "run + manage" problem gmux has, using only flags.
- **Management mode vs run mode is implicit.** If any action flag is set, gmux is in management mode and refuses a command. Otherwise, the first positional starts the command and flag parsing stops (stdlib `flag` does this by default). Mutually exclusive action flags are rejected at parse time.
- **`--attach`, `--kill` route through gmuxd** so remote peer sessions work transparently; gmuxd proxies the WS to the owning node. Verified that gmuxd serves `/ws/{id}` on its Unix socket with no auth wrapper (TCP listener is auth-wrapped, Unix socket uses the bare mux).
- **`--tail` talks directly to the session's Unix socket** via a new `GET /scrollback/tail?n=N` endpoint on the ptyserver. Plain-text output (ANSI stripped), scrollback + visible screen, trailing blank rows trimmed. Remote peers are rejected with a clear, case-specific error.
- **`--no-attach` reuses the existing "nested gmux" detached-spawn path**, so a single code path handles both "we're inside another gmux" and "user asked for detach".

## Session references

Any of the following resolves a session, matching what a user would plausibly type:

- The full ID (`sess-abcd1234`) or full slug (`fix-auth`).
- **The short form shown by `--list`** (`abcd1234`) — this was the nonobvious one; without it, copying the first column of `--list` into `--kill` would mysteriously fail to match.
- A unique prefix of any of the above.

Exact matches always beat prefix matches so an unambiguous short id can't be clobbered by a longer one that happens to start with the same characters. Ambiguous prefixes refuse to guess and list the candidates.

Dead sessions are pre-checked before `--attach` so the user sees "session X is not running (open the UI to resume)" instead of a cryptic WebSocket backend-unavailable error.

## Structure

`cli/gmux/cmd/gmux/main.go` used to be one 750-line file holding the entry point, gmuxd plumbing, browser detection, and the session runner. This PR breaks it apart:

- `main.go` — entry point + dispatch + UI mode
- `cli.go` — flag parsing and mode selection (pure function, fully tested)
- `run.go` — runSession + spawnDetached
- `actions.go` — list / kill / tail + session-reference matching
- `attach.go` — WS attach client wired to the existing `localterm`
- `daemon.go` — ensureGmuxd, register/deregister, health parsing
- `browser.go` — Chromium detection + tailnet URL masking

## Tests

- `cli_test.go` — dispatch table covering every valid shape and every user-facing validation (mutually exclusive flags, missing/extra positionals, `--no-attach` without a command, etc.). Asserts modes and flag values, not error strings.
- `actions_test.go` — `matchSession` covers the six shapes a user might type (full id, short form, slug, three kinds of prefix), plus the adversarial cases: ambiguous prefix, exact-beats-prefix, empty ref, empty session list, no match. `shortID` has a small idempotency/defensive table.
- `ptyserver_test.go::TestScrollbackTailEndpoint` — produces 20 colored lines in a 10-row viewport, requests `n=5`, asserts the last 5 arrive in order with no ANSI bytes.

## Docs

`reference/cli.md` updated with the new flags, examples, and the "flags before, command after" rule.

## Compatibility

The one behavior change: gmux now rejects unknown flags before the first positional. In practice this only bites invocations like `gmux --foo bar`, which previously tried to exec a file literally named `--foo` (always failing). Real use (`gmux pytest --watch`, `gmux ./my-script --flag`) is unchanged because the first non-flag positional ends gmux's flag parsing. `--` remains available for the rare case where the command itself starts with a dash.

### DIFF
--- a/apps/website/src/content/docs/reference/cli.md
+++ b/apps/website/src/content/docs/reference/cli.md
@@ -11,11 +11,13 @@ sidebar:
 
 The session runner and primary entry point. Auto-starts gmuxd if needed.
 
+`gmux` has no subcommands. Flags before the command apply to gmux itself; once the first positional argument is seen, everything after it is the command to run, verbatim, including its own flags. Use `--` to disambiguate when the command starts with a dash.
+
 ### `gmux`
 
 Open the gmux UI in a browser. Starts gmuxd if it is not already running. Prefers Chrome/Chromium in `--app` mode for a standalone window; falls back to the default browser.
 
-### `gmux <command> [args...]`
+### `gmux [--no-attach] <command> [args...]`
 
 Run a command inside a gmux session. The session registers with gmuxd and appears in the web UI.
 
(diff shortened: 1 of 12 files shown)


mgabor3141:
@greptile

PR #160, opened 2026-04-17
feat(cli): send input to a running session with --send

mgabor3141:
Adds `gmux --send <id> [text]` for injecting keystrokes/text into a running session's PTY. Useful for scripted interaction with agent sessions ("type this prompt", "press enter", "send Ctrl-C") from outside the web UI.

## Shape

```bash
gmux --send a3f20187 $'describe yourself\n'    # inline, shell handles escapes
echo 'describe yourself' | gmux --send a3f2    # stdin mode
printf '\x03' | gmux --send a3f2               # send Ctrl-C
```

- Inline text is sent verbatim, no auto-newline (callers include `\n` explicitly via `$'...\n'` or a pipe).
- Without an inline arg, gmux reads stdin until EOF (capped at 1 MiB).
- Session references follow the same rules as other management flags: full ID, short form shown by `--list`, full slug, or a unique prefix of any of these.

## Security model

`--send` is powerful — bytes land in the session's PTY indistinguishable from real keyboard input. Access is delegated to the session socket's filesystem permissions (`0700`, owner-only, created via `syscall.Umask(0o077)` when the runner starts). That means:

- Only the user who started the session can `connect()` to the socket. Other local users get `EACCES` from the kernel.
- The server-side handler has **no auth wrapper** — it can't — and that's fine: if you can reach the socket you already own the child process and could do worse things directly.
- Remote peer sessions are rejected outright. Cross-machine sending would need an explicit per-peer authorization model that doesn't exist yet; silently proxying through gmuxd would be genuinely dangerous. Clear error: "is only supported for local sessions (X is on peer Y)".

The 1 MiB server-side `LimitReader` on `/input` is not a trust boundary, just ergonomics — a bounded-size failure mode if someone pipes a huge file through by accident.

## Implementation

- **`POST /input` on the session's Unix socket.** New handler on the ptyserver: read up to 1 MiB from the body, write it straight to the PTY master, return 204. Same socket and access-control regime as the existing `POST /kill`, `PUT /status`, and `GET /scrollback/*` endpoints.
- **CLI flag + dispatch.** `--send` joins the management action set (mutually exclusive with `--list/--attach/--tail/--kill`). Accepts 1 or 2 positionals.
- **Client timeout disabled.** The shared `sessionSocketClient` has a 5s timeout for quick GETs (`--tail`); `--send` explicitly clears it because stdin can be legitimately slow.

## Tests

- `TestInputEndpoint` — end-to-end round-trip: child runs `read line; echo got:$line`, `POST /input` with body `hello\n`, then connect via WS and assert `got:hello` appears in the output. Demonstrates that the endpoint (a) reaches the child's stdin, (b) is accessible without any auth layer (which would break local callers), (c) integrates cleanly with the existing WS attach path used by the UI.
- `TestInputEndpointEmpty` — posting an empty body returns 204, not 500. Matters because a user piping an empty file into `gmux --send` should be a silent no-op.
- `TestParseCLI` / `TestParseCLIErrors` — two new happy-path cases (inline text, stdin) and four new validation cases (missing id, too many args, mutual-exclusion with `--kill`, mutual-exclusion with `--no-attach`).

## Docs

`reference/cli.md` gets a new section explaining the shapes, three usage examples (inline, pipe, Ctrl-C via `printf`), and a short "Access control" paragraph that makes the local-only scope explicit instead of leaving it as a surprise.

### DIFF
--- a/apps/website/src/content/docs/reference/cli.md
+++ b/apps/website/src/content/docs/reference/cli.md
@@ -77,6 +77,20 @@ Terminate a running session. Sends the same signal chain the UI's kill button do
 gmux --kill a3f20187
 ```
 
+### `gmux --send <id> [text]`
+
+Inject input into a running session, as if the bytes had been typed at the terminal. When `text` is given inline it is sent verbatim — no trailing newline is added, so you use shell `$'...\n'` (or pipe stdin) to submit:
+
+```bash
+gmux --send a3f20187 $'describe yourself\n'
+echo 'describe yourself' | gmux --send a3f20187   # equivalent
+printf '\x03' | gmux --send a3f20187              # send Ctrl-C
+```
+
+When `text` is omitted, gmux reads from stdin until EOF and sends whatever it sees (capped at 1 MiB). Stdin mode is the natural shape for piping, and avoids shell-escaping headaches for multi-line input.
+
+**Access control.** `--send` is powerful — anything you send lands in the session's PTY, and the child has no way to distinguish it from keyboard input. Access is gated by filesystem permissions on the session's Unix socket (owner-only, `0700`), which means only the user that started the session can send to it. Other users on the same machine cannot connect to the socket at all. For the same reason, `--send` is local-only: cross-machine sending would need an explicit authorization model that doesn't exist yet.
+
 ## gmuxd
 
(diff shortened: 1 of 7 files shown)


mgabor3141:
@greptile

PR #167, opened 2026-04-24
feat(web): improve mobile toolbar with dedicated arrow keys and shift

### LINKED ISSUE #163: Make mobile button row more intuitive
Surprising behavior, like paste behind ctrl (#156) cause friction.

There are also omissions like shift tab not being pressable.

Proposal: Have common features on the default row, add less common buttons to an extra openable menu, making the keyboard feature complete.

jinbe:
## Summary
- Add always-visible **up/down arrow** buttons (no longer hidden behind ctrl/alt toggle)
- Add **shift** modifier button for uppercase letters and shift+key combos
- **Pin** hamburger menu (left) and send button (right); make middle buttons **horizontally scrollable** for small screens
- Move send/paste button outside the scrollable area so it's always reachable

Closes #163

## Test plan
- [ ] Open web UI on a phone or with touch device emulation
- [ ] Verify all buttons visible: esc, tab, ctrl, alt, shift, ↑, ↓, ←, →
- [ ] Scroll the middle button row left/right on a narrow screen
- [ ] Tap shift, then type a letter — verify uppercase is sent
- [ ] Verify hamburger and send buttons stay pinned at edges
- [ ] Verify ctrl+paste and hold-to-repeat on arrows still work

### DIFF
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -193,30 +193,35 @@ function MobileTerminalBar({
   ctrlArmed,
   altArmed,
   shiftArmed,
+  metaArmed,
   onMenu,
   onSend,
   onPaste,
   onToggleCtrl,
   onToggleAlt,
   onToggleShift,
+  onToggleMeta,
   onFocusTerminal,
 }: {
   canSend: boolean
   ctrlArmed: boolean
   altArmed: boolean
   shiftArmed: boolean
+  metaArmed: boolean
   onMenu: () => void
   onSend: (data: string) => void
   onPaste: () => void
   onToggleCtrl: () => void
   onToggleAlt: () => void
   onToggleShift: () => void
+  onToggleMeta: () => void
   onFocusTerminal: () => void
 }) {
   // Read signals directly; no props needed for these.
   const bgActivity: DotState = backgroundActivity.value
   const unread = unreadCount.value
   const arrival = useArrivalPulse(bgActivity, unread)
+  const hostOs = health.value?.os
 
   const keepFocus = (ev: Event) => ev.preventDefault()
   const tap = (seq: string) => { onSend(seq); onFocusTerminal() }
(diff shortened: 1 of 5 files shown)


mgabor3141:
Create the meta change as a stacked PR please. I don't get how or why we would want to interact with the host OS.

mgabor3141:
Regarding shift: as far as I can tell, Shift Tab is the only key that is not served by the soft keyboard's built in shift.

We map the keyboard native enter to "shift enter" and I can't think of any other key that might make sense to be shifted that can't already be shifted from the soft keyboard normally.

mgabor3141:
To be honest I'm not sold on the side-to-side scrolling. It is better for screen real estate, but scrolling means you lose the other half of the buttons when you scroll to the side. This means if your common buttons are on the left, you lose them until you scroll back to them. You also can't ever combo buttons that are farther than `width` away from each other.

I'm still leaning towards preferring a design that has common buttons on the row like right now, but adding an `^` toggle button that opens a drawer to less commonly used buttons.

The question then becomes what do you put into each group. Ideally this would be configurable, but I think some good defaults could be reasonable for everyone.

Example:
"Extras" row:
`alt | shift+tab | del | home | end | paste`?
Normal row, always visible:
`hamburger | ^extras | esc | tab | ctrl | arrow keys | send`

Issue is I don't think this fits on a normal phone screen, but I don't know what I could cut that's less frequently used. Probably up and down? Maybe esc?

This definitely needs more discussion before I'm happy with a design, let's discuss further. I'm happy to see the experiments though.

mgabor3141:
The idea is that "typing" (commands or to an agent) is where you want to minimize needing to reach for an extras row the least, so an argument could be made for a default row like this:

`ham | tab | word left | left | right | word right | send`

everything else would be infrequent in comparison, so could go in the extras row.

Also note that if the extras row works as an overlay it is likely only covering the TUI footer...

jinbe:
So I borrowed the scrolling button area idea from the RVNC app on iPhone. That's been working pretty well for me and is intuitive now.

jinbe:
Perhaps if it's a combo thing then you could argue for a dedicated shift + tab, copy, and paste button.

Copy is a bit useless right now because you can't select on iOS.

Simply adding the shift and command buttons allow flexibility.

I did the OS detection thing because it seemed to use the same key combos as the host OS, if gmux was running on Windows then it wouldn't need the cmd key.

mgabor3141:
> scrolling button area idea

I've seen this in terminal emulator apps too, not a fan.

> Copy is a bit useless right now

Completely agree, it is high priority for a rework

Closing this as I'd like to implement this myself. Thank you for your input and sharing your use case!

PR #168, opened 2026-04-24
fix(web): enable long-press paste on iOS

### LINKED ISSUE #165: Can paste images on iPhone via CTRL button + v, can't paste text

jinbe:
## Summary
- On iOS Safari, xterm.js hides its textarea offscreen, so long-press never shows the native paste callout
- Detect a 500ms touch-hold without movement and temporarily reposition the textarea under the user's finger
- iOS then presents its native paste menu; the paste event fires on the textarea and flows through the existing `attachPasteHandler`
- Textarea position resets after a 2s delay so the paste callout has time to fire

Closes #165

## Test plan
- [ ] On iPhone Safari, long-press on the terminal area — paste callout should appear after ~500ms
- [ ] Tap "Paste" — clipboard text should be sent to the terminal
- [ ] Normal taps and scrolling should not be affected
- [ ] Dragging during a touch should cancel the long-press (no callout)

### DIFF
--- a/apps/gmux-web/src/terminal.tsx
+++ b/apps/gmux-web/src/terminal.tsx
@@ -492,7 +492,19 @@ export function TerminalView({
       startScrollTop: 0,
     }
 
+    // Long-press paste support for iOS.
+    // xterm hides its textarea offscreen, so iOS can't show a paste callout.
+    // When a touch is held without moving for 500ms, we temporarily position
+    // the textarea under the finger so iOS presents its native paste menu.
+    let longPressTimer: ReturnType<typeof setTimeout> | null = null
+
+    const clearLongPress = () => {
+      if (longPressTimer !== null) { clearTimeout(longPressTimer); longPressTimer = null }
+    }
+
     const handleTouchStartCapture = (ev: TouchEvent) => {
+      clearLongPress()
+
       if (ev.touches.length !== 1 || isInteractiveTarget(ev.target)) {
         touchPanState.active = false
         touchPanState.moved = false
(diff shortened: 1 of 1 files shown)


mgabor3141:
This is a clever idea.

Rather than repositioning the textarea I think it would be simpler to call the paste action directly.

Or even better, maybe long press could bring up a separate menu with the paste button, and maybe the ability to copy (part of) the last agent message or user text?

jinbe:
oh that's a good idea. let me have a play

jinbe:
tried a few things. popup menu with "Paste" is trivial, but on iOS it triggers the native paste button. which does work - but it's not a nice solution.

trying to update my other PR with a CMD or meta (for linux) button. which will let the host do CMD + v for paste. works with a real keyboard in the simulator so hoping it works if mapped to a button.

jinbe:
Yeah running the button thing on my phone now. I like it.

mgabor3141:
> which will let the host do CMD + v for paste

How does it know what's on your phone clipboard in that case?

jinbe:
Somehow it just does. I can also paste screenshots with ctrl + v.

All been working great so far. I haven't had a chance to test what happens on Android though.

mgabor3141:
Closing this as the textarea repositioning will not be considered.

Also see my comment on #165. Current feature set allows sending ctrl + v or ctrl + shift + v to the host, if the host uses that to paste that's a feature, but not the intended long-term paste solution.

Gmux should send the device's clipboard to the host, both text and images.

PR #171, opened 2026-04-24
docs(cli): document piped / non-tty flow and scripting patterns

mgabor3141:
Two docs changes that close a gap the `gmux <cmd>` flag reference left behind: the piped / non-tty flow (the foundation of any scripted or agent-driven use of gmux) was never documented, and there was no single page that tied `--tail`, `--send`, `--list`, `--kill`, and `--no-attach` together into usable patterns.

## `reference/cli.md`: document all three flows explicitly

The `gmux [--no-attach] <command>` section previously described two modes (attach / `--no-attach`) and noted nested-gmux auto-detach as an aside. It did not describe what happens when stdin is not a terminal, which is the common case for scripts, CI, and agent harnesses.

Replaced the two-paragraph description with a three-row table covering the flows as the runtime actually picks them: by stdin-is-a-tty and by `GMUX` in the environment. Explicit about the bounded metadata output in the non-tty flow, the exit-code propagation contract (previously not stated anywhere), and why nested auto-detach exists without making scripts worry about it. Cross-links to the new integrations page for the narrative version.

## `integrations/scripts-and-agents.md`: the narrative page

A new page under Integrations (auto-generated into the sidebar via the existing `autogenerate` config; no `astro.config.mjs` change). Covers:

- `gmux <cmd> | tail` as the canonical "block, bound output, let the user watch live, propagate exit code" pattern, with a step-by-step of why each property falls out of the non-tty flow.
- Why that shape is specifically useful to agents (blocking, bounded context, reliable exit, human visibility).
- `--tail` for post-hoc output access.
- `--list` for discovery.
- `--send` for programmatic input.
- `--no-attach` for fire-and-forget contrasted with the piped blocking shape.
- `--kill` for cleanup.
- A worked build-and-report script tying the primitives together.
- Nested gmux behavior, with accurate semantics: the auto-detach triggers on stdin-is-a-tty, so agent harnesses (whose stdin is a pipe) naturally land in the blocking non-tty flow and do the right thing without any special handling.
- Links out to the agent-specific pages (pi, Codex, Claude Code) as dedicated adapters that layer on top of these primitives.

## Verification

- `npx astro build` succeeds; all internal cross-links resolve to real anchors (spot-checked `dist/reference/cli/index.html` for the heading anchors referenced from the new page).
- Nested-gmux claim double-checked against `run.go`: `if os.Getenv("GMUX") == "1" && localterm.IsInteractive()` gates on stdin-is-a-tty, so piped/agent callers in a nested context fall through to the non-tty flow as documented. Confirmed with an end-to-end test using `env -i GMUX=1 bash -c 'script -qfc "gmux echo hi | tail" ...'` (auto-detaches because stdin-in-script is a pty) versus the agent-harness shape where stdin is actually a pipe (blocks as expected).

No behavior changes; docs only.

### DIFF
--- a/apps/website/src/content/docs/integrations/scripts-and-agents.md
+++ b/apps/website/src/content/docs/integrations/scripts-and-agents.md
@@ -1,0 +1,125 @@
+---
+title: Scripts and agents
+description: Drive gmux from shell scripts, CI pipelines, and coding-agent harnesses.
+---
+
+gmux is built around long-running processes that someone wants to supervise loosely. That makes it a natural fit not just for interactive use but for anything that *launches* long-running processes on a user's behalf: shell scripts, CI pipelines, and coding agents (pi, Codex, Claude Code, your own harness).
+
+The primitives are already in the CLI; this page shows how they compose.
+
+## The piped run: `gmux <cmd> | tail`
+
+The single most useful pattern for scripts and agents:
+
+```bash
+gmux make build | tail -n 20
+```
+
+What happens, step by step:
+
+1. Because gmux's stdin is a pipe (not a tty), it takes the [non-tty flow](/reference/cli/#gmux---no-attach-command-args): no raw mode, no PTY passthrough to stdout.
+2. gmux prints a short metadata header (`session:`, `pid:`, `socket:`, `serving...`), then **blocks** until the child exits.
+3. The child's actual output goes to the gmux session: visible live in the web UI, persisted in scrollback, and available to `gmux --tail`.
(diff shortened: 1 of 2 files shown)


mgabor3141:
Superseded by #211, which expands the original docs scope to include the CLI ergonomics fixes (--send submitting by default + --no-submit opt-out, --wait primitive) that the new docs would otherwise have to caveat. Closing as superseded; the docs work originally proposed here lands in #211.

PR #172, opened 2026-04-24
feat(cli): printable session id for --no-attach, post-exit --tail, and race-free handoff

mgabor3141:
Three small fixes driven by trying to use `gmux` as an agent tool. Each commit is atomic and TDD-covered.

### 1. `--no-attach` and nested-gmux auto-background print the new session id on stdout

Today both detaching paths only print "started ... in background" to stderr, so the caller has to `--list`-and-guess to find its own session. After this PR the short id goes to stdout, the human confirmation stays on stderr:

```bash
id=$(gmux --no-attach pytest --watch)
gmux --tail 100 "$id"
```

The parent pre-generates the id, passes it to the detached child through a new internal env var (`GMUX_RUNNER_FORCE_SESSION_ID`), and prints the short form on stdout. The child clears the env var after adoption so a grandchild can't silently inherit and collide.

### 2. `gmux --tail` now works on dead sessions

`ptyserver` flushes the final scrollback (plain text, ANSI stripped, same shape as `/scrollback/tail`) to `<socket_path>.tail` with owner-only permissions when the child exits. `cmdTail` falls back to that file whenever the live socket refuses a connection.

Before this PR, once a session exited the socket lingered but the server was gone, so `gmux --tail` returned `dial unix: connection refused`. Our own `scripts-and-agents.md` recommends the pattern "block with `gmux <cmd> | tail`, peek with `--tail` when you need the real output" — which didn't actually work for the common case of a build or test that had already finished. Now it does.

A new `Server.TailPersisted()` channel signals completion of the write so tests (and future consumers) can observe it without polling.

### 3. Close the races that made the printed id flaky for fast commands

After (1), `id=$(gmux --no-attach echo hi); gmux --tail "$id"` still failed about half the time on fast-exiting commands because three concurrent lifecycle events had no ordering guarantee:

- The child's gmuxd registration is a fire-and-forget goroutine — for ultra-fast commands the runner's own exit wins the race, leaving gmuxd with no record.
- `waitChild` writes the tail file *after* `Done()` fires, so `os.Exit` in `runSession` could cut the disk write off mid-flight.
- The detaching parent announced the id before gmuxd had acked registration, so the caller could see an id that gmuxd didn't yet recognize.

This PR closes all three:

- `spawnDetached` polls gmuxd's `/v1/sessions` for the id it pre-generated (2 s cap) before printing to stdout.
- The runner blocks its own exit on the registration goroutine (3 s cap — matches `registerWithGmuxd`'s retry budget).
- The runner also blocks exit on `Server.TailPersisted()` (500 ms cap).

Tight `--no-attach → --tail` reliability on fast-exiting commands goes from ~50 % to ~96 % in local stress tests. The remaining 4 % is a pre-existing lag deeper in gmuxd's registration pipeline, orthogonal to this PR.

## TDD

Each behavior change lands with tests that cover real code behavior, not implementation details:

- `nextSessionID`: forced-env adoption, env clearing after read, malformed fallback.
- `announceDetached`: stream contract (stdout = id, stderr = message; empty message stays silent).
- `Server.persistTail`: content correctness after real child exit, owner-only file permissions.
- `readPersistedTail`: content, request-more-than-available, missing file (distinguishable via `os.IsNotExist`).
- `lastNLines`: exhaustive subcases — the tests caught a real off-by-one in my first draft on input without trailing newline, exactly the kind of thing these tests exist to catch.
- `waitForRegistration`: fast path (already-present), realistic workflow (appears mid-poll), timeout (doesn't hang).
- Existing tests continue to exercise the `TailPersisted` and register-wait paths via their channel signals.

- Cleanup of orphaned `.tail` files (probably on `gmux --dismiss` and/or a TTL prune in gmuxd).
- The remaining registration race lives in gmuxd's discovery pipeline; worth investigating separately.
- Parent-session linkage for nested sessions, so the UI can collapse the "duplicate-looking sibling" that auto-background produces when a human types `gmux pi` in a gmux shell.
- The in-flight `docs/scripting-and-agents` branch currently says `--tail` works on live sessions only; once this lands, that line needs updating.

### DIFF
--- a/apps/website/src/content/docs/reference/cli.md
+++ b/apps/website/src/content/docs/reference/cli.md
@@ -74,7 +74,9 @@ gmux --tail 100 a3f20187
 gmux -t 20 fix-auth-bug
 ```
 
-`--tail` currently only works for sessions owned by the local node; remote peer sessions are rejected with a clear error.
+`--tail` works on both live and exited sessions. Live sessions are read from the runner's in-memory scrollback; exited sessions fall back to a plain-text file the runner flushed next to the session socket on exit (`<socket>.tail`, owner-only). The fallback exists so a build or test run that has already finished is still inspectable without re-executing it.
+
+`--tail` only works for sessions owned by the local node; remote peer sessions are rejected with a clear error.
 
 ### `gmux --kill <id>` (`-k`)
 
--- a/apps/website/src/content/docs/reference/cli.md
+++ b/apps/website/src/content/docs/reference/cli.md
@@ -33,6 +33,13 @@ With `--no-attach` the session is spawned in the background and appears in the U
 
 When run inside an existing gmux session (detected via the `GMUX` environment variable), `gmux` automatically detaches into a headless background process instead of nesting PTY-within-PTY. The new session appears in the UI.
 
+Both detaching paths (explicit `--no-attach` and the nested-gmux auto-background) print the new session's short ID to **stdout** and the "started ... in background" confirmation to stderr. Scripts can capture just the ID:
+
+```bash
+id=$(gmux --no-attach pytest --watch)
+gmux --tail 100 "$id"
+```
+
 ### `gmux --list` (`-l`)
 
 List known sessions, alive first, newest first.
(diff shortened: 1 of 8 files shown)


mgabor3141:
Superseded by #NEW. Closing in favor of a smaller, focused PR (#TBD) that drops the .tail file approach in favor of a state-dir lifecycle plan; see LIFECYCLE_PLAN.md (local scratchpad) for the multi-step decomposition. The new approach replaces polling with a pipe(2) parent/child handshake and removes the persistTail/.tail file machinery, deferring post-exit --tail to a follow-up that brokers scrollback through gmuxd.

mgabor3141:
Reopening reference: superseded by #184, which lands the smaller scriptable-id slice (pipe handshake instead of polling). Post-exit `--tail` and the rest of the lifecycle work moves into follow-up PRs per the worktree's lifecycle plan.

PR #173, opened 2026-04-25
perf(daemon): cache session file stats so unchanged conversations aren't re-parsed every scan

mgabor3141:
## Why

User reported steady-state ~90% CPU in gmuxd 1.2.0 with **0 alive sessions** and 32 known session dirs. The hot path is `convIndex.Scan()`, which runs every 30s in `gmuxd/cmd/gmuxd/main.go` and calls each adapter's `ParseSessionFile` (a full `os.ReadFile` + per-line `json.Unmarshal`) on every `.jsonl` under each adapter's session root.

Pi sessions accumulate: a single conversation file can grow to tens of MB (the source already calls this out in `extractPiText`). On my own dev machine (~380 MB across 490 pi/claude/codex files) a single `Scan()` already takes **4.6 seconds**. On a user with more or larger files, scans easily run longer than the 30s ticker interval, leaving the goroutine pinned to a core.

## What

Add a `(mtime, size)` snapshot per file path, taken before `ParseSessionFile`. If the snapshot matches the previous indexed value, skip the parse entirely. The snapshot is recorded for **every** file we visit (even ones we skip due to parse errors, missing cwd, or non-resumable content) so we don't re-pay parse cost on every interval just to drop the file again. Any future write bumps mtime or size, which invalidates the cache and re-parses.

## Numbers

Same dataset (490 files, ~380 MB), four sequential scans:

```
Scan #1: 4.63s, indexed 490 files
Scan #2: 2.76ms
Scan #3: 2.80ms
Scan #4: 2.56ms
```

~1700x speedup on the steady-state path. The first scan still pays full cost (correct: cold cache).

## State discipline

Per AGENTS.md, the new `fileStat` map is justified: it can't be derived from `byKey`/`byToolID` because we need a path-keyed lookup before parsing, and the avoided work (full reads of multi-MB JSONL) is exactly what was burning CPU. `Remove()` drops the corresponding cache entry so a recreated file at the same path is re-parsed.

## Testing

`TestScan_SkipsUnchangedFiles` verifies the behavior end-to-end: index a real pi session file, stamp a sentinel via `Upsert`, re-`Scan`, and assert the sentinel survives (cache hit). Then `os.Chtimes` to bump mtime and assert the next `Scan` overwrites the sentinel with the on-disk title (cache miss). I also confirmed the test fails when the cache check is forcibly disabled.

### DIFF
--- a/services/gmuxd/internal/conversations/index.go
+++ b/services/gmuxd/internal/conversations/index.go
@@ -43,6 +43,21 @@ type Index struct {
 	// byToolID maps "kind/toolID" → slug for reverse lookup
 	// (e.g., finding a conversation's slug from its tool UUID).
 	byToolID map[string]string
+	// fileStat caches (mtime, size) per file path so Scan() can skip
+	// re-parsing session files that haven't changed since the last
+	// scan. Pi/Claude/Codex session JSONL files can grow to tens of
+	// MB; without this cache, periodic rescans read and re-parse the
+	// entire session corpus every interval.
+	fileStat map[string]fileStat
+}
+
+// fileStat snapshots a file's mtime and size at index time. Two
+// snapshots compare equal iff the file is byte-identical to the last
+// indexed version (modulo mtime/size collisions, which are negligible
+// for append-only JSONL files).
+type fileStat struct {
+	modTimeUnixNano int64
+	size            int64
 }
 
 // New creates an empty index.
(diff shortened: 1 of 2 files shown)


mgabor3141:
Closing in favor of a watcher-only architecture. The cache hides the smell rather than addressing it: with a working watcher we don't need a periodic scan at all, just a startup bootstrap plus event-driven updates. Tracking that work separately.

PR #176, opened 2026-04-25
fix(web): keep scroll position across pi/TUI scrollback wipes

mgabor3141:
## Why

When pi (or any agent) finishes a turn it emits the canonical
"clear visible + cursor home + clear scrollback" sequence inside
a `BSU…ESU` block:

```
\x1b[?2026h\x1b[2J\x1b[H\x1b[3J  …redraw…  \x1b[?2026l
```

The `\x1b[3J` resets xterm's `ybase`/`ydisp` to 0 mid-block, and
the existing BSU/ESU restore logic then asks `scrollToLine(min(adjustedY=0, baseY=N))` for some `N>0`, parking the
viewport at the top.

User-visible: scroll up to read something, pi finishes, screen
yanks to the top. Reproduces deterministically with the captured
fixture in `e2e/fixtures/pi-turn.bin`.

## What's in this PR

Two atomic commits (rebase-merge → both land on `main` as
separate changelog bullets):

1. **`fix(web): restore distance from bottom when agent clears scrollback`** — detect a wipe (`baseY` shrinks across BSU/ESU)
   and restore the user's pre-wipe distance from the bottom
   instead of snapping to it.
2. **`fix(web): match line content as the strongest scroll-restore anchor`** — additionally capture the visible text of
   the line under the user at BSU time, scan the **whole** rebuilt
   buffer for that text, and restore there. Falls through to (1),
   then to bottom-snap.

The wipe branch in `maybeRestoreScroll`, end to end:

```ts
const anchorY = snap.prevAnchorLine !== null
  ? findAnchorMatch(scroll, snap.prevAnchorLine, baseY, rows, snap.prevDistanceFromBottom)
  : null
if (anchorY !== null)               scroll.scrollToLine(anchorY)         // (2)
else if (snap.prevDistanceFromBottom <= baseY)
                                    scroll.scrollToLine(baseY - snap.prevDistanceFromBottom)  // (1)
else                                scroll.scrollToBottom()              // last resort
```

`baseY < prevBaseY` is the wipe trigger: in steady-state
operation xterm's scrollback eviction leaves `baseY` pinned at
the cap, so this branch only fires on real `\x1b[3J` events.

## Search range: full buffer, not just scrollback

A match in the **visible region** (`y > baseY`) can't be the
viewport top (`scrollToLine` clamps to `baseY`), but
`scrollToBottom` keeps the line in view at offset `y - baseY`.
So `findAnchorMatch` searches `[0, baseY + rows)` and uses
`min(matchY, baseY)` as the restore target:

- `matchY ≤ baseY` (scrollback): anchor lands at viewport top,
  exactly where the user had it.
- `matchY > baseY` (visible): viewport snaps to bottom and the
  anchor sits inside the visible region, still readable.

Multiple matches collapse onto a single tiebreaker — closeness
of the resulting restore-distance to the captured pre-wipe
distance. Visible matches all collapse to the same target
(`baseY`, restore-distance 0); among scrollback matches the y
closest to `baseY - prevDistance` wins. The two regions compete
on equal terms: a small pre-wipe distance can prefer a visible
match over a faraway scrollback match.

## Pi specifically

Pi's wipe block is **13.4 kB containing 53 newlines** — a full
screen rewrite (version banner, keybind hints, the user prompt,
all 30 streamed fruit names, status bar). Outcomes by where the
user was scrolled:

- **At-bottom** (most common): unchanged, `wasAtBottom`
  short-circuits to `scrollToBottom()`.
- **Reading the version banner / keybind hints (large distance,
  anchor in post-wipe scrollback)**: anchor matches at low y, the
  user lands exactly there. The pi-fixture e2e test exercises
  this (pre-wipe distance 22, post-wipe `baseY` 15: distance
  alone could not have rescued).
- **Mid-distance, watching streaming output (anchor in post-wipe
  visible region)**: anchor matches in the visible region, we
  `scrollToBottom` and the line stays in view. Without the
  full-buffer search, this would fall through to distance
  restoration which scrolls the line off the viewport.
- **Tiny distance, top-of-viewport is blank** (eg pi's whitespace
  rows): trivial-anchor filter rejects the capture, falls
  through to distance restoration.

## Tests

**Unit** (`apps/gmux-web/src/terminal-io.test.ts`, +6 cases):

- Distance preserved when new `baseY` is large enough; snap to
  bottom when distance > new `baseY`; baseY-grows-from-zero
  doesn't trigger the wipe branch (commit 1).
- Single-anchor match (commit 2).
- Multi-anchor tiebreaker — winning match `y=16` deliberately
  differs from the distance fallback `y=15` so a regression that
  disables anchor matching is caught.
- **Visible-region match** — pre-wipe distance 5, post-wipe
  `baseY=5` with anchor at `y=20`. Mutation-sensitive: re-narrowing
  the search to `[0, baseY]` puts the user at `viewportY=0`
  instead of `viewportY=baseY`.
- **Mixed-region tiebreaker** — both regions match; pre-wipe
  distance 1 makes the visible match's restore-distance (0) win
  over the scrollback match's restore-distance (13).
- Null-anchor and missed-anchor both fall through to distance.

**E2E** (`e2e/tests/terminal-scroll.spec.ts`, +3 cases):

- Synthetic scrollback match: terminal pinned to 120x40 so the
  anchor-target y (=20) and distance-fallback y (=30) provably
  differ.
- **Synthetic visible-region match**: 45-line redraw on a
  rows=40 terminal, anchor at y=40 (visible). Asserts
  `viewportY === baseY` and walks the visible rows to confirm the
  anchor line is in the viewport.
- **Pi-fixture: user reading the version banner stays on it
  across the wipe** — replay chunks 0–97 of the real pi capture,
  scroll to the version banner (y=2, distance=22), replay the
  wipe + redraw chunks. Post-wipe `baseY=15`, distance fallback
  could not have rescued; anchor match lands the user at y=1
  with the same banner text.

The original `real pi end-of-turn does not jump to top` fixture
test continues to pass: when the user is deep in seeded
scrollback (no overlap with pi's redraw content), anchor matching
falls through and distance/bottom-snap takes over without
regressing the bug fix.

## Mutation tests

- Replace the wipe branch with `scrollToBottom()` ⇒ fixture-based
  e2e test from commit 1 fails (viewportY=baseY instead of
  preserving distance).
- Disable anchor matching (`const anchorY: number | null = null`)
  ⇒ 4 unit tests + 3 e2e tests fail (single-match, tiebreaker,
  visible-match, mixed-tiebreaker, synthetic scrollback e2e,
  visible-region e2e, pi-fixture e2e).
- Re-narrow search to `[0, baseY]` ⇒ 2 unit tests + 1 e2e test
  fail (visible-region unit, mixed tiebreaker unit, visible-region
  e2e). Other tests still pass since they exercise scrollback
  matches.

## Test plan

- [x] `pnpm exec vitest run` → 271/271
- [x] `pnpm test:e2e` → 31/31
- [x] `pnpm exec tsc --noEmit` clean
- [x] Fixture sanity: `\x1b[3J` is at offset 69408, inside BSU at
  69393 / ESU at 82833, redraw block has 53 newlines (full screen
  rewrite), not just status bar.
- [x] Mutation tests above.

## Limitations (deliberately not addressed)

- **Wrapping**. If a logical line wraps to a different number of
  buffer lines after the redraw, `translateToString` returns the
  partial buffer line and the match fails. We fall through to
  distance, no regression vs commit 1. Could be addressed with
  `isWrapped`-aware merging if real reports warrant it.
- **`\x1b[3J` outside BSU/ESU**. Detection currently lives only
  on the `maybeRestoreScroll` path; an unwrapped `\x1b[3J` would
  not trigger restoration. No real-world reproducer for this
  shape yet.

### DIFF
--- a/apps/gmux-web/src/terminal-io.test.ts
+++ b/apps/gmux-web/src/terminal-io.test.ts
@@ -392,27 +392,50 @@ describe('scroll preservation across BSU/ESU', () => {
     h.cleanup()
   })
 
-  it('snaps to bottom when scrollback is wiped (\\x1b[3J) inside BSU/ESU and user was scrolled up', () => {
-    // The pi end-of-turn shape that the e2e fixture exercises: user is
-    // reading earlier output, agent emits a BSU + clear-scrollback +
-    // redraw + ESU. Scrollback shrinks underneath the user; their
-    // pre-BSU line is gone, so the only meaningful restore is bottom.
+  it('after scrollback wipe, restores user\'s distance from bottom when new buffer is large enough', () => {
+    // The user was reading 3 lines above the latest content (eg the
+    // last line of pi's previous turn). Pi ends a turn with
+    // \x1b[3J + redraw, scrollback gets rebuilt to ~15 lines. The
+    // user's pre-BSU absolute line is gone, but their *intent* ("keep
+    // me 3 lines above the newest content") is preserved.
     const h = makeScrollHarness({ scrollbackLimit: 5000, rows: 40 })
     h.io.reset(1)
-    h.addLines(200)            // baseY=200, plenty of scrollback
-    h.userScrollTo(100)        // user reading the middle
-    expect(h.viewportY).toBe(100)
-    expect(h.baseY).toBe(200)
+    h.addLines(200)            // baseY=200
+    h.userScrollTo(197)        // 3 lines above the bottom
+    expect(h.baseY - h.viewportY).toBe(3)
 
     h.io.enqueue(wrapBSU('redraw'), 1)
-    // Inside the BSU/ESU: agent emits \x1b[3J then redraws ~15 lines.
     h.clearScrollback()
-    h.addLines(15)
-    h.flushOne(0)              // ESU write callback
-    h.flushRAF()               // restore rAF
+    h.addLines(15)             // new baseY=15, plenty of room for distance=3
+    h.flushOne(0)
+    h.flushRAF()
+
+    // Distance preserved: viewportY == baseY - 3.
+    expect(h.scrollToLineCalls).toContain(h.baseY - 3)
+    expect(h.scrollToBottomCalls.length).toBe(0)
+    expect(h.baseY - h.viewportY).toBe(3)
+    h.cleanup()
+  })
+
+  it('after scrollback wipe, snaps to bottom when distance exceeds new buffer', () => {
+    // The pi end-of-turn shape from the e2e fixture: user was reading
+    // far back in scrollback (100 lines above the bottom), the new
+    // buffer is only ~15 lines, so the user's intent has nowhere to
+    // anchor. Snap to bottom: nothing else is meaningful.
+    const h = makeScrollHarness({ scrollbackLimit: 5000, rows: 40 })
+    h.io.reset(1)
+    h.addLines(200)
+    h.userScrollTo(100)        // 100 lines above the bottom
+    expect(h.baseY - h.viewportY).toBe(100)
+
+    h.io.enqueue(wrapBSU('redraw'), 1)
+    h.clearScrollback()
+    h.addLines(15)             // new baseY=15, distance(100) > baseY
+    h.flushOne(0)
+    h.flushRAF()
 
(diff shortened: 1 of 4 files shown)


mgabor3141:
@greptile

PR #177, opened 2026-04-25
fix(web): match line content as the strongest scroll-restore anchor

mgabor3141:
## Why

[#176](https://github.com/gmuxapp/gmux/pull/176) preserves the
user's distance from the bottom across a scrollback wipe. That's
strictly better than snapping to the bottom, but it still throws
away one piece of state we have right at BSU time: the actual
text of the line under the user.

Follow-up suggestion from review: capture that line content as a
marker, and after the wipe, scan the rebuilt buffer for it. If we
find it, jump there directly. If multiple lines match (a common
"ok" or " ✓ Done"), tiebreak by closeness to the captured
distance from bottom.

## Stacked on top of #176

This PR sits on `fix/scrollback-preserve-distance` (#176), not
`main`. Once #176 lands, GitHub will rebase this onto `main`
automatically. Until then, review the diff against
`fix/scrollback-preserve-distance` to see only the anchor-matching
delta.

## Change

`ScrollAccessor` gains a `getLine(y)` method, implemented in the
production accessor by `term.buffer.active.getLine(y).translateToString(true)`.
The implementation filters trivial anchors (empty, whitespace-only,
< 4 visible chars) and returns `null` for them: separators, blank
lines, and short status markers match too many candidate lines to
be useful as anchors.

`savedScroll` gains `prevAnchorLine: string | null`, captured at
BSU time only when the user is scrolled up. The wipe branch tries
three anchors in order, falling through on each miss:

1. **Anchor match** — scan `[0, baseY]` for a line whose
   `translateToString` equals `prevAnchorLine`. Multiple matches
   are tiebroken by `|baseY - y - prevDistanceFromBottom|`. Search
   range is the scrollback area `scrollToLine` can target;
   visible-only lines aren't valid scroll targets so we skip them.
2. **Distance restoration** (already in #176).
3. **Bottom snap** (already in #176).

```ts
const anchorY = snap.prevAnchorLine !== null
  ? findAnchorMatch(scroll, snap.prevAnchorLine, baseY, snap.prevDistanceFromBottom)
  : null
if (anchorY !== null) {
  scroll.scrollToLine(anchorY)
} else if (snap.prevDistanceFromBottom <= baseY) {
  scroll.scrollToLine(baseY - snap.prevDistanceFromBottom)
} else {
  scroll.scrollToBottom()
}
```

## Who actually benefits

General TUIs that refresh their display with substantially the
same content: log viewers (`tail -f` reload), file browsers,
code editors with gutter redraws, package managers showing
identical install summaries.

**Not pi specifically**: pi's `\x1b[3J` redraw at end-of-turn
emits only the status bar, not the conversation content the user
was reading. That case still falls through to distance
restoration, exactly as in #176, and is covered by the
`user scrolled up: real pi end-of-turn does not jump to top`
fixture test.

## Tests

**Unit** (`terminal-io.test.ts`, +4 cases):

- single-anchor match: jumps to the rebuilt line
- multi-anchor tiebreaker: deliberately positioned so the
  winning match (y=16) differs from the distance fallback (y=15);
  this is what makes the test sensitive to the anchor branch
  being active rather than coincidentally matching the fallback
- null anchor: getLine returned null (trivial line) ⇒ falls
  through to distance restoration
- anchor-not-in-redraw: line under user not present in new
  buffer ⇒ falls through to distance restoration

The harness gains a `setLine(y, text)` method and a per-y content
map cleared on `clearScrollback()`.

**E2E** (`terminal-scroll.spec.ts`, +1 case): seed 200 lines,
scroll up 10, inject a single `BSU + \x1b[2J \x1b[H \x1b[3J + 80
redraw lines + ESU` frame containing the same seeded line at a
new position. Terminal is pinned to 120x40 so the anchor target
(y=20) and the distance fallback (y=30) provably differ; the
viewport must land on y=20 with content equal to the captured
anchor.

## Mutation tests

Disabled the anchor branch (`const anchorY: number | null = null`)
and confirmed:

- `terminal-io.test.ts`: 2 failures (single-match + tiebreaker)
- `terminal-scroll.spec.ts`: 1 failure (anchor-match e2e)

Restoring the branch turns all green.

## Test plan

- [x] `pnpm exec vitest run` → 269/269 (was 265, +4 anchor cases)
- [x] `pnpm test:e2e` → 29/29 (was 28, +1 anchor case)
- [x] `pnpm exec tsc --noEmit` clean
- [x] Mutation: disable anchor branch ⇒ 3 tests fail; restore ⇒ green

## Limitations (deliberately not addressed)

- **Wrapping**. If the same logical line wraps differently after
  the redraw (eg different terminal width), `translateToString`
  returns the partial buffer line and the match fails. We fall
  through to distance restoration, no regression vs #176. Could
  be addressed with `isWrapped`-aware merging if real reports
  warrant it.
- **Single-line marker**. A short anchor line might match a
  similar-but-not-identical line elsewhere. The trivial-anchor
  filter (4 visible chars) and the distance tiebreaker keep this
  fairly contained in practice.

### DIFF
--- a/apps/gmux-web/src/terminal-io.test.ts
+++ b/apps/gmux-web/src/terminal-io.test.ts
@@ -47,6 +47,10 @@ function makeScrollHarness(opts: { scrollbackLimit: number; rows: number }) {
 
   const scrollToLineCalls: number[] = []
   const scrollToBottomCalls: number[] = [] // tracks call count
+  // Per-line content for anchor-matching tests. y → visible text. Lines
+  // not in the map return null from getLine() (the production accessor
+  // also returns null for empty / trivial lines).
+  const lineContent = new Map<number, string>()
 
   const scroll: ScrollAccessor = {
     getState: () => ({ viewportY, baseY }),
@@ -57,6 +61,9 @@ function makeScrollHarness(opts: { scrollbackLimit: number; rows: number }) {
     scrollToBottom() {
       scrollToBottomCalls.push(baseY)
       viewportY = baseY
+    },
+    getLine(y: number): string | null {
+      return lineContent.get(y) ?? null
     },
   }
 
@@ -135,6 +142,12 @@ function makeScrollHarness(opts: { scrollbackLimit: number; rows: number }) {
       totalLines = rows
       baseY = 0
       viewportY = 0
+      lineContent.clear()
+    },
+
+    /** Set the visible text of the buffer line at `y`. */
+    setLine(y: number, text: string) {
+      lineContent.set(y, text)
     },
 
     /**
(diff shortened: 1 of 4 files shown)


mgabor3141:
Heads-up: the CI checks beyond the conventional-commit validators don't run here because the workflow only triggers on PRs targeting `main` (`.github/workflows/ci.yml` `pull_request: { branches: [main] }`), and this PR is stacked on `fix/scrollback-preserve-distance`.

Once #176 merges, this PR's base auto-updates to `main` and the full suite (Go, JS, E2E, etc.) will run. Locally on this commit:

- `pnpm exec vitest run` → 269/269
- `pnpm test:e2e` → 29/29
- `pnpm exec tsc --noEmit` clean
- mutation-tested (disable anchor branch ⇒ 3 tests fail)

mgabor3141:
Folded into #176 per review. Both commits now live there as a 2-commit stack with rebase-merge in mind.

PR #178, opened 2026-04-25
feat(release): improve release flow (discord roles, highlights SOT, preview, tests)

mgabor3141:
Two commits stacked on this PR.

## Commit 1: `fix(release): ping all relevant discord roles and suppress link embed`

- Each release now pings every role at or below its severity (patch always pings; feature adds feature; breaking adds feature + breaking).
- Discord message sets `flags: 4` (SUPPRESS_EMBEDS) so the changelog link no longer renders an autogenerated preview card.

## Commit 2: `feat(release): make highlights the source of truth, add preview comment, harden helpers`

Several improvements to the release flow.

### `RELEASE_HIGHLIGHTS.md` is now the single source of truth

Previously, `version.sh` cleared `RELEASE_HIGHLIGHTS.md` while generating the release PR, so once the PR was open the file on `release/next` was already empty. To tweak prose post-PR you had to edit *both* `RELEASE_NOTES.md` and `apps/website/src/content/docs/changelog.mdx` by hand, and they could drift.

Now:

- `version.sh` no longer clears `RELEASE_HIGHLIGHTS.md`. It stays populated on `release/next` until release time.
- A new `version.sh --clear-highlights` mode resets it to a stub.
- `release.yml` invokes that mode after a successful release and pushes a `chore: clear release highlights for vX.Y.Z` commit to `main`. Verified main's branch protection ruleset only blocks `deletion` and `non_fast_forward`, not direct fast-forward pushes.
- New `scripts/regen-release-notes.sh` helper: edit `RELEASE_HIGHLIGHTS.md` on `release/next`, run the script, force-push. It rolls back the auto-generated commit, restores your highlight edits on top of clean state, regenerates, and re-creates the release commit. Includes safeguards: refuses to run if not on `release/next`, if `HEAD~1` isn't `origin/main`, or if there are uncommitted changes outside `RELEASE_HIGHLIGHTS.md`.

### Sticky preview comment on the release PR

`version.yml` now posts (and updates) a single comment showing exactly what `notify-discord.sh` will send when the release ships. Reviewers can verify wording without expanding the PR body or running the script locally. Comment body is built via `jq --rawfile` and POSTed via `--input -` so arbitrary markdown content (quotes, backticks, control chars) round-trips correctly.

### Smarter Discord truncation

`notify-discord.sh` now cuts at the last paragraph break (then sentence break, then newline) under the 2000-char limit, instead of a hard byte cut mid-sentence. Adds an ellipsis (`…`) instead of `...`. The truncation logic is extracted into `trim_summary` and message composition into `compose_message`, removing duplication.

### `git describe` for previous tag

`notify-discord.sh` previously did `git tag -l | sort -V | grep -B1 ...` to find the prior tag, which silently broke if `VERSION` wasn't yet locally known or tags were out of order. Replaced with `git describe --tags --abbrev=0 "${VERSION}^"`, which always picks the actual predecessor on the release ancestry.

### Batched GraphQL PR lookup

`version.sh` previously made one REST call per commit to resolve PR links (`/commits/{sha}/pulls`). On larger releases that's dozens of round trips and silently no-ops without `GITHUB_TOKEN`. Now a single GraphQL query resolves every commit at once, and the script fails loudly when running in CI without auth so we never ship a changelog with bare commit subjects by accident. Falls back silently outside CI so iteration without `gh` keeps working.

### Testability and tests

`notify-discord.sh` gains a `DRY_RUN=1` env var that prints the JSON payload to stdout instead of POSTing. Useful for local debugging and powering a new test file `notify_discord_test.sh` covering:

- Patch / minor / major bumps ping the right roles (and only those).
- `flags: 4` is set so embeds are suppressed.
- Long summaries truncate at a paragraph break under the 2000-char limit, ellipsis is added, footer link survives.
- First-ever release (no predecessor tag) pings patch role only.

`version_test.sh` was updated: replaced the "highlights cleared after release" assertion with one verifying highlights survive PR generation, plus a new test for `--clear-highlights`.

All 36 tests pass locally (`version_test.sh`: 28, `notify_discord_test.sh`: 8). Tests were pruned from initial drafts to focus on behavior the scripts actually own; brittle string-matching against git-cliff templates and Discord rendered text was dropped.

### Docs

- `AGENTS.md` describes the new highlights flow and the `regen-release-notes.sh` workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -109,6 +109,31 @@ jobs:
           VERSION: ${{ steps.version.outputs.version }}
         run: bash .github/workflows/scripts/notify-discord.sh
 
+      # The release/next PR carries RELEASE_HIGHLIGHTS.md untouched so
+      # editors can tweak prose right up until merge. Now that the
+      # release has shipped, reset it to the stub on main so the next
+      # cycle starts clean.
+      - name: Clear release highlights on main
+        env:
+          GH_TOKEN: ${{ github.token }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          bash .github/workflows/scripts/version.sh --clear-highlights
+          if git diff --quiet RELEASE_HIGHLIGHTS.md; then
+            echo "RELEASE_HIGHLIGHTS.md already clean, nothing to do"
+            exit 0
+          fi
+          git config user.name 'github-actions[bot]'
+          git config user.email '41898282+github-actions[bot]@users.noreply.github.com'
+          git add RELEASE_HIGHLIGHTS.md
+          git commit -m "chore: clear release highlights for ${VERSION}"
+          # The checkout at the start of this job is stale by now
+          # (goreleaser, Discord notify, etc. take minutes). Rebase
+          # onto current main so the push fast-forwards even if other
+          # commits landed in the meantime.
+          git pull --rebase origin main
+          git push origin main
+
   deploy-docs:
     needs: release
     if: needs.release.result == 'success'
--- a/.github/workflows/scripts/notify-discord.sh
+++ b/.github/workflows/scripts/notify-discord.sh
@@ -7,6 +7,9 @@
 #   ROLE_BREAKING  - Discord role ID for breaking changes
 #   ROLE_FEATURE   - Discord role ID for feature releases
 #   ROLE_PATCH     - Discord role ID for patch releases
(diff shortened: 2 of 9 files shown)


mgabor3141:
@greptile

mgabor3141:
@greptile

mgabor3141:
@greptile

PR #179, opened 2026-04-25
feat(daemon): drop periodic conversations index scan in favor of watcher

mgabor3141:
## Why

User report: gmuxd 1.2.0 sits at ~90% CPU steady state with **zero alive sessions** and 32 session dirs. Tracked it down to `conversations.Index.Scan()` running every 30s and re-reading every adapter `.jsonl` file (pi files reach tens of MB). Once the corpus exceeds ~30s of parse work, the scan never finishes before the next tick, pinning a core indefinitely.

Previous attempt (#173) added a `(mtime, size)` cache. Closed in favor of removing the ticker entirely: with the watcher wired up, idle = zero events = zero CPU. The ticker was masking an architectural duplication where filemon already had fsnotify infrastructure but its wiring to the conversations index had been left dead (`Index.ScanFile` defined, never called).

## What

Five atomic commits:

1. **`test(e2e): add adapter session file helpers and smoke spec`** — `GMUX_TEST_HOME` export, `apiGet` / `pollUntil` / `writeFakeSession` helpers, three smoke tests pre-seeding fixtures so the bootstrap scan picks them up. Drift-detector for fixture-vs-parser compatibility.
2. **`feat(daemon): drop periodic conversations index scan in favor of watcher`** — always-on root watches, `rootToAdapter` map, `Index.ScanFile` wired to fsnotify Create/Write, `Index.RemoveByPath` (linear walk, no new map state) wired to Remove/Rename, catch-up scan in `handleNewSubdirLocked` to close the `mkdir+touch` race, recursive tree walk so codex's `YYYY/MM/DD` nesting is fully covered, 30s ticker deleted, six discovery tests.
3. **`docs(daemon): document watcher-only conversations index in agent-facing docs`** — `e2e/AGENTS.md` notes on the new helpers and specs, `RELEASE_HIGHLIGHTS.md` user-facing prose, comment polish in `filemon.go` calling out the `mkdir` side effect of `WatchRoots`, and `node:` import normalization in `e2e/fixtures.ts`.
4. **`test(e2e): use local time in codex fixture path`** — addresses Greptile review feedback. Codex fixture now uses local time to match Go's `codex.SessionDir` (`time.Now()`); functionally a no-op for tests (bootstrap walks the whole tree, watcher does too) but removes a misleading comment and stylistic divergence.
5. **`perf(daemon): release fm.mu before running catch-up ScanFile`** — addresses Greptile review feedback. `handleNewSubdirLocked` now collects deferred `indexWork` and the caller iterates `ScanFile` after releasing `fm.mu`. The walk still needs the lock (it modifies `watchedDirs`), but the per-file JSONL parse no longer blocks `NotifyNewSession` etc. during a deep catch-up.

## Verification

- `go test ./services/gmuxd/...` clean; `go vet ./...` clean.
- All 36 e2e tests pass (30 existing + 3 smoke + 6 discovery).
- **Regression-catch verified twice**: temporarily disabled `SetConvIndex` + `WatchRoots` on top of both the original rewrite (commit 2) and the post-feedback state (after commit 5). All 6 discovery tests fail in both cases (timeout exhaustion at 2s for tests 1-5, no event ever firing for test 6). Smoke tests still pass because they exercise the bootstrap path.
- Walked through every realistic concurrency scenario (bootstrap vs watcher, NotifyNewSession vs WatchRoots, mid-walk file creation in catch-up scan, codex date-rollover at midnight). Documented the side effect of `WatchRoots` creating empty adapter root directories at startup.

## Design decisions

| | |
|---|---|
| Freshness contract | Call `ScanFile` on every event, no throttle, no first-then-skip. Realistic per-event cost is bounded; idle is zero. |
| State discipline | One new field on `FileMonitor` (`rootToAdapter`, set once at construction). No `byPath` map: `RemoveByPath` does a linear walk over `byKey` (rare event, sub-ms even at 10K entries). |
| Recovery hooks | Skipped (Q_OVERFLOW reconcile, sleep-wake reconcile, `watcher_degraded` health field). Bootstrap scan on daemon restart is the documented recovery. Add later if reports emerge; logs already cover ENOSPC. |
| Lock discipline | Catch-up walks `watchedDirs` modifications under `fm.mu`, but defers per-file `ScanFile` to the caller post-unlock so a large catch-up doesn't block other acquirers. |
| Cross-platform | Linux is CI; macOS is dev-only. fsnotify behaves differently but the bootstrap scan covers both. Called out as a deferred risk. |

## Closes

- The user's 90% CPU report (idle daemon now does no periodic work).
- Incidentally: deleted-file leak in the index (Remove was never wired before; now is).

### DIFF
--- a/RELEASE_HIGHLIGHTS.md
+++ b/RELEASE_HIGHLIGHTS.md
@@ -18,3 +18,9 @@ Example:
     `projects.json` replaces separate `remote` and `paths` fields with a
     unified `match` array.
 -->
+
+The daemon no longer rescans every adapter session file every 30
+seconds. Conversation discovery is now driven directly by filesystem
+events, so an idle gmuxd does no periodic work. Heavy users with
+many large pi/claude/codex sessions will see CPU drop from steady
+background burn to effectively zero when nothing is happening.
--- a/e2e/AGENTS.md
+++ b/e2e/AGENTS.md
@@ -39,6 +39,33 @@ defense-in-depth for future tests that spawn agents like `pi` or
 `claude` (which would otherwise read the operator's real
 `~/.pi/agent/` etc.); audit it manually when touching this setup.
 
+## Adapter-session fixtures
+
+`fixtures.ts` knows how to write minimally valid JSONL session files
+for pi/claude/codex into `$GMUX_TEST_HOME` (set by global-setup). Two
+specs use it:
+
+- `tests/conversation-fixtures.spec.ts` is a drift-detector. It
+  pre-seeds one fixture per adapter via global-setup *before* gmuxd
+  starts, then asserts each is reachable through `/v1/conversations/`
+  once bootstrap completes. If a Go parser grows a new required
+  field and the TS fixture lags, this spec fails first with a clear
+  signal pointing at fixture validity, before anything else.
+- `tests/conversation-discovery.spec.ts` exercises the watcher path:
+  writes JSONL files at test time and asserts the index reflects
+  them within 2s. Tight timeout intentionally: watcher-driven
+  discovery is sub-second; longer would mask regression to a
+  periodic scan. The harness's only alive session is `shell`, so
+  these tests implicitly verify always-on root watches — if root
+  watches were ever re-gated on a live session of the same kind,
+  the discovery tests would time out.
+
+When adding a test that needs a session file on disk, prefer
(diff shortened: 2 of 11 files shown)


mgabor3141:
@greptile

mgabor3141:
@greptile

PR #180, opened 2026-04-25
fix(cli): default TERM to xterm-256color and report real version to children

### LINKED ISSUE #154: bug: browser-launched shells can miss TERM, breaking TUIs like lazygit
Repro:
- open a shell inside gmux
- run `lazygit`

Observed:
- lazygit exits with `terminal entry not found: term not set`

Notes:
- `gmux lazygit` works
- likely the browser/daemon-launched shell path can start without `TERM` in the child env
- a reasonable fix may be to default `TERM=xterm-256color` when unset

mgabor3141:
Fixes #154.

## Problem

Inside a gmux shell:
```
$ lazygit
terminal entry not found: term not set
```

The daemon (or any non-interactive parent of `gmux run`, e.g. a systemd unit) may have no `TERM` set in its environment. `ptyserver` was building the child env from `os.Environ() + cfg.Env` and passing the empty `TERM` straight through to the shell, breaking every curses-based program.

`gmux lazygit` worked because that path comes from an interactive terminal that already has `TERM`.

## Fix

`ptyserver` now defaults `TERM=xterm-256color` when the resolved env has none, matching what the xterm.js frontend actually renders. Existing `TERM` values, from either the parent or a caller-supplied extra, still win.

## Bonus: real version in `TERM_PROGRAM_VERSION`

`TERM_PROGRAM_VERSION` was hardcoded to `0.1.0`, which is what fastfetch picked up and reported as `Terminal: gmux 0.1.0` regardless of the actual build. `Config` now takes a `Version`, plumbed from `cmd/gmux/main.go`'s ldflags-set `version`, so the value tracks real releases. Empty `Version` falls back to `dev`, matching the binary's own default.

## Refactor

Env composition is extracted to a pure `buildChildEnv(parent, extra, version)` helper. Documented layering (parent → extra → terminal capability advertisements → TERM default) makes precedence obvious, and the logic is now testable without spawning processes.

## Tests

Five fast unit tests on `buildChildEnv` covering: TERM default when absent, parent TERM preserved, extra overrides parent TERM, terminal capability advertisements always win over the parent, empty version falls back to `dev`. Plus one end-to-end test that spawns a child and confirms the composed env actually reaches it through `exec.Command`.

### DIFF
--- a/cli/gmux/cmd/gmux/run.go
+++ b/cli/gmux/cmd/gmux/run.go
@@ -120,6 +120,7 @@ func runSession(args []string, attach bool) {
 		SocketPath: sockPath,
 		Adapter:    a,
 		State:      state,
+		Version:    version,
 	}
 	// Always try to inherit terminal dimensions from the parent.
 	// Even non-interactive launches (background, piped) benefit from
--- a/cli/gmux/internal/ptyserver/ptyserver.go
+++ b/cli/gmux/internal/ptyserver/ptyserver.go
@@ -146,6 +146,9 @@ type Config struct {
 	Rows       uint16
 	Adapter    adapter.Adapter
 	State      *session.State
+	// Version is reported to children via TERM_PROGRAM_VERSION.
+	// Defaults to "dev" when empty.
+	Version    string
 	// LocalOut, if non-nil, receives a copy of every PTY output chunk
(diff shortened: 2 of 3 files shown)


mgabor3141:
@greptile

PR #183, opened 2026-04-25
feat(release): polish discord announcements and hide release-scoped commits from changelog

mgabor3141:
Three related polish changes to release announcements and the changelog.

## 1. Fall back to the bullet list when highlights are empty

When `RELEASE_HIGHLIGHTS.md` is left empty for a release (the
"patch-only releases don't need prose" path the doc encourages),
`notify-discord.sh` previously sent a Discord post with just a header
and the changelog link. Subscribers had no way to tell what changed
without clicking through.

Now: if the prose above `<!-- highlights-end -->` is empty/whitespace,
fall back to the auto-generated bullet list below the marker. When
prose IS present, the bullets are still suppressed (the curated
paragraph is the canonical summary, and we don't want to double up).

## 2. Trim blank lines on both ends of the summary

The blank line that sits right after the `<!-- highlights-end -->`
marker in `RELEASE_NOTES.md` propagated into the Discord post,
rendering as a double-spaced gap between the version heading and the
bullets. Same risk for highlights files that start with a blank line.
The summary is now stripped of leading and trailing blank lines
before being composed into the message.

## 3. Hide `(release)`-scoped commits from the changelog

Anything in the `(release)` scope is internal release-flow plumbing
(workflows, `notify-discord.sh`, `cliff.toml` itself). Those changes
affect maintainers, not users, so they don't belong in user-facing
release notes and shouldn't trigger a bump on their own.

Implemented as a `cliff.toml` parser rule that skips on
`field = "scope", pattern = "^release$"`. It's placed AFTER the
breaking-change rules so a hypothetical `feat(release)!:` (e.g. an
env-var rename in `notify-discord.sh`) still surfaces under the
Breaking heading. Note: in git-cliff, `scope = "x"` in a parser SETS
the commit's scope; matching the parsed scope requires the explicit
`field`/`pattern` form. The rule is reflexive — this PR's own
commits would be skipped from the next changelog under the new rule.

## Tests

`version_test.sh` (35 passing, +7):

- **Skipped from changelog**: a `feat:` plus `feat(release):` and
  `fix(release):` history. Asserts the user-visible feat is in the
  changelog, the two release-scoped commits are not.
- **No-op release**: a release-scope-only history exits with "No
  releasable commits" and doesn't bump.
- **Breaking still surfaces**: a `feat(release)!:` commit bumps to
  major and lands in the `### Breaking` section.

Mutation-tested: removing the cliff.toml rule causes 3 of those
assertions to fail.

`notify_discord_test.sh` (13 passing, +5):

- **Empty highlights → bullets**: with no prose, asserts bullet text
  reaches Discord and the changelog link survives.
- **No triple-newline gap**: asserts the rendered content has no
  three-consecutive-newlines, pinning the both-ends trim.
- **Highlights present → bullets suppressed**: with prose, asserts
  the bullet body is NOT in the Discord content (no double-up).

## Docs

`AGENTS.md` now mentions the empty-highlights fallback and the
`(release)` scope convention so authors know the conventions for
both.

### DIFF
--- a/.github/workflows/scripts/notify-discord.sh
+++ b/.github/workflows/scripts/notify-discord.sh
@@ -28,9 +28,23 @@ fi
 # RELEASE_NOTES.md has the highlights prose above a `<!-- highlights-end -->`
 # marker, then the generated bullet list. The marker is invisible in
 # rendered markdown; we stop reading at it.
-summary=$(sed '/<!-- highlights-end -->/,$d' RELEASE_NOTES.md)
-# Trim trailing blank lines.
-summary=$(echo "$summary" | sed -e :a -e '/^\n*$/{$d;N;ba}')
+#
+# When the prose is empty (a release without curated highlights), fall
+# back to the bullet list below the marker so Discord still sees what
+# changed. An empty announcement with only a changelog link is worse
+# than a slightly verbose one: most subscribers just want to know if
+# there's anything they care about without having to click through.
+# Strip leading and trailing blank lines so the rendered message
+# doesn't get extra vertical whitespace from sloppy input or from
+# the blank line that sits right after the highlights-end marker.
+trim_blank() { sed -e :a -e '/^\n*$/{$d;N;ba}'; }
+
+summary=$(sed '/<!-- highlights-end -->/,$d' RELEASE_NOTES.md | trim_blank)
+if [[ -z "${summary//[[:space:]]/}" ]]; then
+  # Skip the marker line itself; everything after it is the auto-generated
+  # bullet list (### Features, ### Fixes, etc.).
+  summary=$(sed -n '/<!-- highlights-end -->/,$p' RELEASE_NOTES.md | tail -n +2 | trim_blank)
+fi
 
 # ── Determine bump type ──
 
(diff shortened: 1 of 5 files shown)


mgabor3141:
@greptile

mgabor3141:
## Replies to greptile review

**Comment 1 (P1, "Unquoted \$'\\n\\n\\n' — test always passes"): false positive.**

The analysis confuses ANSI-C-quoted **literals** (`$'...'`) with **variable expansions**. Bash's word-splitting only applies to the result of parameter expansion, command substitution, and arithmetic expansion (`man bash`, "Word Splitting" section). ANSI-C-quoted literals are not subject to it.

Demonstration on bash 5.3:

```
foo() { echo "argc=$#"; for i in "$@"; do printf '  arg=%q\\n' "$i"; done; }
foo a $'\\n\\n\\n' b
# argc=3
#   arg=a
#   arg=$'\\n\\n\\n'
#   arg=b

nl=$'\\n\\n\\n'; foo a $nl b
# argc=2
#   arg=a
#   arg=b
```

The literal form (which is what the test uses) is preserved. The variable form (which the suggested fix would have used, **without** quoting) would actually have introduced the bug.

I also mutation-tested the assertion before pushing: reverting the `trim_blanks` change made the assertion fail (`12 passed, 1 failed` with `✗ no triple-newline gap`), confirming it pins the behavior. No change.

**Comment 2 ("Duplicated comment block"): valid, fixed.**

Real bug from a multi-pass commit split. Removed the duplicate paragraph in commit 2 (`fix(release): trim blank lines on both ends of discord summary`) where it was originally introduced. Pushed as eb029bd.

mgabor3141:
@greptile

PR #184, opened 2026-04-25
feat(cli): print session id for --no-attach with deterministic handshake

mgabor3141:
Closes part of the work that was in #172 (now closed). Smaller, focused replacement.

## What

Make `gmux --no-attach <cmd>` print the session id on stdout deterministically, so scripts can capture and use it immediately:

```sh
id=$(gmux --no-attach pytest --watch)
gmux --tail "$id"   # works without polling, no race
```

## How

A `pipe(2)` handshake between the parent (the `gmux` process the user invoked) and the detached child (the actual session runner). All wire details, lifetime, and error semantics are documented in one block at the top of `cli/gmux/cmd/gmux/handshake.go`.

- Parent creates an `os.Pipe()`, attaches the write end as `cmd.ExtraFiles[0]` (becomes fd 3 in the child), sets `GMUX_HANDSHAKE_FD=3` in the child env, then `cmd.Start()`.
- Parent closes its copy of the write end and blocks on `readHandshake` with a 5s deadline.
- Child runs `registerWithGmuxd` synchronously; on success writes `"<session-id>\n"` to fd 3 and closes; on failure closes without writing.
- Parent reads the line and prints the trimmed id on stdout.

Strict framing: any non-nil read error (EOF, deadline, anything) is a failure regardless of whether partial bytes arrived. The protocol mandates `"<id>\n"` written atomically; a child that crashes mid-write must not produce a half-id on the parent's stdout.

After dispatching the ack, the child unsets `GMUX_HANDSHAKE_FD`. Without this, the runner's environment would still carry the var, and any nested `gmux` the user's command spawns would call `handshakeAck` against a fd that in its own fd table is closed or unrelated.

## Race fix

For fast-exit commands (`echo`, `true`, `false`), the previous `go registerWithGmuxd(...)` could lose the registration race: the main goroutine reached `os.Exit` before the register goroutine even started. The child died, kernel closed its end of the pipe, parent saw EOF.

This PR adds a `regDone` channel closed by the register/handshake goroutine, and the main goroutine blocks on it before `deregisterFromGmuxd` and `os.Exit`. Bounded by `registerWithGmuxd`'s own retry budget (≤2.5 s). As a side effect, this also fixes the pre-existing race in non-handshake paths where a fast-exit command's register/deregister could arrive at gmuxd out of order.

## Failure diagnostics

The parent's stderr message distinguishes the three failure modes a script developer hits:

- `"child process exited before registering"` — read returned `io.EOF` with zero bytes
- `"registration timed out after 5s"` — `readHandshake`'s deadline fired (child wedged)
- raw error otherwise — covers empty/malformed-line cases

## Nested-gmux path unchanged

Running `gmux foo` inside an interactive gmux session still prints `started ... in background (visible in gmux)` on stderr and returns immediately. No handshake plumbing, no stdout output. Only `--no-attach` triggers the new path.

## Tests

`cli/gmux/cmd/gmux/handshake_test.go` covers three layers:

**`handshakeAckFD` (the IO core, no env, no fd aliasing):**
- Writes `"<id>\n"` on success.
- Closes the fd without writing on failure.
- Closes the fd in both cases (verified by ReadAll terminating).

**`openHandshakeFD` / `handshakeAck` (env-lookup glue):**
- Returns nil for unset env, malformed values, and fds 0–2.
- Unsets `GMUX_HANDSHAKE_FD` after dispatch (regression test for the env-leak fix).
- Second call within the same process is a no-op (follows from the unset).
- Common-case fast path: returns immediately when env is absent.

**`readHandshake` (parent-side framing):**
- Round-trip with `handshakeAckFD` (in-process).
- Returns `io.EOF` when child closes without writing.
- Returns `os.ErrDeadlineExceeded` when child hangs.
- Rejects empty-line writes.
- Rejects partial writes ("`sess-abc`" no newline) — strict framing regression test.

**End-to-end subprocess (the only test exercising real `cmd.ExtraFiles` + fd-3 plumbing):**
- `TestPipeHandshakeRoundTripViaSubprocess` re-execs the test binary with the same env+fd layout `spawnDetached` uses in production, has the child call `handshakeAck`, and verifies the parent reads the id back. Catches regressions in fd numbering, env inheritance, and `setsid + ExtraFiles` compatibility that the in-process tests can't see.

End-to-end smoke also verified locally with a real gmuxd: fast-exit (echo, returns id with valid session in `/v1/sessions`) and long-running (sleep, returns id in 32 ms with `alive=true`) both work.

## Replaces #172

PR #172 took a different approach: parent generates the session id, passes it to the child via env, polls `/v1/sessions` to detect registration, and persisted post-exit scrollback to a `<socket>.tail` file with a CLI-side disk-fallback chain.

The new approach is a tighter slice:

- **Replaces:** polling → pipe handshake (deterministic, no race, no env-forced ids).
- **Drops:** `persistTail` / `tailPersistTimeout` / `Server.TailPersisted()` / `.tail` file write / CLI disk-fallback chain. Post-exit `--tail` is deferred to a separate PR that brokers scrollback through gmuxd from a structured state-dir, which gives us readonly web views and reboot-survival as side effects.
- **Drops:** `envForceSessionID` / `nextSessionID` / `isValidSessionID`. Child generates the id as today; the env var is unnecessary.

### DIFF
--- a/cli/gmux/cmd/gmux/daemon.go
+++ b/cli/gmux/cmd/gmux/daemon.go
@@ -143,7 +143,12 @@ func gmuxdHealthy(timeout time.Duration) bool {
 	return resp.StatusCode == http.StatusOK
 }
 
-func registerWithGmuxd(sessionID, socketPath string) {
+// registerWithGmuxd posts the session's registration to gmuxd and
+// returns true once gmuxd accepts it. Retries a handful of times to
+// cover the case where gmuxd is still starting up. Returns false if
+// every attempt fails: callers that care about registration outcome
+// (e.g. the --no-attach handshake) treat false as "give up".
+func registerWithGmuxd(sessionID, socketPath string) bool {
 	baseURL := gmuxdBaseURL()
 
 	payload, _ := json.Marshal(map[string]string{
@@ -163,9 +168,10 @@ func registerWithGmuxd(sessionID, socketPath string) {
 		}
 		resp.Body.Close()
 		if resp.StatusCode == 200 {
-			return
+			return true
 		}
 	}
+	return false
 }
 
 func deregisterFromGmuxd(sessionID string) {
--- a/cli/gmux/cmd/gmux/handshake.go
+++ b/cli/gmux/cmd/gmux/handshake.go
@@ -1,0 +1,115 @@
+package main
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Handshake protocol used by spawnDetached's --no-attach path so the
+// parent process can return a deterministic session id to its caller
+// (typically a script doing id=$(gmux --no-attach foo)) without
+// polling /v1/sessions or guessing.
+//
+// Wire:
+//   - Parent: os.Pipe(); attach write end as cmd.ExtraFiles[0] so the
+//     child sees it as fd 3; set GMUX_HANDSHAKE_FD=3 in the child env;
+//     cmd.Start; close parent's copy of the write end; readHandshake
+//     on the read end with a 5s deadline.
+//   - Child: after registerWithGmuxd, call handshakeAck. On registered
+//     success it writes "<session-id>\n" to fd 3 and closes the fd;
+//     on registered failure it closes without writing — the parent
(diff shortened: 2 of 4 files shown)


mgabor3141:
@greptile

PR #188, opened 2026-04-29
docs(adr): snapshot push protocol and project ownership from session origin

mgabor3141:
First two ADRs in the repo, locking the architectural contract before
the v2.0 implementation work begins. Two independent decisions, one
PR; each ADR is its own commit.

## ADR 0001: Snapshot push protocol

Replaces today's seven-event-type SSE + four parallel HTTP fetches
with a snapshot-based protocol on \`/v1/events\`. Three event types:

- \`snapshot.sessions\` (high cadence; full sessions array)
- \`snapshot.world\` (low cadence; bundled projects + peers + health
  + frontend_config; browser-only)
- \`session-activity\` (lossy notification, fire-and-forget)

\`?as=peer|browser\` query hint controls which event types and which
sessions are emitted to the consumer. Frontend consolidates around
two private signals (\`_rawSessions\`, \`_rawWorld\`) with public
\`computed\` projections and an optimistic-mutation overlay layer.

This is a breaking wire change; ships as v2.0 with no compat shim.

## ADR 0002: Project ownership from session origin

Project membership is owned by the origin host; sessions carry
\`project_slug\` / \`project_index\` stamps populated by the origin's
\`Reconcile\`. Receivers render based on these stamps; empty stamps
fall through to viewer's local match rules (preserving the
no-projects-on-peer / devcontainer case). Per-host \`projects.json\`
no longer needs cross-host sync. Cross-host actions (reorder,
dismiss-claimed) route through a new generic peer proxy at
\`/v1/peers/{peer}/...\`. Disconnected peers render with an
"unavailable" indicator; cleanup is authoritative on reconnect via
the receiver-rule diff against the next \`snapshot.sessions\`.

## Why land these as docs first

The implementation PR will be a substantial breaking refactor
(snapshot composer, peer protocol bump, frontend projection layer,
project-ownership receiver rule, cross-host action proxy). Reviewers
of that PR can read the ADRs first to understand intent and then
read the diff as "does this implement what's specified."

### DIFF
--- a/docs/adr/0001-snapshot-push-protocol.md
+++ b/docs/adr/0001-snapshot-push-protocol.md
@@ -1,0 +1,308 @@
+# ADR 0001: Snapshot push protocol
+
+**Status:** Proposed
+**Date:** 2026-04-25
+**Related:** ADR 0002 (Project ownership from session origin)
+
+## Context
+
+Today's gmuxd uses a hybrid push/pull model for state delivery to its
+consumers (browsers and peer gmuxd instances):
+
+- **Hot data via SSE** at `/v1/events`: per-row deltas
+  (`session-upsert`, `session-remove`, `session-activity`) plus
+  trigger-only events (`projects-update`, `peer-status`) whose
+  semantics are "the named resource changed; go re-fetch it."
+- **Initial state via parallel HTTP fetches**: `/v1/sessions`,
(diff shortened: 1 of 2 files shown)


mgabor3141:
Folding the ADRs into #191 so the spec lands together with the implementation. The two ADR commits are now the first two commits on `feat/snapshot-and-ownership`.

PR #189, opened 2026-04-29
fix(web): trim trailing whitespace and pad on terminal copy

mgabor3141:
## Problem

Copying text from the terminal includes a significant amount of trailing whitespace. xterm.js's `getSelection()` doesn't consistently trim cells that were never written, and for full-rendered rows it has no way to know which spaces are intentional content vs. row padding.

## Approach

Replace both `term.getSelection()` callers in `keyboard.ts` with a new `selectionToText(term)` that walks the active buffer directly, using the **"trailing pad belongs to the line break"** model used by Terminal.app, Alacritty, WezTerm, GNOME Terminal, and xterm:

- Selections that **cross** a row boundary (or reach `end.x === cols`, e.g. triple-click) drop the trailing pad of the row before emitting `\n`.
- Selections that **end inside** a row preserve the cells the user dragged across, spaces included.
- Soft-wrapped rows (xterm.js's `isWrapped` flag) are joined without an inserted `\n`, matching every modern terminal.

Implementation leans on xterm.js's `translateToString(trimRight=true, ...)`, which trims by cell *codepoint* (`0` = never written) rather than by rendered character. That cleanly distinguishes never-written pad cells from explicitly-typed spaces and is automatically wide-character-aware. The whole algorithm is `trim = ex >= cols`, then one `translateToString(trim, sx, ex)` per row plus the `\n`/soft-wrap decision.

This is deliberately not a config toggle. Ghostty's `clipboard-trim-trailing-spaces` option is considered a smell by its own community ([ghostty-org/ghostty#2709](https://github.com/ghostty-org/ghostty/discussions/2709)) because it conflates two different things; the row-extent model handles both correctly.

## Scope

Fixes the trailing-whitespace pain for shell output and any program that lets the terminal do the wrapping.

**Does not** fix copying from full-screen TUIs (vim, tmux, Claude Code, pi, btop, fzf, …). Those programs draw their own wraps via cursor positioning, so each visual row is a hard newline in the buffer, indistinguishable from a real line break. No terminal recovers source text from a TUI's rendered screen without app cooperation; this is an information-loss problem at the protocol layer (cf. [ghostty-org/ghostty#11215](https://github.com/ghostty-org/ghostty/discussions/11215), [#7962](https://github.com/ghostty-org/ghostty/discussions/7962), [#2780](https://github.com/ghostty-org/ghostty/discussions/2780)).

## Coordinate-system note

`getSelectionPosition()`'s public typings claim 1-based coordinates, but the runtime returns 0-based values directly (verified against the bundled xterm.js source). `start.y`/`end.y` are absolute buffer indices, compatible with `buffer.active.getLine`. The implementation file documents this so future maintainers don't lose an hour to it.

## Tests

`selection.test.ts`, 10 cases. The fake `Terminal` honors `trimRight` the same way xterm.js does (drops never-written cells from the slice's right edge while preserving explicitly-typed spaces), so the tests faithfully exercise the algorithm's real dependency:

- empty selection
- selection crosses a line break → drops pad
- selection ends inside a row past content → preserves pad
- **triple-click / line select (`end.x === cols`, single row)** → trims pad, no `\n` (regression test for first-pass bug)
- **explicitly-typed trailing spaces preserved across a line break** (verifies the codepoint-vs-character distinction)
- soft-wrap rejoin (`isWrapped`) without `\n`
- blank intermediate rows preserved
- mid-row to mid-row across rows
- single-row selection has no trailing `\n`
- multi-row selection with varying content lengths

`selectionToText` is typed against a minimal `SelectionTerminal` interface (subset of `@xterm/xterm`'s `Terminal`) so tests use plain object fakes, no DOM or xterm-headless dependency.

### DIFF
--- a/apps/gmux-web/src/keyboard.ts
+++ b/apps/gmux-web/src/keyboard.ts
@@ -22,6 +22,7 @@ import {
   uploadClipboardBlob,
   type UploadResult,
 } from './clipboard-upload'
+import { selectionToText } from './selection'
 
 type SendFn = (data: string) => void
 
@@ -217,7 +218,7 @@ function executeAction(
     }
 
     case 'copyOrInterrupt': {
-      const sel = term.getSelection()
+      const sel = selectionToText(term)
       if (sel) {
         navigator.clipboard.writeText(sel)
         term.clearSelection()
@@ -228,7 +229,7 @@ function executeAction(
     }
 
     case 'copy': {
-      const sel = term.getSelection()
+      const sel = selectionToText(term)
       if (sel) {
         navigator.clipboard.writeText(sel)
         term.clearSelection()
--- a/apps/gmux-web/src/selection.test.ts
+++ b/apps/gmux-web/src/selection.test.ts
@@ -74,7 +74,7 @@ describe('selectionToText', () => {
 
   it('trims pad on triple-click / line select (end.x === cols, single row)', () => {
     // xterm's selectLineAt sets end.x = cols. We must treat that as
-    // "selection reached the row boundary" and drop the pad — otherwise
+    // "selection reached the row boundary" and drop the pad, otherwise
     // every triple-click copies a wall of spaces. Regression test for the
     // bug introduced when first trying the naive "preserve pad on last
     // row" rule.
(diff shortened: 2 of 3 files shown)


mgabor3141:
@greptile

PR #190, opened 2026-04-29
feat: image and binary clipboard paste into sessions

mgabor3141:
Implements image (and any non-text binary) paste into gmux sessions per the design in #85.

When the user pastes a clipboard item that isn't pure text, gmux-web POSTs the bytes to the gmuxd that owns the session, that gmuxd materializes them as `paste-N.<ext>` in its `os.TempDir()`, and gmux-web types the absolute path into the PTY through the existing `formatPasteText` machinery. CC, pi, or anything else that accepts file paths handles it from there. No app-specific glue, no display server on the daemon, no protocol dependence on the running TUI.

## What's in this PR

**Daemon (`feat(daemon)`)**

- New `internal/clipfile/` package: deep `Writer` interface plus `LocalWriter` impl. Picks the next-free `paste-N.<ext>` via `O_CREAT|O_EXCL` retry against the dir, so concurrent pastes from multiple sessions never overwrite each other. MIME → ext map covers `image/{png,jpeg,gif,webp,avif,heic,heif,bmp,tiff,svg+xml}`, `application/pdf`, `video/mp4`, with `bin` fallback for everything else.
- New endpoint: `POST /v1/sessions/{id}/clipboard`. Body cap of 10MB enforced via `http.MaxBytesReader`; oversized requests get `413 too_large`. Empty bodies get `400 empty_body`. Returns `{ ok: true, data: { path } }` on success.
- Devcontainer and network-peer sessions Just Work via the existing `peer.Forward()` machinery: the request is forwarded to the gmuxd that owns the session (running inside the container or on the peer machine), which writes to its own `os.TempDir()`. No `docker exec`, no path translation, no second writer impl. Simpler than the PRD's draft devcontainer story.

**Web (`feat(web)`)**

- New `clipboard-upload.ts` deep helper: `uploadClipboardBlob(blob, sessionId)` returns `{ ok, path } | { ok: false, error }`. Handles MIME header, size cap, error categorization. Pure logic + `fetch`; trivially mockable.
- All three paste paths route through one exported `handlePasteAction`:
  - the **keybind** (Ctrl+V / Cmd+V) via the keymap's `paste` action
  - the **DOM paste event** (right-click, browser Edit menu, native long-press paste on mobile) via `attachPasteHandler`
  - the **mobile toolbar paste button** via the parameterless `onPasteReady` bridge in `terminal.tsx` (changed from a string-typed bridge so the button is no longer text-only)
- Each path inspects the clipboard for non-`text/*` items first, routes through the upload helper if found, then types the returned path via existing `formatPasteText` + `sendRaw`. Text-only clipboards keep the existing behavior verbatim. Optional `PasteFeedback` callback surfaces success and error toast copy; defaults to `defaultPasteFeedback` (exported) which writes to `console` so failures aren't silent.
- The keybind path reads the clipboard once via `Clipboard.read()` and extracts text from the same items when there's no binary, avoiding a redundant `readText()` call (and the second permission prompt some browsers fire for it).
- `pickBinaryDataTransferItem` is exported so its rule (`kind === 'file'` AND non-`text/*` MIME) is unit-tested in isolation; both predicates are load-bearing and a regression in either would cause confusing failure-toast races.

## Tests

- **`clipfile.LocalWriter`** (Go): 6 tests including a 50-goroutine concurrent-write race regression net (asserts distinct paths, no torn writes, contiguous numbering 1..50). Race-clean across `-count=3 -race`.
- **HTTP handler** (Go): 5 `httptest`-driven tests covering happy path, method rejection, size cap, empty body, content-type → ext mapping (verified with a real `LocalWriter` over `t.TempDir()`).
- **`extForMIME`** (Go): table-driven coverage of every documented MIME plus case/whitespace/parameter tolerance and unknown fallback.
- **`clipboard-upload.ts`** (vitest): 11 cases covering MIME picking, web-side size cap (no `fetch` call), success path, error JSON mapping, network failure, session id encoding.
- **`pickBinaryDataTransferItem`** (vitest): 5 cases covering both predicates of the rule, order preservation, and empty input.
- **`keyboard.ts` paste shim** and the new mobile bridge: deliberately skipped per the PRD; the helpers they delegate to are covered by the unit tests above and the bridge is a single function call.

287 frontend tests pass (was 281). All daemon tests pass without `-race`; with `-race` everything I added stays clean across `-count=3`. The pre-existing `internal/notify` race is unrelated.

## Out of scope

- OSC 5522 / Kitty clipboard protocol responder. Independent follow-up; will compose with this without conflict.
- A real in-page toast UI; current feedback is `console`. Adding a toast component is a follow-up that just routes the existing `PasteFeedback` callback.
- `text/uri-list` (Finder paths) special-casing.

## Coexistence with #189

PR #189 (trailing-whitespace fix on copy) also touches `keyboard.ts`. The two PRs touch different parts of the file (copy actions vs paste actions / new `attachPasteHandler` signature) but share the file, so whichever lands second will be a trivial rebase. Both can land in either order.

Closes #85.

### DIFF
--- a/apps/gmux-web/src/clipboard-upload.test.ts
+++ b/apps/gmux-web/src/clipboard-upload.test.ts
@@ -1,0 +1,119 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import {
+  MAX_CLIPBOARD_BYTES,
+  firstBinaryType,
+  uploadClipboardBlob,
+  type UploadResult,
+} from './clipboard-upload'
+
+describe('firstBinaryType', () => {
+  it('returns the first non-text MIME', () => {
+    expect(firstBinaryType(['text/plain', 'image/png', 'text/html'])).toBe('image/png')
+  })
+
+  it('returns null when only text MIMEs are present', () => {
+    expect(firstBinaryType(['text/plain', 'text/html', 'text/uri-list'])).toBe(null)
+  })
+
+  it('returns null on empty input', () => {
+    expect(firstBinaryType([])).toBe(null)
+  })
+
+  it('treats image/svg+xml as binary even though it is structurally textual', () => {
+    // An SVG copied from a graphics app is the user's intended payload;
+    // upload preserves intent. Same reasoning as image/png: prefer binary
+    // when binary is present.
+    expect(firstBinaryType(['text/plain', 'image/svg+xml'])).toBe('image/svg+xml')
+  })
+
+  it('preserves order: first non-text wins', () => {
+    expect(firstBinaryType(['image/png', 'image/jpeg'])).toBe('image/png')
+    expect(firstBinaryType(['image/jpeg', 'image/png'])).toBe('image/jpeg')
+  })
+})
+
+describe('uploadClipboardBlob', () => {
+  const originalFetch = globalThis.fetch
+  afterEach(() => {
+    globalThis.fetch = originalFetch
+  })
+
+  function makeBlob(size: number, type = 'image/png'): Blob {
+    return new Blob([new Uint8Array(size)], { type })
+  }
+
+  it('rejects oversized blobs without calling fetch', async () => {
+    const fetchSpy = vi.fn()
+    globalThis.fetch = fetchSpy as unknown as typeof fetch
+
+    const blob = makeBlob(MAX_CLIPBOARD_BYTES + 1)
+    const result = await uploadClipboardBlob(blob, 'sess-1')
+
+    expect(result.ok).toBe(false)
+    if (!result.ok) expect(result.error).toBe('too_large')
+    expect(fetchSpy).not.toHaveBeenCalled()
+  })
+
+  it('POSTs to the session-scoped endpoint with the blob MIME and bytes', async () => {
+    let captured: { url: string; init: RequestInit } | null = null
+    globalThis.fetch = (async (url: string, init: RequestInit) => {
+      captured = { url, init }
+      return new Response(
+        JSON.stringify({ ok: true, data: { path: '/tmp/paste-1.png' } }),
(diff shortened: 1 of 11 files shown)


mgabor3141:
@greptile please re-review.

Since the last review I addressed both P2 findings and squashed in some self-review polish:

- **clipboard-upload.ts JSON-parse catch**: collapsed the dead `if (!resp.ok)` branch to a single return as suggested.
- **keyboard.ts paste handler param naming**: renamed `attachPasteHandler`'s `send` parameter to `sendRaw` and updated the doc-comment to call out that the keybind `paste` action takes the same care, so the contract is explicit at both call sites.
- **keybind paste path**: now reads the clipboard once via `Clipboard.read()` and extracts text from those same items when there's no binary, instead of doing a redundant `readText()` call (which on some browsers triggers a second permission prompt).
- **MIME → ext map**: expanded to cover `image/{avif,heic,heif,bmp,tiff,svg+xml}` (iPhone, modern web, macOS Preview).
- **Mobile toolbar paste button**: was previously text-only because the `terminal.tsx → main.tsx` bridge was string-typed (`(text: string) => void`). Changed it to a parameterless trigger and routed it through the now-exported `handlePasteAction`, so all three paste paths (keybind, DOM event, mobile button) share one inspect-and-route implementation.
- **`pickBinaryDataTransferItem`**: exported with a 5-case vitest covering both predicates (`kind === 'file'` AND non-`text/*`), order preservation, and empty input.
- **`defaultPasteFeedback`**: exported so `terminal.tsx` reuses one console sink instead of duplicating it inline.

Final shape is two clean commits (`feat(daemon)` + `feat(web)`), no follow-up `fix:` or `refactor:` bullets to pollute the changelog. 287 vitest pass; daemon races clean across `-count=3 -race`.

PR #191, opened 2026-04-30
feat: snapshot push protocol + cross-host project ownership

mgabor3141:
## Summary

Five commits, rebase-merged: two ADRs, one squashed implementation
commit, and two follow-up bug fixes from post-rebase audit.

Adds protocol 2 (snapshot push) and per-host project ownership to
gmuxd, **keeping the v1 wire surface alongside as deprecated**.
v1.x browsers, v1.x hubs, and scripts hitting the bulk GET
endpoints continue to work against a v2 daemon; a v2 hub continues
to consume sessions from a v1.x spoke through a staggered upgrade.
Releases as a normal `feat` (minor bump); not a breaking change.

> **Note on commit shape.** The branch originally landed as 17 atomic
> commits and was rebase-tracking `main` over several weeks. After
> `main` absorbed two parallel features (the live↔dead view ADRs that
> reshape `sidebar.tsx` / `project-hub.tsx` to remove `onResume`, and
> the new `rehydrateProjects` startup path), every commit on this
> branch became conflicted because the same regions were edited from
> both sides at every step. Resolving in place would have produced
> half-broken intermediate commits with embedded conflict markers.
> Squashed to a single `feat` instead, with the user-facing story
> unchanged. The two ADRs are kept as separate commits because the
> spec is worth preserving as a self-contained reading.

## What changes for users

- **Multi-host project ownership.** Each host owns its own
  `projects.json`. Sessions arriving from a peer are stamped with the
  peer's project assignment and rendered under that peer's folder, in
  the order the peer arranged it. Disclaimed sessions fall through to
  the viewer's match rules. Drag-reorder on a peer-owned folder
  routes through `/v1/peers/{peer}/...` so the peer applies the
  change to its own `projects.json` and re-broadcasts the result.
- **`@<peer>` in URLs.** Peer-owned project hubs are
  `/@<peer>/<slug>`; adopted-from-peer sessions inside local folders
  keep an `@<host>` mid-path segment. Local sessions on local folders
  have no `@` segment.
- **"Unavailable" indicator.** Sessions whose peer is disconnected,
  reconnecting, or unknown render dimmed with a muted X icon (sidebar
  and project-hub).
- **Snapshot-push protocol.** SSE gains `snapshot.sessions` /
  `snapshot.world` events; the v2 web UI hydrates from those rather
  than polling `/v1/sessions`, `/v1/projects`, `/v1/health` on load
  or reconnect. Reconnect is a single round-trip.

## Backwards compatibility

Per ADR 0001's "Backwards compatibility" section:

- **Wire**: `session-upsert`, `session-remove`, `projects-update`,
  `peer-status` continue to be emitted to non-asPeer subscribers
  (v1 browsers, v1 hubs that don't send `?as=peer`). v2 consumers
  attach listeners only for the new snapshot events and ignore the
  per-event chatter. The v2 hub's `handleEvent` decodes both
  per-event and snapshot frames idempotently, so it works against
  v1 and v2 spokes.
- **HTTP**: `GET /v1/projects` is restored; `GET /v1/sessions` and
  `GET /v1/health` were never removed. All three are documented as
  deprecated in v2: no new fields, but will not be removed in a
  minor release.
- **Asymmetric semantics for `?as=peer` subscribers**: v2 hubs
  using `?as=peer` get `snapshot.sessions` (+ filtered activity)
  only; the v1 per-event stream is suppressed for them to avoid
  double-processing alongside the snapshot.

Additive-only frontend changes:

- New session fields `project_slug` / `project_index`. Old clients
  ignore unknown fields.
- New URL path shape `/@<peer>/<slug>/...` and adopted-peer
  `@<host>` segments. Reachable only when v2 web is served.

## Commits

| # | Type | Subject |
|---|------|---------|
| 1 | docs(adr) | capture snapshot push protocol decision |
| 2 | docs(adr) | capture project ownership from session origin decision |
| 3 | feat | snapshot push protocol and cross-host project ownership |
| 4 | fix(web) | route navigateToSession through viewToPath for peer-owned URLs |
| 5 | fix(web) | strip peer namespace and drop adopted-peer keys on folder reorder |

The squashed commit walks through the implementation in the same
logical chunks the original 17 commits used:

- Snapshot push protocol (ADR 0001): pump routing, lazy compose,
  health factoring, asPeer activity gating, peering reconciliation,
  protocol-1 dispatch coexistence, frontend consumer.
- Cross-host project ownership (ADR 0002): origin-stamped
  ProjectSlug / ProjectIndex, Reconcile pivot, peer-write proxy
  with allowlist, partial-reorder PATCH, frontend folder identity
  refactor.
- Unavailable indicator: pure helper + dim/muted-X UX.
- Frontend signal layout: `_rawSessions` + `_rawWorld` +
  `_pendingMutations` overlay.

## Post-rebase audit & follow-up fixes

**Commit 4 — peer-owned `navigateToSession`.** The helper was
building URLs via `matchSession` + `sessionPath`, ignoring peer
ownership. For a peer-owned session it would either short-circuit
(no local match rule) or omit the `/@<peer>` prefix. Routed both
that helper and the in-place slug-rename URL update through
`viewToPath`, which already understands ADR-0002 ownership.

**Commit 5 — namespace-stripping on folder reorder.** Greptile's
review caught that `s.slug || s.id` for a slugless peer-owned
session evaluates to the hub-namespaced id (`"orig@peer"`), which
the peer's `ReorderSessions` treats as a brand-new key and prepends
to its `projects.json`. Same shape of bug locally for adopted-peer
rows. Extracted `reorderKeysForFolder(sessions, folder.peer)` as a
pure helper that filters to owner-owned sessions and strips the
`@<peer>` chain from slugless ids.

## Verification

- `pnpm --filter @gmux/web test` — 384 tests passing
- `cd services/gmuxd && go test -race ./...` — green except for one
  pre-existing data race in `internal/notify` (unrelated to this
  branch; reproduces on `main`, surfaces only under `-race`)
- `pnpm --filter @gmux/web exec tsc --noEmit` — clean
- `pnpm install --frozen-lockfile` — clean

## Notable correctness fixes folded into commit 3

- **Activity for peer-owned sessions silently dropped.** Browser
  subscribers were running the same activity filter as `?as=peer`
  consumers, which discards namespaced ids. Without the fix, the UI
  never animates activity dots on peer-owned sessions in v2.
  `shouldForwardActivity(asPeer, id, isLocalPeer)` gates on
  `asPeer`.

- **Browser `snapshot.sessions` excluded peer-mirrored sessions.**
  `composeSessions` unconditionally applied the `isOwned` filter,
  so the v2 browser would never see peer-owned sessions and the
  ADR-0002 peer folders wouldn't render. Now `asPeer` keeps the
  filter; browser passes `nil` (full store).

- **Stale `ProjectIndex` after every `projects.json` edit.**
  `projects.Manager.Broadcast` runs `Reconcile` to re-stamp every
  owned session's `ProjectSlug` / `ProjectIndex`, then fires
  `projects-update`. Routing that to the world coalescer alone left
  `snapshot.sessions` un-emitted, so the UI kept rendering each
  session under its previous project. `snapshotPumpRoute(eventType)`
  returns `(pushSessions, pushWorld)`; `projects-update` fires both.

- **Rehydrated sessions un-stamped on first snapshot.**
  `reconcileProjectStamps(state)` now runs immediately after
  `rehydrateProjects(...)` on startup so the first SSE subscriber
  sees stamps without needing a separate projects-update.

## Release prose

`RELEASE_HIGHLIGHTS.md` no longer exists on `main` (release prose
is now driven from the `release/next` PR body between
`<!-- prose-start -->` / `<!-- prose-end -->` markers). Prose for
this release will go onto that PR when it opens.

### DIFF
--- a/apps/gmux-web/src/home.tsx
+++ b/apps/gmux-web/src/home.tsx
@@ -72,7 +72,7 @@ export function Home() {
         <section class="home-projects">
           <h2 class="home-section-title">Projects</h2>
           <div class="home-project-grid">
-            {foldersVal.map(f => <ProjectCard key={f.path} folder={f} />)}
+            {foldersVal.map(f => <ProjectCard key={f.key} folder={f} />)}
           </div>
         </section>
       )}
@@ -101,9 +101,13 @@ export function Home() {
 function ProjectCard({ folder: f }: { folder: Folder }) {
   const alive = f.sessions.filter(s => s.alive).length
   const resumable = f.sessions.filter(s => !s.alive && s.resumable).length
+  const href = f.peer ? `/@${f.peer}/${f.slug}` : `/${f.slug}`
   return (
-    <a class="home-project-card" href={`/${f.path}`}>
-      <div class="home-project-name">{f.name}</div>
+    <a class="home-project-card" href={href}>
+      <div class="home-project-name">
+        {f.peer && <PeerLabel name={f.peer} />}
+        {f.name}
+      </div>
       <div class="home-project-count">
         {alive > 0 && <span class="home-project-alive">{alive} alive</span>}
         {alive > 0 && resumable > 0 && <span class="home-project-rest"> · </span>}
--- a/apps/gmux-web/src/launcher.tsx
+++ b/apps/gmux-web/src/launcher.tsx
@@ -106,11 +106,15 @@ interface LaunchButtonProps {
 }
 
 export function LaunchButton({ className, onLaunch, beforeLaunch, cwd, peer, sessions, selectedId, fallbackCwd }: LaunchButtonProps) {
-  // Resolve the target: context-aware mode (sessions provided) takes
-  // priority over explicit cwd/peer.
+  // Resolve the target: context-aware mode (sessions provided) derives
+  // a smart cwd, while an explicit `peer` prop is always authoritative
+  // (the caller knows the folder's owner; ADR 0002). This matters for
+  // peer folders whose sessions are all dead-resumable, where context
+  // mode would otherwise lose the peer in resolveTarget's fallback.
   const target = useMemo((): LaunchTarget => {
     if (sessions && fallbackCwd !== undefined) {
-      return resolveTarget(sessions, selectedId ?? null, fallbackCwd)
+      const ctx = resolveTarget(sessions, selectedId ?? null, fallbackCwd)
+      return peer ? { ...ctx, peer } : ctx
     }
     return { peer, cwd: cwd || '' }
   }, [sessions, selectedId, fallbackCwd, cwd, peer])
--- a/apps/gmux-web/src/project-hub.tsx
+++ b/apps/gmux-web/src/project-hub.tsx
@@ -9,7 +9,7 @@ import type { HostNode } from './projects'
 import { buildProjectTopology } from './projects'
 import { sessionPath } from './routing'
 import { LaunchButton } from './launcher'
-import { sessions, projects, peers } from './store'
+import { sessions, projects, peers, peerStatusByName, isSessionUnavailable } from './store'
 import { PeerLabel } from './peer-label'
 
 function projectRemote(p: ProjectItem | undefined): string | undefined {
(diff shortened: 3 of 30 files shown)


mgabor3141:
@greptile

PR #192, opened 2026-04-30
fix(web): preload terminal font before mounting xterm

mgabor3141:
On first load the terminal often ends up shorter than the available area, leaving a gap at the bottom on desktop or pushing content below the fold on mobile. The gap is one mobile-bar's worth of pixels (~52 px), which made it look like a layout problem. It isn't.

## Cause

xterm sizes its row count from the first cell measurement (`measureTerminalFit` divides shell height by `term.dimensions.css.cell.height`). That measurement runs as soon as `term.open(container)` is called, which happens before the terminal's font has finished downloading: `@fontsource/fira-code` only registers the `@font-face` declaration eagerly, the actual woff2 isn't fetched until something uses the family. xterm renders, browser starts the fetch, 30-100 ms later the font arrives, xterm internally re-measures (cell height drops 18 to 17 px) and the rendered xterm shrinks. Nothing else triggers our resize pipeline (shell height didn't change), so the row count stays at the fallback-font value and we end up with a row's worth of unused space at the bottom.

Captured in chrome-devtools against the running gmuxd:

```
t=132 rows=41 cellH=18 xtermH=744
t=172 fonts loadingdone -> cellH=17 xtermH=703   (xterm re-measured, fit didn't)
```

## Why `document.fonts.ready` isn't enough

The first thing I tried was awaiting `document.fonts.ready` before opening xterm. Doesn't work: `ready` resolves when the *currently loading* set is empty. At mount, no fonts are in-flight (only declared), so `ready` resolves immediately. The fetch only starts when xterm renders, after our await has unblocked.

## Fix

Use `document.fonts.load(<spec>)` instead. That API explicitly forces the font to download and resolves with the loaded `FontFace[]`. Gate xterm mount on it via a tiny `fontReady` state:

```ts
const [fontReady, setFontReady] = useState(false)

useEffect(() => {
  const spec = `${terminalOptions.fontSize}px ${terminalOptions.fontFamily}`
  document.fonts.load(spec).then(() => setFontReady(true))
}, [terminalOptions.fontFamily, terminalOptions.fontSize])

useEffect(() => {
  if (!fontReady) return
  // ... create Terminal, term.open(...), etc.
}, [..., fontReady])
```

After: xterm's first cell measurement uses the real font metrics, no second-frame resize, no flicker.

```
t=216 terminal mounts  rows=44 cellH=17 xtermH=754   (correct from first frame)
```

The preload runs in parallel with everything else (data fetch, store hydration, WebSocket connect), so the visible delay is negligible: typically already covered by the existing `terminal-loading` overlay.

## Drive-by: tighter `terminalOptions` type

xterm's `ITerminalOptions` marks every field optional, but our schema fills in defaults for `fontSize`/`fontFamily`/etc. Added a `ResolvedTerminalOptions` type (`Omit<ResolvedSettings, 'keybinds' | 'macCommandIsCtrl'> & { theme }`) so consumers don't need defensive `?? '...'` fallbacks at every call site. The store signal and `<TerminalView>` prop both use the narrower type.

### DIFF
--- a/apps/gmux-web/src/settings-schema.ts
+++ b/apps/gmux-web/src/settings-schema.ts
@@ -10,7 +10,7 @@
  * validation happens here in the frontend.
  */
 import * as v from 'valibot'
-import type { ITerminalOptions, ITheme } from '@xterm/xterm'
+import type { ITheme } from '@xterm/xterm'
 
 // ── Helpers ──
 
@@ -209,6 +209,14 @@ export const SettingsSchema = v.pipe(
 export type SettingsConfig = v.InferInput<typeof SettingsSchema>
 export type ResolvedSettings = v.InferOutput<typeof SettingsSchema>
 
+/** xterm options after our schema has filled in defaults. Strictly
+ *  narrower than xterm's `ITerminalOptions` (which marks every field
+ *  optional), so consumers can rely on `fontFamily`, `fontSize`, etc.
+ *  being defined. */
+export type ResolvedTerminalOptions =
+  & Omit<ResolvedSettings, 'keybinds' | 'macCommandIsCtrl'>
+  & { theme: ITheme }
+
 // ── Default theme colors ──
 
 export const DEFAULT_THEME_COLORS: ITheme = {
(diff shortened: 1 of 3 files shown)


mgabor3141:
Good catch from greptile. Reverted to `.finally()` and added a comment explaining the failure mode (network/CSP, not just CSS parse error). My earlier reasoning that `.then()` would surface user-input issues only held for the parse-error case; network failures shouldn't permanently block terminal mount. Pushed in 0310bda.

PR #195, opened 2026-05-02
feat(release): drive release notes from the release PR body

mgabor3141:
Reworks the release pipeline so the open `release/next` PR body is the single source of truth for release prose. Editing prose no longer requires a separate file or a manual regeneration script.

## What changes

**For maintainers writing release prose:**

- Edit prose directly in the `release/next` PR body, between the `<!-- prose-start -->` and `<!-- prose-end -->` markers. The regen workflow re-syncs `changelog.mdx` and the bullets section of the body on every edit.
- No more `RELEASE_HIGHLIGHTS.md` file.
- No more `scripts/regen-release-notes.sh`.
- No more sticky "Discord preview" comment: the prose section *is* the preview.
- No more duplicated install snippet in the body: `pr-build.yml`'s sticky comment is the single install pointer.

**Mechanics:**

- `version.yml` → `regen.yml`. Triggers on `push` to `main`, `pull_request` (opened/edited/synchronize/reopened) to `main`, and `workflow_dispatch`. Three modes:
  - `skip` for non-`release/next` PRs (early-exits cleanly so branch protection's required `regen` check doesn't break unrelated PRs).
  - `noop` for events from `github-actions[bot]` (the workflow's own force-pushes and body edits): breaks the loop while still reporting the status check on the new SHA.
  - `regen` otherwise: reads existing prose from the PR body, regenerates bullets via git-cliff, writes `changelog.mdx` + `.github/release-target` + `pr-body.md`, and lets peter-evans force-sync `release/next`.
- All triggers share a single concurrency group `regen` so push-to-main and a concurrent PR-body edit can't race on the force-push.
- `version.sh` reads optional prose via the `$PROSE` env var. New `--extract-prose` mode parses prose out of a PR body on stdin (handles HTML comments, preserves user horizontal rules) so the workflow can round-trip existing prose through a regen.
- `extract-release-notes.sh` is the single source of truth for "what GoReleaser and Discord see": both `release.yml` and `notify-discord.sh` go through it. Strips the `## vX.Y.Z` heading and exactly one *trailing* `---` separator (a horizontal rule inside prose survives).
- `release.yml`'s `tag` job is unchanged in shape; `release` job extracts notes via `extract-release-notes.sh`. `RELEASE_NOTES.md` is gone; the post-release `--clear-highlights` step is gone.

## Branch protection

After merge: add the `regen` job's status check as required on `main`. Don't enable "require branches to be up to date before merging"; the regen workflow already keeps `release/next` synced with `main` via the push-to-main trigger.

## Tests

- `version_test.sh`: rewritten for the new shape. Pins prose round-tripping (`--extract-prose` is the inverse of the body's prose injection), marker handling, the empty-prose hint behavior, `release-target` content, and changelog insertion order. 41 cases.
- `notify_discord_test.sh`: rewritten to use `changelog.mdx` fixtures (latest entry extraction, prose/bullet split at `### `, horizontal-rule survival, regression test that the trailing `---` doesn't leak into the Discord summary). 17 cases.
- `extract_release_notes_test.sh` (new): direct contract tests for the shared helper. Regression for the trailing-separator bug Greptile caught, plus prose-only entries, first-ever release, and degenerate-input cases. 11 cases.
- All three wired into `ci.yml`'s `release-scripts` job.

## Migration

This lands first; on the next push to `main` after merge, `regen.yml` force-recreates `release/next` (#186) with the new shape, and v1.5.0 ships via the new pipeline.


mgabor3141:
Fixed in c4524a73: stop awk extraction at the trailing `---` separator (matching how `notify-discord.sh` handles bullets) so it doesn't render as an `<hr>` in the GoReleaser body. Verified locally against the current changelog.

mgabor3141:
Follow-up improvements after self-review:

- **Concurrency**: changed regen.yml to a single `regen` group so push-to-main and a concurrent PR-body edit can't both fire peter-evans force-pushes in parallel.
- **Explicit `ref: main`** in regen.yml's checkout: version.sh always sees main's history regardless of trigger; peter-evans always bases release/next on main.
- **Refactored** the latest-changelog-entry extraction (which Greptile flagged a bug in) into a shared `extract-release-notes.sh` helper. Both `release.yml` and `notify-discord.sh` now go through it, so the trailing-`---` regression class can't reappear in only one of them. Prose extraction in notify-discord.sh simplified accordingly.
- **New `extract_release_notes_test.sh`**: 11 direct contract tests for the helper, including a pinned regression test for the trailing-separator bug and tests for horizontal-rule preservation, prose-only entries, first-ever release, and degenerate input.
- **Stale comment cleanup** in `changelog.mdx` and AGENTS.md.

Total test coverage: 41 + 17 + 11 = 69 cases across the three test scripts, all passing locally.

mgabor3141:
@greptile

PR #201, opened 2026-05-05
fix(web): preserve scroll distance from bottom across pi end-of-turn redraws

mgabor3141:
## The reported bug

> Sometimes I scroll up from the bottom in a session that I haven't touched in a while and I jump to the middle of the conversation on the first scroll action.

Always while pi was actively outputting, with the user scrolled up at the moment a frame arrives. A live observation later confirmed the bug also fires with **no user interaction** \u2014 just a pi frame landing while the user is already scrolled up. So the precondition is simply "user not at bottom when pi emits a redraw frame", not "user just scrolled".

## Diagnosis

The BSU/ESU scroll-restore path in `apps/gmux-web/src/terminal-io.ts` had two branches:

- **Wipe branch** (`baseY < prevBaseY`): used distance-from-bottom restore. Designed for pi's end-of-turn frames that emit `\\x1b[3J` followed by a small redraw, shrinking scrollback.
- **Else branch**: trusted xterm's post-parse `viewportY` as `adjustedY`, then `scrollToLine(min(adjustedY, baseY))`.

The `baseY < prevBaseY` heuristic was a *symptom* of buffer reset, not the *cause*. Pi can also emit `\\x1b[3J` + a long redraw that grows `baseY` past its pre-frame value (long thread, conversation rerendered top-down). When the user is scrolled up, `isUserScrolling` stays true through the synchronized block; `\\x1b[3J` resets `ydisp` to 0 mid-parse and the redraw's appended lines don't pull `ydisp` along. Post-parse `adjustedY = 0`. The else branch ran `scrollToLine(0)`. The user landed at the top of the rerendered buffer \u2014 which, because pi rerenders the conversation chronologically, the user perceives as being yanked into the middle (or beginning) of the conversation history.

## Verification

Probe PR #203 ran the new e2e test from this PR against `main` alone. It failed with:

```
[grown-baseY] prevBaseY=66 postBaseY=211 viewportY=0 distance=211
```

Confirming all three claims above: xterm pins `ydisp=0`, `baseY` grew past `prevBaseY` so the old heuristic missed, and the else branch produced `viewportY=0`. All other e2e tests on `main` passed. The fix in this PR makes the test pass.

## Fix

Replace the indirect `baseY < prevBaseY` discriminator with the direct cause: byte-presence of `\\x1b[3J` (`CSI_3J`) anywhere in the BSU/ESU block, OR-accumulated across split chunks via `maybeMarkBufferReset()`.

- **Buffer-reset frames** (any `\\x1b[3J`): anchor \u2192 distance \u2192 bottom restoration runs regardless of which side of `prevBaseY` the rebuilt buffer ends up on.
- **Plain streaming output** (no `\\x1b[3J`, eg `tail -f`): line-based restore stays. Eviction tracking continues to keep the user's specific line visible \u2014 the right behavior for append-only logs.

Drops the now-redundant `prevBaseY` field from `savedScroll`. Adds `bufferReset` boolean lifecycle-coupled to `savedScroll`. Net state count unchanged.

## Tests

Unit tests (`apps/gmux-web/src/terminal-io.test.ts`):

1. **Pi-style redraw with grown baseY**: pins the discriminator behavior at the algorithm level. Models xterm's pinned `ydisp` via the harness.
2. **`tail -f` streaming**: pins line-based restore for non-`\\x1b[3J` frames so a future "always distance-from-bottom" regression fails loudly.
3. **Split BSU / 3J / ESU chunks**: pins the OR-accumulation contract.
4. Existing wipe-branch tests updated to put `\\x1b[3J` in the bytes.

E2E (`e2e/tests/terminal-scroll.spec.ts`, second commit):

5. **Real-xterm grown-baseY**: 100-line seed + 250-line redraw with `\\x1b[3J` so post-wipe `baseY > prevBaseY`. This is the test that failed on `main` in the probe.

## Considered but rejected

- **Always use distance-from-bottom**: simpler, but yanks `tail -f` users toward the bottom as new lines stream in even though their line is still in the buffer.
- **Anchor-line presence check** (`getLine(adjustedY) === prevAnchorLine`): more general but indirect. Byte-presence is more direct and easier to reason about.
- **Dual signal** (`\\x1b[3J` byte OR `baseY < prevBaseY`): would catch hypothetical reset codes other than `\\x1b[3J`. None of those are realistic inside a synchronized output block in practice; rejected for simplicity.

### DIFF
--- a/apps/gmux-web/src/replay.ts
+++ b/apps/gmux-web/src/replay.ts
@@ -19,6 +19,17 @@ export const BSU = new Uint8Array([0x1b, 0x5b, 0x3f, 0x32, 0x30, 0x32, 0x36, 0x6
 /** End Synchronized Update: CSI ? 2026 l */
 export const ESU = new Uint8Array([0x1b, 0x5b, 0x3f, 0x32, 0x30, 0x32, 0x36, 0x6c])
 
+/**
+ * Erase Saved Lines (clear scrollback): CSI 3 J.
+ *
+ * Resets xterm's `ybase` and `ydisp` to 0 mid-frame, breaking the
+ * line-tracking invariant that the BSU/ESU restore path relies on. Its
+ * presence inside a synchronized-output block is the signal to fall back
+ * to distance-from-bottom restoration instead of trusting the post-parse
+ * `viewportY`. Pi emits this as part of its end-of-turn redraw shape.
+ */
+export const CSI_3J = new Uint8Array([0x1b, 0x5b, 0x33, 0x4a])
+
 export type FlushCallback = (chunks: Uint8Array[]) => void
 
 export type ReplayState = 'waiting' | 'buffering' | 'done'
--- a/apps/gmux-web/src/terminal-io.test.ts
+++ b/apps/gmux-web/src/terminal-io.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it, vi } from 'vitest'
 import { createTerminalIO, type ScrollAccessor } from './terminal-io'
-import { BSU, ESU } from './replay'
+import { BSU, CSI_3J, ESU } from './replay'
 
 function makeHarness() {
   const writes: Array<string> = []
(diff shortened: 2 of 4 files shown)


mgabor3141:
## Probe verified the bug mechanism

The new e2e test in this PR was first run against `main` alone via probe PR #203. It failed with this diagnostic:

```
[grown-baseY] prevBaseY=66 postBaseY=211 viewportY=0 distance=211
```

This pins down what was previously hypothesis:

1. xterm **does** pin `ydisp=0` when `isUserScrolling=true` through a synchronized block containing `\\x1b[3J` + a long redraw. Post-frame `viewportY=0`.
2. `baseY` grew past `prevBaseY` (66 \u2192 211), so the `baseY < prevBaseY` heuristic on `main` misses.
3. The else branch trusts `viewportY=0` and the user lands at the top of the rebuilt buffer.

This also matches a live user observation that the bug fires with **no user interaction** \u2014 just a pi frame arriving while the user is scrolled up. The wheel-tick precondition I described in the original write-up was sufficient but not necessary; the real precondition is "user not at bottom when a `\\x1b[3J` frame arrives".

All other e2e tests on `main` passed. The fix in this PR (byte-presence discriminator + distance-from-bottom restore) makes the same test pass.

Probe PR #203 closed.

PR #203, opened 2026-05-05
probe: does grown-baseY scroll preservation fail on main?

mgabor3141:
## Purpose

Verification probe for #201. Pushes the e2e regression test from that PR alone, against `main`, to confirm whether the bug it pins actually reproduces under real xterm.

## Expected outcomes

- **Test fails** \u2192 hypothesis confirmed, the bug exists. Close this PR. Merge #201 (which contains the same test + the fix).
- **Test passes** \u2192 hypothesis wrong. The user's "jump to middle" report has a different root cause. Close #201 and reopen investigation.
- **Test flakes** \u2192 tighten it before shipping either way.

## What the test does

Real xterm via the `__gmuxInject` hook. Seeds 100 lines of scrollback (`prevBaseY \u2248 60`), scrolls up by 3 rows, then injects `BSU + \\x1b[2J\\x1b[H\\x1b[3J + 250-line-redraw + ESU` so the post-wipe `baseY \u2248 210` ends up *larger* than `prevBaseY`. Asserts the user lands 3 rows above the new bottom.

The pre-existing wipe-branch logic only triggered the buffer-reset restore when `baseY < prevBaseY`; this shape falls through to the line-based else branch, which trusts xterm's post-parse `viewportY`. The hypothesis under test: when `isUserScrolling` is true and `\\x1b[3J` resets `ydisp` mid-frame, that post-parse `viewportY` is unreliable.

**Do not merge.** This branch is a probe; if it fails on CI we have our answer.

### DIFF
--- a/e2e/tests/terminal-scroll.spec.ts
+++ b/e2e/tests/terminal-scroll.spec.ts
@@ -708,4 +708,69 @@ test.describe('terminal scrollback (jump-to-top bug)', () => {
     // don't pin it), but the line text is stable.
     expect(landedText).toBe(anchorText)
   })
+
+  /**
+   * Synthetic redraw whose post-wipe `baseY` ends up LARGER than the
+   * pre-wipe value. The earlier baseY-shrink heuristic missed this
+   * shape — it only fired the buffer-reset branch when scrollback
+   * shrank during the frame. Anything that grew baseY past prevBaseY
+   * fell through to the line-based else branch, which trusted xterm's
+   * post-parse `viewportY`.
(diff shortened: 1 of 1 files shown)


mgabor3141:
Probe successful: the new e2e test fails on `main` with `viewportY=0, baseY=211, distance=211` \u2014 the literal `scrollToLine(0)` outcome we predicted. Hypothesis verified, no further action needed on this branch.

#201 carries the same test plus the fix; merging that resolves the bug.

PR #206, opened 2026-05-05
feat(web): replay scrollback for dead sessions

mgabor3141:
## What

Clicking a dead session in the sidebar now shows a read-only terminal populated with that session's persisted scrollback, instead of the previous "Session ended" placeholder. This is the user-visible payoff of #185 (sessionmeta) and #193 (scrollback persistence + broker).

Live sessions still go through `TerminalView` exactly as before; only the dead-session arm of `main.tsx`'s selection switch changes.

## How

- New `apps/gmux-web/src/replay-fetch.ts`: typed wrapper around `GET /v1/sessions/<id>/scrollback`, returning `{ kind: 'bytes' | 'empty' | 'not-found' | 'error' }`. URL-encodes the session ID so peer-owned (`@peer`) IDs round-trip correctly.
- New `apps/gmux-web/src/replay-view.tsx` (~125 lines): mounts a fresh xterm on `session.id` change, fetches scrollback once, writes the bytes, scrolls to bottom. No WebSocket, no input handlers, no resize messages, no mobile bar wiring. OSC 52 (set clipboard) sequences in the captured bytes are suppressed so replaying can't silently overwrite the operator's clipboard. xterm's `scrollback` line cap is bumped from the default 1000 to 10000 to actually fit gmuxd's ~2 MiB persistence window.
- `main.tsx`: adds one ternary arm above the existing fallback. Dead-session path: `selectedVal && !alive && termOpts && !USE_MOCK → <ReplayView />`. Mock mode unchanged (mock sessions are all alive); the previous `'Session ended'` placeholder is now only reached for live sessions whose socket isn't ready yet, so its label simplifies to `'Connecting…'`.

## What's deliberately not in this PR

- **No "session ended" banner inside the replay view.** The sidebar already distinguishes alive/dead; the main header still renders `MainHeader` with its existing restart action via the session menu. Adding chrome inside the terminal area felt like premature decoration; can iterate from real use.
- **No streaming.** The broker is GET with `Content-Length`; one `arrayBuffer()` is enough for the 2 MiB max payload.
- **No `gmux --tail` change.** That was listed as optional in the lifecycle plan; orthogonal to the web piece.

## Tests

`apps/gmux-web/src/replay-fetch.test.ts`: 6 unit tests for the network helper. Verified the assertions bite by perturbing the production code (200/empty case returned `bytes` instead of `empty` ➝ test failed as expected, then restored).

The component itself is intentionally thin: it composes xterm + the tested fetch helper. Component-level rendering tests would require a DOM-test setup that doesn't yet exist in this repo. Manual install verification + the existing E2E harness give better signal here than a contrived component unit test.

## Manual verification

Will install the prerelease and verify on real dead sessions before merging. Plan:
1. `gmuxd restart`, click a dead pi/claude session in the sidebar, expect to see its actual scrollback in a read-only terminal that scrolls to the latest output.
2. Click a pre-S3 dead session (the ones surfaced by the convIndex fallback), expect the inline "No scrollback was captured" notice.
3. Resize the window while viewing replay, expect xterm to reflow without errors.
4. Switch between dead sessions, expect each to fetch its own scrollback and not bleed bytes from the previous one.

### DIFF
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -5,6 +5,7 @@ import '@xterm/xterm/css/xterm.css'
 import './styles.css'
 
 import { TerminalView } from './terminal'
+import { ReplayView } from './replay-view'
 import { useArrivalPulse } from './use-arrival-pulse'
 import { Sidebar } from './sidebar'
 import type { DotState } from './store'
@@ -517,9 +518,17 @@ function App() {
             onPasteReady={handleTerminalPasteReady}
             onFocusReady={handleTerminalFocusReady}
           />
+        ) : selectedVal && !selectedVal.alive && termOpts && !USE_MOCK ? (
+          <ReplayView
+            session={selectedVal}
+            terminalOptions={termOpts}
+            onResume={handleResume}
+            onDismiss={handleCloseSession}
+            resuming={resumingId === selectedVal.id}
+          />
         ) : selectedVal ? (
           <div class="state-message">
-            <div class="state-subtitle">{selectedVal.alive ? 'Connecting…' : 'Session ended'}</div>
+            <div class="state-subtitle">Connecting…</div>
           </div>
         ) : (
           <Home />
--- a/apps/gmux-web/src/project-hub.tsx
+++ b/apps/gmux-web/src/project-hub.tsx
@@ -150,12 +150,6 @@ function SessionCard({
     <a
       class="session-card"
       href={href}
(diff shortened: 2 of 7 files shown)


mgabor3141:
Superseded by #209, which ports the same changes onto the resume-id-passthrough stack (#208). The backend fix in #208 is what makes the broker's scrollback bytes match the runner's writes after Resume; without it, ReplayView would show the wrong directory's contents (the original symptom that motivated this work).

PR #208, opened 2026-05-06
feat(web): replay scrollback for dead sessions

mgabor3141:
Resume by Session.ID passthrough (ADR 0003), the visible UX that makes the new flow reachable through the browser (originally PR #209, merged in), and a tail of correctness fixes uncovered during verification.

## What's in here

Two ADRs and 14 implementation commits, all atomic, all conventional.

### Design records

- **`docs(adr): 0003`** — Daemon-initiated resume passes `Session.ID` to the runner via a dedicated `GMUX_RESUME_ID` env var. Argues against the prior match-by-command merge, against a `RunnerID` field, and lays out the collision-handling story.
- **`docs(adr): 0004`** — `SessionStream` and the live↔dead view abstraction. The frontend follow-up that 0003 unblocks; not implemented in this PR but landed as the design of record.
- **`docs(adr): align 0003 with shipping field-preservation and child-env strip`** — refreshes 0003 to match what actually shipped: full field-preservation list, plus the explicit `buildChildEnv` strip that makes "GMUX_RESUME_ID is never inherited" true.

### Backend implementation (id-passthrough)

- **`feat(cli): adopt session id from GMUX_RESUME_ID env var`** — runner reads `GMUX_RESUME_ID` and uses it as its session id when set. Distinct from `GMUX_SESSION_ID` so nested-gmux invocations don't accidentally re-bind the parent session's id.
- **`feat(daemon): pass session id to resumed runner via GMUX_RESUME_ID`** — `/v1/resume` and `/v1/restart` set the env var on the forked runner. `/v1/launch` continues to leave the runner to generate a fresh id.
- **`refactor(cli): bind session socket before sessionID-dependent setup`** — hoists the bind to before scrollback / env / state setup; on collision with a live runner the runner regenerates its id before any per-id setup happens. `ptyserver.New` now takes a pre-bound `Listener`.
- **`refactor(discovery): collapse resume merge into single Register path`** — `Register` now has one branch (re-registration vs fresh) instead of three. `pendingResumes`, `OnResumeMergedFunc`, and `resume.go` all delete.
- **`refactor(cli): move parent-dir mkdir into BindSocket`** — small ergonomics cleanup so callers stop having to remember `os.MkdirAll(filepath.Dir(...))`; the function now owns the entire socket-path setup.

### Bug fixes uncovered during implementation

- **`fix(scrollback): drop rotated previous file on Open so resume gets a clean slate`** — under id-passthrough, a resumed runner reuses the prior runner's directory. Without this, `scrollback.0` from the dead run would splice into the new run's readback.
- **`fix(cli, daemon): release session socket path during /kill; trust SSE subscription as liveness signal in Scan`** — found via UI testing: clicking Restart was producing a surprise sibling session in the sidebar. Two coupled races, fixed together. **Race 1**: `/kill` returned 204 before the runner's listener was torn down, so a replacement runner's `BindSocket` collided with the dying listener and fell back to a fresh id. Fixed by `handleKill` `os.Remove`ing the canonical sockfile path right before responding 204; the listener stays alive on the orphan inode for SSE/WS drains. **Race 2**: the early unlink opened a window where Discovery's phase-2 death-detection saw the missing path while the SSE was still streaming the runner's exit event, mis-classified as a crash, and called `NotifySessionDied` — which clears the file→session attribution that resume / restart need to preserve. Fixed by treating the active SSE subscription as the authoritative liveness signal in phase 2; the path / probe checks are only consulted when the subscription drops.
- **`fix(daemon): preserve attribution and history fields across re-registration`** — caught in self-review after the `Register` collapse landed: the rewrite preserved only `Slug`, dropping `CreatedAt`, `AdapterTitle`, `Subtitle`, `WorkspaceRoot`, and `Remotes` on every resume. Restored the field-by-field overwrite shape (mutate `existing` in place, copy only runtime-authoritative fields from the runner's `/meta` payload), keyed off id-collision.
- **`fix(cli): strip GMUX_RESUME_ID from PTY child env to prevent inheritance`** — also caught in self-review: `buildChildEnv` was passing parent env through unchanged, so `GMUX_RESUME_ID` leaked into PTY children, and a nested `gmux foo` inside a session would try to re-bind the parent's id and depend on the collision fallback. Strip at the runner→child boundary; `GMUX_SESSION_ID` is intentionally retained because adapters / hooks consume it.
- **`fix(daemon): replace existing SSE subscription on Subscribe so /restart cannot lose its event stream`** — caught in a deeper verification pass after the rest of the stack was already green. `Subscriptions.Subscribe` was a no-op when the active map already held an entry for the session id. During /restart, between the dying R1's SSE still being live and the fresh R2 calling `/v1/register`, R2's Subscribe could find the slot occupied, return without subscribing, and end up with no event stream. The store treated R2 as alive but status / meta / exit events were silently dropped until the next discovery scan. Switched to cancel-and-replace semantics: the active map now stores a `*subEntry` sentinel, the cleanup defer in `runSubscription` compares by pointer identity and skips the delete when its entry has been replaced, so a dying goroutine cannot trample a freshly-installed replacement. Subscribe holds `sub.mu` across cancel-and-overwrite so the old defer serializes after the swap. Logs `"replacing existing subscription"` when the path is exercised, which keeps the race window diagnosable in production.

### Frontend (originally PR #209, merged in)

- **`feat(web): replay scrollback for dead sessions`**
  - `replay-view.tsx` (new) — read-only xterm that fetches `/v1/sessions/<id>/scrollback`, replays the captured bytes, and offers Resume / Rerun in an action bar.
  - `replay-fetch.ts` + `replay-fetch.test.ts` (new) — small typed wrapper around the broker endpoint with branches for bytes / empty / 404 / error.
  - `main.tsx` — routes `selectedVal && !alive` to `ReplayView`; live sessions still go through `TerminalView`.
  - `sidebar.tsx` + `project-hub.tsx` — drop the auto-resume-on-sidebar-click handler. Selection (navigate to view) and resumption (explicit button) are now separate.
  - `styles.css` — replay layout (terminal fills, action bar pinned).
- **`fix(web): guard replay against disposed terminal and aborted response body`** — Greptile review:
  - `term.write`'s callback fires async; between scheduling and callback running, the effect's cleanup may have disposed the terminal. Guarded with the `cancelled` flag.
  - `arrayBuffer()` was outside the try/catch; a body-stream abort after headers arrived would reject the `fetchScrollback` promise and the UI would silently stall. Wrapped.
  - Bundles a lint cleanup: dropped `onResume` threading through `Sidebar` / `ProjectHub` (now exclusively driven from `ReplayView`'s button) and the unused `sleeping` const.
- **`feat(web): jump-to-bottom button, drop redundant Dismiss, and label by adapter kind`**
  - New `jump-to-bottom.tsx` — self-contained `<JumpToBottom term={...} />` overlay component that subscribes to `term.onScroll`, hides when the viewport sits at the buffer baseline, renders a floating ↓ button otherwise. Used by `ReplayView`; `TerminalView` can adopt it later as a one-line addition without duplicating logic.
  - Drops the action-bar Dismiss button: the sidebar's per-session close is the single way to remove a dead session. Two affordances with different verbs (X vs Dismiss) for the same operation was ambiguous.
  - Labels the resume button by adapter kind: agent adapters (`claude`, `codex`, `pi`) say "Resume" because their runners have an explicit resume protocol (`--resume <id>`); shells and one-off commands say "Rerun" because there's no agent state to resume from. Allowlist is explicit so unknown kinds default to the safe "Rerun" label.

## What this fixes (the original report)

The "no scrollback was captured" report after Resume: the runner used to generate its own native id at startup, write its scrollback to `sessions/<runner-native-id>/scrollback`, and rely on the daemon's `pendingResumes` merge to reconcile that against the canonical session id. The broker, keying on the canonical id, would read from a different directory than the runner wrote to. With id-passthrough, the runner uses the canonical id from the start; one directory per session, structurally. The frontend then surfaces those bytes via `ReplayView` so dead sessions are no longer a "Session ended" placeholder.

## What this does not touch

- Wire protocol (HTTP, SSE, WS): unchanged.
- ADR 0004's deeper architecture (`SessionStream`, pooled byte buffers, replacing `TerminalView`'s WS handling, seamless live↔dead transitions) is **not** implemented; the ADR stays as design-of-record. `TerminalView` is 1037 lines of intricate WS / resize / scroll / paste / mobile-input / image-protocol logic; the seamless-flip and image-state-preserving wins are real but quality-of-life. We can pick up 0004 when there's a concrete motivating scenario.
- `TerminalView` does not yet adopt `JumpToBottom`; the component is structured for that one-line addition when desired.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` clean across all `go.work` modules.
- 310 frontend tests (`vitest run`), `tsc --noEmit` clean.
- New tests, all verified to fail when their corresponding fix is reverted:
  - `TestBindSocketStaleFile` / `TestBindSocketLiveOwnerLeftIntact` / `TestBindSocketCreatesParentDir` (`ptyserver`)
  - `TestRegisterReRegistrationPreservesPersistedSlug` / `TestRegisterReRegistrationPreservesAttributionAndHistory` / `TestRegisterFreshSessionRunsOnRegisterForShell` / `TestScanIgnoresMissingPathWhileSubscriptionAlive` (`discovery`)
  - `TestKillReleasesSocketPathBeforeResponding` / `TestBuildChildEnv_StripsResumeID` (`ptyserver`)
  - `TestOpenClearsPreviousRotatedFile` (`scrollback`)
  - `TestSubscribeReplacesExistingForSameID` / `TestSubscribeOldDeferDoesNotEvictNewEntry` (`discovery`, race detector clean across 200x stress runs)
  - 7 cases in `replay-fetch.test.ts` covering bytes / empty / 404 / 5xx / network failure / arrayBuffer-rejection / URL encoding.
- End-to-end: the original "no scrollback was captured" case (sess-93db1e98) now returns 1423 bytes from the broker after Resume, matching the runner's writes. Sibling-session-on-Restart no longer reproduces. Restart preserves attribution.

## Closes

#206, #209 (#209's commits live on this branch directly).


mgabor3141:
@greptile

mgabor3141:
@greptile

mgabor3141:
@greptile

mgabor3141:
@greptile

PR #209, opened 2026-05-06
feat(web): replay scrollback for dead sessions (stacked on #208)

mgabor3141:
**Stacked on #208.** Lifts the visible UX changes from #206 onto the resume-id-passthrough stack so the full resume flow is reachable through the browser.

## What's in here

Two commits.

- **`feat(web): replay scrollback for dead sessions`**
  - **`replay-view.tsx`** (new) — read-only xterm that fetches `/v1/sessions/<id>/scrollback`, replays the captured bytes, and offers Resume / Dismiss in an action bar.
  - **`replay-fetch.ts`** + **`replay-fetch.test.ts`** (new) — small typed wrapper around the broker endpoint with branches for bytes / empty / 404 / error.
  - **`main.tsx`** — routes `selectedVal && !alive` to `ReplayView`; live sessions still go through `TerminalView`.
  - **`sidebar.tsx`** + **`project-hub.tsx`** — drop the auto-resume-on-sidebar-click handler. Selection (navigate to view) and resumption (explicit button) are now separate.
  - **`styles.css`** — replay layout (terminal fills, action bar pinned).

- **`fix(web): guard replay against disposed terminal and aborted response body`** — caught by Greptile review:
  - `term.write`'s callback fires async; between scheduling the write and the callback running, the effect's cleanup may have run (component unmount or `session.id` switch) and disposed the terminal. Guarded with the `cancelled` flag.
  - `arrayBuffer()` sat outside the try/catch in `replay-fetch.ts`; a body-stream abort after headers arrived would reject the `fetchScrollback` promise and the UI would silently stall on "Loading scrollback…". Wrapped in try/catch.
  - Bundles a lint cleanup: dropped `onResume` threading through `Sidebar` / `ProjectHub` (now exclusively driven from `ReplayView`'s button) and the unused `sleeping` const.

## Why stacked instead of just rebasing #206

The id-passthrough work in #208 makes resume actually preserve scrollback at the canonical session directory (the mismatch documented in ADR 0003 is what produced the original "no scrollback was captured" report). Without #208, `ReplayView` would render against a broker that reads from one path while the runner writes to another. Stacking ensures the two ship coherent: the user clicks a dead session, sees its scrollback, clicks Resume, the runner reuses the same id, and the post-resume scrollback continues to be served correctly.

## What this defers

ADR 0004's deeper architecture — `SessionStream`, pooled byte buffers, replacing `TerminalView`'s WS handling, seamless live↔dead transitions — is **not** in this PR. The ADR stays in the repo as the design of record (committed in #208). Rationale: `TerminalView` is 1037 lines of intricate WS / resize / scroll / paste / mobile-input / image-protocol logic; the seamless-flip and image-state-preserving wins are real but quality-of-life, with no specific user-reported bug that's their unique fix. We can pick up 0004 when there's a concrete motivating scenario or a slack moment.

`terminalOptions` deliberately omitted from `ReplayView`'s `useEffect` deps array (Greptile flagged): adding it would cause remount-storms unless the parent memoizes the options object, and there's no concrete need (theme/font preferences don't change while staring at a dead session, and switching unblocks naturally on next session selection).

## Verification

- All 310 frontend tests pass (`vitest run`), including the 7 `replay-fetch.test.ts` cases (one new for the arrayBuffer-rejection branch).
- `tsc --noEmit` clean.
- Backend story (PR #208) verified end-to-end against `sess-93db1e98`: Resume → runner adopts canonical id → 1423 bytes at the canonical scrollback path → broker serves them. With this PR on top, that flow is reachable through the UI.

## Closes

#206 (superseded by this stacked port + the id-passthrough fix it depends on)

### DIFF
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -461,7 +461,6 @@ function App() {
     <div class="app-layout">
       <Sidebar
         resumingId={resumingId}
-        onResume={handleResume}
         onCloseSession={handleCloseSession}
         onManageProjects={() => { setSidebarOpen(false); setManageProjectsOpen(true) }}
         open={sidebarOpen}
@@ -501,7 +500,6 @@ function App() {
         ) : viewVal?.kind === 'project' ? (
           <ProjectHub
             projectSlug={viewVal.projectSlug}
-            onResume={handleResume}
             onCloseSession={handleCloseSession}
           />
         ) : selectedVal && (canAttach || USE_MOCK) && termOpts && keybindsVal ? (
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -4,6 +4,7 @@ import { LocationProvider, Router, Route, lazy, useLocation } from 'preact-iso'
 import '@xterm/xterm/css/xterm.css'
 import './styles.css'
 
+import { ReplayView } from './replay-view'
 import { TerminalView } from './terminal'
 import { useArrivalPulse } from './use-arrival-pulse'
 import { Sidebar } from './sidebar'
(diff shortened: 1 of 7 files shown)


mgabor3141:
@greptile

mgabor3141:
@greptile

PR #212, opened 2026-05-07
feat: add gmux skill, improve gmux CLI for agents

mgabor3141:
Closes #171.

Three CLI changes that turn scripted / agent-driven gmux from "works if you know the shell trivia" into a first-class workflow, plus the docs and skill that make the pattern discoverable.

### `--send` submits by default

Sending a prompt to an agent used to require manually appending a carriage return:

```sh
{ cat prompt.txt; printf '\r'; } | gmux --send <id>
```

Without it, bytes landed in the agent's input box but nothing got dispatched. Heredocs, `cat`, and editor saves all produce `\n`; agents expect `\r` for submit. The failure mode was silent (no error, no output, agent just sits there) and the workaround required three pieces of shell trivia (brace-grouped subshells, `printf`, the `\r` vs `\n` distinction) for what is conceptually one operation: send a message.

`--send` now appends `\r` after the payload by default. `--no-submit` opts out for the rare flow where you want to pre-fill the input box without dispatching it. Matches what every other "send a message" CLI does (`gh issue comment`, `slack send`, `jira issue comment add`).

Backwards compatible for callers that already sent `\r`: a redundant `\r` submits an empty line at most.

### `--wait` blocks until idle

`gmux --wait <id>` blocks until the agent has finished its turn, replacing the previous `sleep + tail + parse` workaround for scripted orchestration. Pairs with `--send` for the canonical sequential-step shape:

```sh
gmux --send <id> < prompt.txt
gmux --wait <id>
gmux --tail 50 <id>
```

Optional `--timeout N` caps the wait at N seconds. No default timeout: shell `timeout` handles caps when callers want one, and a non-zero default would surprise long-running orchestration loops.

The wait runs server-side at `POST /v1/sessions/{id}/wait` so the idle-detection rules (which adapter kinds emit `Status.Working`, what counts as "died") live in one place. The CLI maps the daemon's response to distinct exit codes:

- 0: agent reached idle (`Status.Working == false`)
- 2: session died before becoming idle
- 3: `--timeout` elapsed
- 1: usage / network / other errors

Shell sessions don't emit a working signal and are rejected with 422 (`no_idle_signal`) rather than silently returning idle, which would foot-trap the obvious composition `gmux make build && gmux --wait <id>` into doing nothing. To wait for a shell command to finish, use the existing piped flow: `gmux make build < /dev/null && deploy.sh`.

### Flags accepted in any order in management modes

Run mode (`gmux <cmd> [args]`) needs POSIX runner flag parsing — flags after the wrapped command must reach the child, not gmux. But the same rule applied to management modes (`--list`, `--attach`, `--kill`, `--send`, `--wait`, `--tail`) was a silent foot-trap: `gmux --wait <id> --timeout 60` looked correct and errored with `--wait requires a session id` because `--timeout 60` had become extra positionals. This pattern was reported three times during review.

Now: management modes parse flags in any order; run mode keeps strict POSIX semantics. `--` is honoured as a terminator so `gmux --send <id> -- --no-submit` sends the literal text `--no-submit` instead of consuming it as a flag.

```sh
# all four work
gmux --wait --timeout 60 <id>
gmux --wait <id> --timeout 60
gmux --send --no-submit <id> 'draft '
gmux --send <id> 'draft ' --no-submit
```

## Docs (the original #171 scope)

- **`reference/cli.md`** updated: replaces the two-paragraph attach / `--no-attach` description with a three-row table mapping stdin (TTY vs pipe) and `GMUX` env state to the actual runtime behavior. The piped / non-tty row was previously undocumented even though it is the foundation of every script and agent harness that uses gmux. The exit-code propagation contract was never stated anywhere; now it is.
- **`integrations/scripts-and-agents.md`** new: the narrative companion. Canonical `gmux <cmd> < /dev/null` shape, `--send` / `--wait` / `--tail` worked together, parallel-agent orchestration recipe, nested-gmux behavior with the actual stdin-is-a-tty gate so agent harnesses know they don't need special-casing.
- **`reference/cli.md`** `--send` and `--wait` sections updated for the new flags and exit-code contract.

## Skill

`skills/gmux/SKILL.md` follows the [agentskills.io](https://agentskills.io/) specification, so users can install it via:

```sh
gh skill install gmuxapp/gmux gmux                    # any host
gh skill install gmuxapp/gmux gmux --agent claude-code
gh skill install gmuxapp/gmux gmux --agent codex
```

Lives in this repo rather than a separate one so the skill version tracks the gmux release that introduced its primitives — `gh skill install` resolves to the latest tagged release for free, and the skill is never out of sync with the CLI flags it documents.

The skill is deliberately concise: agents are smart enough to extrapolate from the four primitives (`<cmd> < /dev/null`, `--send`, `--wait`, `--tail`) and the worked composition patterns. The triggering description front-loads the use cases ("background command", "send input to a session", "wait for an agent's turn", "orchestrate multiple agents in parallel") so implicit invocation matches naturally when the user describes their problem rather than the tool.

## Verification

- New tests, all verified to fail when their corresponding fix is reverted:
  - `TestBuildSendBody` — pins the wire-level `--send` contract: by default the bytes end with `\r`, `--no-submit` suppresses exactly that byte, both inline-text and stdin paths covered.
  - `TestParseCLI` / `TestParseCLIErrors` — extends the existing parser tests with `--no-submit`, `--wait`, `--timeout` invariants and the mutual-exclusion / requires-`--send` / requires-`--wait` rules.
  - `TestWaitReturnsImmediatelyWhenAlreadyIdle` — `gmux --send X && gmux --wait X` works when the agent races ahead between the two CLI hops.
  - `TestWaitBlocksUntilWorkingFlipsFalse` — the headline behavior: drives the store asynchronously so the test exercises the subscription path, not the initial-snapshot fast path.
  - `TestWaitReturnsDiedWhenSessionDies` — death mid-turn returns `reason=died` so the CLI maps to exit code 2.
  - `TestWaitTimesOut` — distinct 408 / exit code 3 so callers can tell timeout apart from idle / died.
  - `TestWaitRejectsShellSessions` — explicit allowlist (claude / codex / pi) catches future kinds that forget to wire `Working` too.
  - `TestWaitNotFound` — keeps "missing session" from being confused with "session is busy forever".
  - `TestWaitDoesNotTreatNilStatusAsIdle` — adapter-not-yet-ready (Status nil) waits for a real state instead of returning idle immediately.
  - `TestWaitUnderHighEventLoad` — 500ms ticker fallback rescues us when the 64-slot subscriber buffer drops the critical Working→false event.
  - `TestWaitReturnsDiedWhenSessionRemoved` — UI dismiss (or any `sessions.Remove()` call) during a wait returns `reason=died` instead of hanging forever.
  - `TestRunModeKeepsPOSIXRunnerSemantics` — pins that lenient management-mode parsing didn't break the run-mode contract: `gmux pi --pi-flag prompt` still forwards `--pi-flag` to pi.
  - `TestParseCLI` adds three cases for the new flexibility (`--wait <id> --timeout`, `--send <id> --no-submit text`, `--send <id> text --no-submit`) plus one for `--` flag-termination of inline text.
- Backend tests under `-race -count=20` clean.
- End-to-end smoke against running agents: `--wait --timeout 5` returns 0 against idle pi sessions, 3 against busy ones, 1 (with clear message) against shell and remote-peer sessions.
- `npx astro build` clean across all 40 pages.

## Out of scope (deliberately, per the issue's prioritisation)

- `--wait-all`: composes via shell `&` + `wait`. Worked example in the integrations page.
- `--send-wait`: composes via `&&`.
- `--status-all`, `--history`, `--wait-condition`: deferred. The first three primitives close the orchestration gap; the rest is sugar.
- Remote peer sessions: rejected with a clear error until peer subscriptions stream `Status` events back to the hub.

### DIFF
--- a/apps/website/src/content/docs/integrations/scripts-and-agents.md
+++ b/apps/website/src/content/docs/integrations/scripts-and-agents.md
@@ -1,0 +1,132 @@
+---
+title: Scripts and agents
+description: Drive gmux sessions from shell scripts, CI, and agent harnesses.
+sidebar:
+  order: 0
+---
+
+`gmux` was designed so that the same binary works whether you're attaching to a session by hand or driving sessions from a script. This page covers the scripted shape: how to start sessions non-interactively, how to send input and wait for output, and how to compose the primitives into agent-orchestration patterns.
+
+:::tip[Driving gmux from an agent?]
+Install the [gmux skill](https://github.com/gmuxapp/gmux/blob/main/skills/gmux/SKILL.md) so your agent picks up these patterns automatically:
+
+```sh
+npx skills add gmuxapp/gmux
+```
+
+The skill follows the [agentskills.io](https://agentskills.io/) standard and works with Claude Code, Codex, Cursor, Copilot, Gemini CLI, OpenCode, and 50+ other agents. Or drop the `SKILL.md` into your agent's skills directory by hand if you prefer not to install the CLI.
+:::
+
+## The piped flow
+
+The most useful primitive for scripting is `gmux <cmd>` with stdin redirected away from a terminal:
+
+```bash
+gmux make build < /dev/null
+gmux pi -p "summarize this PR" < /dev/null
+```
+
+The `-p` (print) flag tells `pi` to process the prompt and exit instead of staying interactive. Other agents have similar one-shot modes (`claude -p`, `codex exec`); without one, the agent stays running and `gmux <cmd>` blocks indefinitely. For multi-turn orchestration, see the [parallel orchestration](#parallel-orchestration) section below: spawn with `--no-attach`, then drive with `--send` / `--wait`.
+
+When stdin is not a TTY, `gmux <cmd>`:
+
+- **Blocks** until the child exits.
+- **Streams bounded metadata** to stdout (session id, kind, exit status), not the full PTY output. Your script's logs stay readable.
+- **Exits with the child's exit code**, so `gmux make build < /dev/null && deploy.sh` works.
+- **Keeps the session in the UI** for the duration: a human can watch it live in the browser without affecting the script.
+
+This is the shape every other line on this page builds on. It works the same in CI, in cron jobs, in agent harnesses (whose stdin is a pipe by default), and in any scripted invocation.
+
+## Sending input
+
+Use [`--send`](/reference/cli/#gmux---send---no-submit-id-text) to push input into an already-running session, as if the bytes had been typed at the keyboard. By default `--send` submits the input (appends the carriage return that signals Enter), so the canonical shape is just:
+
+```bash
+gmux --send <id> < prompt.txt
+gmux --send <id> 'shorter inline message'
+```
+
+Use `--no-submit` for the rare case where you want to pre-fill the input box without dispatching, e.g. agent-assisted human authoring or sending a control character without an extra Enter:
+
+```bash
+printf '\x03' | gmux --send --no-submit <id>   # Ctrl-C, no extra Enter
+gmux --send --no-submit <id> 'draft '          # leave "draft " in the input
+```
+
+`--send` is local-only and gated by Unix-socket file permissions (owner-only `0700`); see the CLI reference for the access-control story.
+
+## Waiting for the turn to finish
+
+`gmux --wait <id>` blocks until an agent session has finished its current turn. It's the primitive that turns sequential orchestration into a one-line-per-step pattern:
+
+```bash
+gmux --send <id> < step-1.txt
+gmux --wait <id>
+
+gmux --send <id> < step-2.txt
+gmux --wait <id>
+
+# extract the final answer
+gmux --tail 200 <id>
+```
+
+The idle signal is the same `Status.Working` flag the UI's spinner consumes, so `--wait` returns the moment the agent emits its closing message for the turn. If the session dies first, `--wait` exits 2; if you set `--timeout N` and N seconds pass, it exits 3. Idle is exit 0.
+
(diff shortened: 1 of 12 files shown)


mgabor3141:
Greptile review followups (both squashed into the relevant existing commits, no new commits):

**P1 — `session-remove` events silently ignored**: real bug. `sessions.Remove()` (UI dismiss path) broadcasts an event with `Session == nil` and the periodic ticker also no-ops on missing sessions, so `gmux --wait $id` against a dismissed session would hang forever. Fixed in `feat(daemon): add /v1/sessions/{id}/wait endpoint`: the event-loop now treats `ev.Session == nil` for the watched id as a terminal `reason=died`, and the ticker treats `sessions.Get` returning `!ok` the same way. Verified by adding `TestWaitReturnsDiedWhenSessionRemoved` and confirming it hangs to timeout without the fix.

**Nit — `Content-Type: application/json` with nil body**: switched to `client.Post(endpoint, "", http.NoBody)` in `feat(cli): add --wait`.

mgabor3141:
@greptile please re-review. Since your initial pass:

- **P1 session-remove silent hang**: fixed. `handleWait` now returns `reason=died` on `ev.Session == nil` (session-remove broadcast) and on `sessions.Get` returning `!ok` from the periodic ticker. Pinned by `TestWaitReturnsDiedWhenSessionRemoved` (verified to hang-to-timeout against the unfixed code).
- **Nit Content-Type with nil body**: fixed. `client.Post(endpoint, "", http.NoBody)`.
- **New: lenient flag parsing for management modes** (commit `feat(cli): allow flags after positional in management modes`). `gmux --wait <id> --timeout 60` and friends now work; run mode keeps strict POSIX so `gmux pi --pi-flag prompt` still forwards verbatim. `--` is honoured as a terminator (`gmux --send <id> -- --no-submit` sends literal text). New tests: `TestRunModeKeepsPOSIXRunnerSemantics`, three `TestParseCLI` cases for flag-after-positional in management modes, and one for the `--` escape.

PR #215, opened 2026-05-08
feat(web): replace mobile-input handler with EditContext addon

mgabor3141:
Replaces the mobile-input.ts patch stack with [`xterm-addon-edit-context`](https://github.com/gmuxapp/xterm-addon-edit-context), which uses the [EditContext API](https://w3c.github.io/edit-context/) to give the terminal explicit text-replacement ranges instead of having to diff a hidden textarea.

## Why

xterm.js' default input path treats its hidden textarea as a keystroke buffer. Mobile virtual keyboards aren't keystroke generators; they're text editors that replace words (autocorrect), insert phrases (dictation), and manipulate cursor position (iOS spacebar trackpad). The textarea model can't distinguish these operations from normal typing, which is why we accumulated:

- iOS `insertReplacementText` autocorrect detection in `mobile-input.ts`
- Android `deleteContentBackward` + `insertText` pair detection
- The `(pointer: coarse)` guard to keep that handler off desktop
- A composition shrink fix in our xterm fork for Gboard "edit previous word"

EditContext fires `textupdate` events with explicit replacement ranges ("replace characters 3-7 with 'hello'"). No diffing, no race conditions, no platform-specific signal detection.

## What changes

- **Adds** `xterm-addon-edit-context@v0.1.0` (gmuxapp fork) as a dependency. The addon transitively pulls in `@neftaly/editcontext-polyfill` for Safari/Firefox, dynamically imported (~28 KB chunk, only loaded when native EditContext is unavailable).
- **Loads** `new EditContextAddon()` before `term.open()` so its capture-phase suppress listeners are wired before xterm's textarea handlers fire.
- **Deletes** `apps/gmux-web/src/mobile-input.ts` (199 lines) and `mobile-input.test.ts` (517 lines, 16 tests).
- **Removes** the `attachMobileInputHandler` call site and its dispose hook.

The composition shrink fix in the xterm fork is now dead code (CompositionHelper isn't reached) but stays on the `gmux` branch as an upstream-PR seed.

## Browser support

| Browser | Path |
|---|---|
| Chrome / Edge 121+ | Native EditContext |
| Chrome for Android | Native EditContext |
| Samsung Internet 25+ | Native EditContext |
| Safari 15.4+ (incl. iOS) | Polyfill |
| Firefox 125+ | Polyfill |

## Testing

- 296/296 web unit tests pass (was 312; the 16 mobile-input tests are deleted with the file).
- The addon itself has 23/23 unit tests in its own repo covering the diff algorithm and key classification.
- Build succeeds; polyfill correctly code-splits into a separate chunk.

## Caveats / known limits

From the addon's README + source comments:

- **Replacement encoding is best-effort for raw-mode programs.** Mid-line word replacements emit cursor-to-end backspaces + retype-the-suffix, which works for readline (bash/zsh), Claude Code's input box, vim insert mode, and most TUIs. Programs that intercept arrow keys for non-cursor purposes will see corrupted input on autocorrect.
- **No `updateText()` line-sync wired in this PR.** That means hold-backspace word deletion and the iOS spacebar-as-trackpad gesture won't work yet (the OS doesn't know what's on the current line). Basic dictation, autocorrect, and swipe typing don't need it. Plumbing line-sync requires deciding what counts as "the current input line" given the PTY just streams bytes; deferred to a follow-up.
- **Real-device validation pending.** The addon has solid unit-test coverage of its diff algorithm but has never been exercised on a real iPhone/Android device. Expect to file follow-up issues; the previous mobile-input.ts behavior is still recoverable from git history if we need to revert.

Refs: `gmuxapp/xterm-addon-edit-context` (v0.1.0)


mgabor3141:
Note: this is pretty broken, approach will likely be abandoned

PR #229, opened 2026-05-24
feat(daemon): stamp sessions with last_activity_at on noteworthy state transitions

mgabor3141:
## Summary

Adds `last_activity_at` to the Session model: an RFC3339 timestamp set by the owning daemon whenever a session goes through a noteworthy state transition. Frontend will consume this in upcoming PRs to power the home dashboard's "Recent" section and to provide a stable activity sort key across views.

This PR is wire/protocol/store plumbing only; no UI consumer yet.

## What bumps the timestamp

Exactly these prev→next transitions, computed in `shouldBumpActivity`:

- `alive: true → false` (the session exited)
- `unread: false → true` (new output the user hasn't seen)
- `status.working: false → true` (adapter started a task)
- `status.error: false → true` (status went into error)

**Not** bumped on: title/slug/cwd/stamps changes, status label-only updates, brand-new session creation, or sessionmeta rehydrate at daemon startup. Brand-new sessions arrive with the field unset; their first follow-up transition stamps it. This keeps daemon restart from timestamping every resumable-dead session to "now".

## Architecture

- `Upsert` and `Update` invoke the bump check; **`UpsertRemote` does not.** Peer sessions arrive over the SSE forwarding with the stamp already set by the owning daemon (per ADR 0002). The hub forwards as-received; no hub-side observer of peer transitions, no risk of double-stamping.
- The bump check operates on a pre-captured `activitySnapshot` (just the four scalar bits: `alive`, `unread`, `working`, `errored`). This is load-bearing: `Update` does `sess := prev`, a shallow copy that leaves `prev.Status` and `sess.Status` pointing at the same `Status` struct. A mutator that wrote through that pointer (`sess.Status.Working = true`) would silently mutate `prev` too, hiding the transition from the comparison. Snapshotting the booleans before `fn` runs sidesteps the alias entirely.
- Field is on `store.Session` directly, so `sessionmeta`'s existing `persistedSession` type alias picks it up automatically. Resumable sessions retain their last-activity timestamp across daemon restart with no new persistence code.
- Session has a custom `MarshalJSON` (an explicit `wire` struct, not default tag reflection). Adding a field to `Session` with a json tag is **silently insufficient**: it has to be enumerated in the `wire` literal too, or it never reaches any client. A new reflection-based test enforces this contract for every future field.

## Commits

1. **`feat(daemon): stamp sessions with last_activity_at on noteworthy state transitions`** — adds the field, the bump helper, the bump-on-(Upsert|Update) call sites, protocol/Zod wiring, and unit tests for each transition.

2. **`fix(daemon): include last_activity_at in Session wire MarshalJSON; harden bump check against pointer aliasing`** — adds the missing `wire` struct entry, refactors the bump check to operate on a pre-captured `activitySnapshot`, and adds three classes of regression tests: pointer-aliasing safety, wire serialization presence + omitempty, and the all-fields-on-wire reflection test that catches "added a json tag but forgot the wire literal" for any future field.

## Test coverage

- Bump-on-each-transition (exited / unread / working / error).
- No-bump-on-noise (title / slug / cwd / status-label).
- No re-bump while staying in the same state.
- New-session creation does not stamp.
- Peer payloads through `UpsertRemote` preserve the wire-supplied timestamp unchanged, including on alive→false transitions.
- **Pointer aliasing**: a mutator that writes `sess.Status.Working = true` through the existing pointer still bumps. Would have failed under the first draft.
- **Wire serialization**: `last_activity_at` appears on the wire when set, omitted when unset.
- **Wire round-trip**: a session marshaled and unmarshaled preserves the timestamp.
- **All-fields-on-wire (reflection)**: every json-tagged field on `Session` must either appear in `MarshalJSON`'s output or be listed in an explicit `internalSessionFields` exclusion set. New fields fail loudly, with an error message that tells the author exactly which list to update.
- **Sessionmeta round-trip**: the existing `TestRoundTripFullSession` now populates `LastActivityAt`, so a future drop of the json tag would surface as a `reflect.DeepEqual` diff.
- **Frontend boundary**: `toUISession` passes through both the present and absent cases.

## Follow-ups

Subsequent atomic PRs on the same dashboard track:

1. `·host` suffix on project headers + kill the HOSTS section on `/`.
2. Wide `<SessionRow>` component shared by home dashboard and project page.
3. Home `/` as activity-section dashboard (Needs attention / Running / Recent) + projects list + "+ Add project" modal.
4. Project page rewrite using the same three-section pattern + adaptive cwd subtitle.

### DIFF
--- a/apps/gmux-web/src/store.test.ts
+++ b/apps/gmux-web/src/store.test.ts
@@ -77,6 +77,22 @@ describe('toUISession project stamp passthrough', () => {
     expect(ui.project_slug).toBeUndefined()
   })
 
+  it('passes last_activity_at through from the wire', () => {
+    // The owning daemon stamps this; the UI uses it for the home
+    // dashboard's Recent section sort. Pure passthrough at the
+    // boundary; no client-side derivation.
+    const ui = toUISession({
+      id: 'sess-1', alive: true,
+      last_activity_at: '2026-01-15T08:00:00Z',
+    } as any)
+    expect(ui.last_activity_at).toBe('2026-01-15T08:00:00Z')
+  })
+
+  it('leaves last_activity_at undefined when the wire omits it', () => {
+    const ui = toUISession({ id: 'sess-1', alive: true } as any)
+    expect(ui.last_activity_at).toBeUndefined()
+  })
+
   it('treats empty-string project_slug as unstamped', () => {
     // Go's omitempty drops empty strings, but legacy / dev paths may
     // emit them. buildProjectFolders treats an empty stamp the same
--- a/apps/gmux-web/src/store.ts
+++ b/apps/gmux-web/src/store.ts
@@ -479,6 +479,7 @@ export function toUISession(s: ProtocolSession): Session {
     status: s.status ?? null,
     unread: s.unread ?? false,
     resumable: s.resumable ?? false,
+    last_activity_at: s.last_activity_at ?? undefined,
     socket_path: s.socket_path ?? '',
     terminal_cols: s.terminal_cols ?? undefined,
     terminal_rows: s.terminal_rows ?? undefined,
--- a/apps/gmux-web/src/types.ts
+++ b/apps/gmux-web/src/types.ts
@@ -28,6 +28,14 @@ export interface Session {
   status: SessionStatus | null
   unread: boolean
   resumable?: boolean
+  /**
+   * RFC3339 timestamp of the session's most recent noteworthy state
+   * transition (exited / unread on / working on / error on). Set by
+   * the owning daemon. Drives the home dashboard's "Recent" section
+   * and acts as a stable sort key for activity-ordered views.
+   * Unset for brand-new sessions until their first transition.
+   */
+  last_activity_at?: string
   socket_path: string
   terminal_cols?: number
   terminal_rows?: number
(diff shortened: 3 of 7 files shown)


mgabor3141:
Addressed Greptile P2 (Upsert bump path uncovered): added `TestLastActivityAt_UpsertOnExistingBumps` exercising all four transitions through `Upsert` with `hadPrev=true`. The path now has the same coverage as `Update`.

mgabor3141:
Greptile flagged this in the #235 review (correctly identifying the daemon-side bug despite it living here): no-bump Upserts silently zeroed the previously stamped `LastActivityAt`. Adapters build fresh `Session` structs from runner state without `LastActivityAt`, so a routine title/cwd refresh would clobber the stamp and break Recent ordering.

`upsertCommon` now carries `prev.LastActivityAt` forward when `hadPrev` and the incoming payload's field is empty. New test `TestLastActivityAt_UpsertNoBumpPreservesStamp` pins the regression: seed via transition, Upsert with only an `AdapterTitle` change, assert the stamp survives.

The carry-forward only triggers when the payload's field is empty, so peer payloads (which legitimately carry their own stamp via `UpsertRemote`) are unaffected.

PR #231, opened 2026-05-24
feat(web): activity-first home dashboard

mgabor3141:
## Summary

Step 3 of the home/project/sidebar polish thread. Replaces the project-grid homepage with an activity-first dashboard, introduces a shared wide `<SessionRow>` for use across surfaces.

**Stacks on #230.** Merge order: #229 → #230 → this.

## What changes

- **New `<SessionRow>` component** (`session-row.tsx`). Two-line wide variant of the sidebar's `SessionItem`: title on line 1, optional `project · cwd · status · age` plus host suffix on line 2. Same dot conventions, sleeping/unavailable icons, hover-revealed `×` close. Deliberately a separate component from `SessionItem`: the sidebar variant is dense, drag-aware, and folder-scoped, this one is loose and standalone. Unifying behind a density prop would accumulate flags faster than it would save code.
- **Recency partitioning in `store.ts`** (`partitionForHome`). Pure function over a session list and a `now`, returns `{ needsAttention, running, recent }`. Section priority: `needsAttention > running > recent` (each session appears in at most one). Sort: newest-first by `last_activity_at`, falling back to `created_at` for sessions that have not transitioned yet. Recent shape: entries within `RECENT_WINDOW_MS` (1h), with floor `RECENT_FLOOR` (3) and cap `RECENT_CAP` (10).
- **Home dashboard** (`home.tsx`). Three activity sections at top (empty sections hidden), projects list below (each row: project name + host suffix + counts + launch `+`, click navigates), `+ Add project` footer button reusing the existing manage-projects modal unchanged.
- **`sessionDotState` lifted to `store.ts`** as the single source of truth for both the sidebar row and the new wide row.

## Why

The home page was a project picker that mostly duplicated what the sidebar already does. The dashboard answers a different question: *what across all my hosts needs my attention right now?* The activity sections cut to that without forcing the user to scan a project tree mentally.

Recency is sourced from `last_activity_at` (added in #229), bumped by the owning daemon on exactly: exited, unread on, working on, error on. A brand-new idle session has no timestamp and falls back to `created_at`, landing at the top of Running where it belongs.

## Tests

`apps/gmux-web/src/home-partition.test.ts` (12 tests):

- Section assignment: unread → needsAttention; error (no unread) → needsAttention; alive non-attention → running; dead → recent.
- Priority dedup: a dead-and-unread session goes to needsAttention, not recent.
- Sort: newest-first by `last_activity_at`, `created_at` fallback for never-transitioned sessions.
- Recent window: ≥floor inside window → ship only the in-window set; <floor inside → top-floor by recency overall.
- Recent cap: pathological-many recent transitions clip at `RECENT_CAP`.
- Empty leftover (all alive) → empty Recent.

## Out of scope (deferred)

- Project page rewrite using the same three-section pattern + adaptive cwd subtitle.
- Manage-projects modal rework (the `+ Add project` button opens it unchanged).
- Realtime tick signal to drop entries out of Recent as wall-clock advances. The floor handles the empty-section case; if Recent staleness becomes a felt issue we add a 30s ticker.

## Verification

- `pnpm lint` and `pnpm test` clean (406 passing, 12 new).
- Dev daemon visual check: section ordering correct, sleeping icon and `exited (N)` status render on Recent rows, project counts and launch `+` work, `+ Add project` opens the existing modal, hover reveals the `×` button.

### DIFF
--- a/apps/gmux-web/src/home-partition.test.ts
+++ b/apps/gmux-web/src/home-partition.test.ts
@@ -1,0 +1,187 @@
+// Pins the home dashboard's section partition. The partition is the
+// only piece of dashboard logic with non-trivial behavior (the row
+// component is visual, the section layout is presentation); these
+// tests guard the behavior contracts the design conversation pinned
+// down: section priority, sort key, and the Recent floor/window/cap.
+
+import { describe, it, expect } from 'vitest'
+import { partitionForHome, RECENT_WINDOW_MS, RECENT_FLOOR, RECENT_CAP } from './store'
+import { makeSession } from './test-helpers'
+
+const NOW = Date.parse('2026-06-01T12:00:00Z')
+
+function stamp(offsetMs: number): string {
+  return new Date(NOW + offsetMs).toISOString()
+}
+
+describe('partitionForHome', () => {
+  describe('section assignment', () => {
+    it('puts unread alive sessions in needsAttention', () => {
+      const s = makeSession({ id: 's1', cwd: '/x', alive: true, unread: true })
+      const { needsAttention, running, recent } = partitionForHome([s], NOW)
+      expect(needsAttention.map(s => s.id)).toEqual(['s1'])
+      expect(running).toEqual([])
+      expect(recent).toEqual([])
+    })
+
+    it('does NOT escalate error-only sessions to needsAttention', () => {
+      // status.error mostly fires from background subcommands that
+      // exited non-zero without the agent itself halting, so it
+      // doesn't earn the "needs attention" slot. The per-row error
+      // dot still surfaces the state. This test pins the predicate
+      // and prevents a regression to the old `unread || error`
+      // shape that lit up the section on every failed subcommand.
+      // We deliberately don't assert which fallthrough bucket the
+      // session lands in; that's covered by other tests.
+      const s = makeSession({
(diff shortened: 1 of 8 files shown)


mgabor3141:
Addressed Greptile feedback:

- **P2 `showHost` silently dropped** (`session-row.tsx`): meta-row render gate is now `metaSegments.length > 0 || showHost`, so a local-no-status-no-timestamp session with `showHost={true}` still gets its host suffix.
- **P1 Fragment keys**: replaced `<>` with `<Fragment key={i}>` (long-form fragments do accept `key`). Index is the honest key here because the content is positional, but at least Preact's diffing now operates on stable identities.

mgabor3141:
Addressed Greptile re-review:

- **`.home` → `.page` CSS rename moved into this commit.** The rename had been split between this commit (the JSX side) and a later one in the stack (the CSS side), which broke this PR in isolation. Both halves now land atomically.
- **HostSuffix leading separator dropped when meta is otherwise empty.** Added a `leading` prop (default `true`) to `<HostSuffix>`; `<SessionRow>` passes `leading={metaSegments.length > 0}`. So a local-no-status-no-timestamp session with `showHost={true}` now renders ` host` cleanly instead of ` · host`.

PR #232, opened 2026-05-24
feat(web): activity-first project page with adaptive cwd subtitle

mgabor3141:
## Summary

Step 4a of the home/project/sidebar polish thread. Rewrites the project hub page to mirror the home dashboard's three-section pattern (Needs attention / Running / Recent), filtered to one project. Drops worktree grouping in favor of an adaptive cwd subtitle.

**Stacks on #231.** Merge order: #229 → #230 → #231 → this.

## What changes

- **`project-hub.tsx` rewrite.** Same three sections as the home dashboard, scoped to the project's sessions. Renders `<SessionRow>` with `showProject={false}` (the project name is in the page header) and `showHost` only when sessions disagree on `.peer` (i.e. devcontainer case).
- **Adaptive cwd subtitle.** If every session in the project shares one cwd, render it as a monospace subtitle under the title and omit cwd from rows. Otherwise render cwd on each row. Single-cwd projects (most personal repos) get a clean header; multi-cwd projects (like `gmux` with its `.grove/*` worktrees) get per-row disambiguation.
- **Project-level launch button** lives in the header, top-right. Replaces the per-cwd launch buttons of the old layout.
- **CSS cleanup.** Removed the no-longer-used `.hub-host*`, `.hub-folder*`, `.session-card*` styles and the `.mock-mode .session-card-close` rule. Added `.hub-header*` / `.hub-title*` / `.hub-cwd` for the new layout.

## Why

The old project page had a clear concept (host topology → cwd folders → session pills) but used the space poorly: most projects had 1-2 sessions per cwd, leaving rows with single pills floating in whitespace. Switching to the activity-first three-section pattern matches the home dashboard's job ("what here needs my attention") and uses the full row width.

Worktree grouping doesn't disappear; it moves into the cwd-per-row line on multi-cwd projects. If anyone misses worktree-grouped view we can add a toggle later.

## Verification

- `pnpm lint` and `pnpm test` clean.
- Visual check on dev daemon:
  - Multi-cwd project (`/gmux`, 12 sessions, several `.grove/*` worktrees): cwd shown per row, no subtitle. Sections render correctly with sleeping icons on Recent rows.
  - Single-cwd project (`/pi-mono`): cwd shown as subtitle, rows clean.
  - Single-cwd project with Recent (`/dots`): subtitle + Running + Recent sections all render; `exited (N)` status visible.

## Out of scope (next step)

- Manage-projects modal trim. Currently the modal still shows the "Your sidebar" list, which is now duplicated by the home dashboard's projects section. The clean endpoint is "modal is additions only, home owns reorder/remove" — but that needs a design pass on how prominent drag handles and remove × should be on the home dashboard. Open question, deferred to a follow-up PR.

### DIFF
--- a/apps/gmux-web/src/home-partition.test.ts
+++ b/apps/gmux-web/src/home-partition.test.ts
@@ -5,7 +5,13 @@
 // down: section priority, sort key, and the Recent floor/window/cap.
 
 import { describe, it, expect } from 'vitest'
-import { partitionForHome, RECENT_WINDOW_MS, RECENT_FLOOR, RECENT_CAP } from './store'
+import {
+  partitionForHome,
+  partitionForProject,
+  RECENT_WINDOW_MS,
+  RECENT_FLOOR,
+  RECENT_CAP,
+} from './store'
 import { makeSession } from './test-helpers'
 
 const NOW = Date.parse('2026-06-01T12:00:00Z')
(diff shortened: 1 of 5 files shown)


mgabor3141:
Addressed Greptile P1 (idle sessions silently disappear):

`partitionForHome` was the wrong tool: its recency-window + cap discards exactly what a project page must keep visible. Added `partitionForProject(sessions)` in `store.ts` which preserves Needs-attention / Running but routes everything else into an unbounded, recency-sorted `rest` bucket. The project page now renders `rest` under an "All sessions" section. Four new tests in `home-partition.test.ts` pin the no-window / no-cap behavior.

mgabor3141:
Addressed Greptile re-review (duplicate launch button on empty-project page): the page header already carries a `<LaunchButton>` regardless of session count, so `EmptyProject` rendering another one was duplicative. Stripped `EmptyProject` down to the descriptor text and removed the unused `launchCwd`/`peer` props plus the now-orphaned `.hub-empty-path` / `.hub-empty-launch` CSS.

The dangling-comma finding (`.mock-mode .session-close-btn,`) is unrelated to the activity-first work and landed as its own PR: #235.

PR #233, opened 2026-05-24
feat(web): home owns reorder/remove; modal trimmed to additions only

mgabor3141:
## Summary

Step 4b of the home/project/sidebar polish thread. Moves reorder + remove of configured projects to the home dashboard's projects list, then trims the manage-projects modal to additions-only. Closes the loop on the "home subsumes manage-projects" design pinned in the conversation.

**Stacks on #232.** Merge order: #229 → #230 → #231 → #232 → this.

## What changes

**Commit 1: home gets reorder + remove.**

- Drag-to-reorder on the home dashboard's projects list. Always-visible muted drag handle (`⠷` braille glyph at `opacity: 0.4`, brighter on hover) so the list announces itself as reorderable; touch users get the same affordance as mouse users.
- Hover-revealed `×` on each row to remove the project (or its peer reference). Matches the existing pattern on `<SessionRow>`.
- Wires `updateProjects` and `removeProject` / `removePeerReference` directly from the home component.

**Commit 2: modal trim.**

- Removes the "Your sidebar" section, drag state, and `<ProjectRow>` from `manage-projects.tsx`. The modal now contains exactly three addition flows: peer references, discovered, manual path.
- Modal title: "Manage projects" → "Add project".
- Sidebar bottom button: "Manage projects" → "+ Add project".
- Sidebar onboarding hint copy updated to match.
- Dead CSS removed: `.mp-project-list`, `.mp-project-row*`, `.mp-drag-handle`, `.mp-rule*`, `.mp-remove-btn*`, plus their `@media (hover: none)` overrides. `.mp-project-info` / `.mp-project-name` are kept because `<PeerReferencesSection>` still uses them.

## Design notes

The drag-handle visibility question (always-visible muted vs hover-only vs edit-mode toggle) was settled on always-visible per the conversation. Reasoning: it's the only honest signal that the list is reorderable; hidden affordances confuse users and exclude touch devices entirely. The choice of `opacity: 0.4` keeps it from competing with the project name when scanning.

The `×` remove on hover follows the existing dashboard SessionRow pattern, so the two row shapes feel consistent.

The modal rename was the more contentious copy change: "Manage projects" is the older sidebar copy that screenshots and bookmarks might reference. The rename is justified because the modal now does only adding (no longer manages anything else); leaving the old label would mislead users about what's inside.

## Verification

- `pnpm lint` and `pnpm test` clean (406 passing).
- Visual check on dev daemon:
  - Home dashboard projects list shows drag handles + `+` launch + `×` remove on each row.
  - Drag-reorder works (drop target highlight visible on hover during drag).
  - Modal opens from sidebar button and home dashboard footer; shows "Add project" title with only Discovered + Add local path (plus the conditional From-other-hosts when peers have shareable projects).

## Out of scope

Nothing identified. This closes the design conversation's step 4 (project page rewrite + manage-projects consolidation).

### DIFF
--- a/apps/gmux-web/src/home.tsx
+++ b/apps/gmux-web/src/home.tsx
@@ -8,17 +8,63 @@
 // Section semantics, sort order, and the Recent floor/window/cap
 // live in store.ts:partitionForHome. This file is presentation.
 
-import { health, folders, homePartition, dismissSession } from './store'
+import { useCallback, useState } from 'preact/hooks'
+import {
+  health, folders, homePartition, dismissSession, projects,
+  updateProjects, removeProject, removePeerReference,
+} from './store'
 import { HostSuffix } from './host-suffix'
 import { SessionRow } from './session-row'
 import { LaunchButton } from './launcher'
 import { sessionPath } from './routing'
-import type { Folder, Session } from './types'
+import type { Folder, Session, ProjectItem } from './types'
+
+/** Drag-to-reorder transient state. `from` is the index lifted off,
+ *  `over` is the current insertion target. Set to null when nothing
+ *  is being dragged. Lives in the component, not the store: it's
+ *  view-only and reset on every mouse-up. */
+interface DragState { from: number; over: number }
+
+function reorder<T>(arr: readonly T[], from: number, to: number): T[] {
+  const out = [...arr]
+  const [moved] = out.splice(from, 1)
+  out.splice(to, 0, moved)
+  return out
+}
 
 export function Home({ onManageProjects }: { onManageProjects: () => void }) {
   const healthVal = health.value
   const foldersVal = folders.value
+  const projectsVal = projects.value
   const { needsAttention, running, recent } = homePartition.value
+
+  // Drag-to-reorder state for the Projects list. `configured` is the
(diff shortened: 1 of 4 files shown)


mgabor3141:
Addressed Greptile P2 (side effect inside `setDrag` updater): the `updateProjects()` call is now in the body of `handleDragEnd` before `setDrag(null)`, with `drag` added to the dep array. Preact's current behavior doesn't double-invoke updaters but the pattern is a footgun if that ever changes.

mgabor3141:
Addressed Greptile re-review (`.mp-project-list` and `.mp-project-detail` were over-deleted): both classes are still emitted by `PeerReferencesSection` and `DiscoveredRow` respectively in `manage-projects.tsx`, so I restored them in `styles.css` with comments noting their consumers. Verified by grepping for `mp-` class usages against the remaining rules; the rest of the cleanup stands.

mgabor3141:
Addressed Greptile re-review (double PUT on drop): `onDrop` now only calls `preventDefault()` to consume the event; the commit happens exclusively in `onDragEnd`, which fires for both successful drops and cancellations (Esc, release outside any target). One gesture, one PUT.

PR #234, opened 2026-05-25
feat(web): auto-reload tab when bundle goes stale relative to daemon

mgabor3141:
## Summary

Auto-reload the tab onto a fresh bundle when the daemon is upgraded while the tab stays open. Triggered on the *next in-app navigation*, not on a timer, so the user never sees the page yank out from under them.

**Stacks on #233.** Merge order: #229 → #230 → #231 → #232 → #233 → this.

## Why nav-triggered

The first iteration of this PR used `focusout` to detect when the user was idle and reloaded then. That had a real bug Greptile flagged: `focusout` fires on the *outgoing* focus owner while `document.activeElement` is already reset to `body`, so the busy-deferral guard was inverted (the watcher could decide "user is idle" mid-tab-key between two inputs and reload while they were typing).

The fix wasn't a better busy check; the fix was that we shouldn't pick the reload moment ourselves. We piggy-back on a moment the user already chose: the next time they click a sidebar link, hit back, or any other in-app navigation, we convert the soft route change into a full document load (`location.assign`). The browser's normal nav transition hides the reload entirely.

Side benefit: the watcher's machinery collapses dramatically. No focus listeners, no busy heuristic, no timer.

## How

- **`version-watch.ts`** exposes a `bundleStale` signal that flips true when `health.version !== __GMUX_VERSION__` and stays false otherwise. The effect resets the loop-guard marker whenever the versions converge.
- **`store.ts navigate()`** calls `navigateWithReload(url)` before its usual `pushState`-style soft nav. When `bundleStale` is set, it writes the loop-guard marker and `location.assign(url)`s.
- **Loop guard**: `sessionStorage['gmux:reload-from']` records the bundle version that triggered the last reload. On mount, if the marker still matches the current bundle (meaning the reload didn't actually pick up a new bundle, e.g. CDN cache or broken deploy), `decideBundleState` returns `stuck` and the watcher stays quiet until a version match clears the marker.

## Edge cases

- **User stays on one page forever.** They never reload. Fine. Daemon-tab mismatch only matters across interactions; `gmuxd` is backward-compatible across these patch versions.
- **First load after a deploy hits a stale-version tab.** First click reloads. Expected.
- **Reload loop.** Marker latches once. After one reload, `bundleStale` becomes false until versions match again. No timer-based escape; if the daemon and bundle are permanently mismatched, the watcher gives up rather than churning.
- **Programmatic redirects (`navigateToSession`).** These go through the same `navigate()`, so resume / notification clicks also pick up the new bundle.

## Verification

- `pnpm lint` clean, `pnpm test` 418 passing.
- `decideBundleState` and `navigateWithReload` covered by unit tests in `version-watch.test.ts`.

## Out of scope

A background timer that reloads after N hours of staleness when the tab is hidden. Worth considering if the "user stays on one page forever" case becomes a real support burden, but not now.

### DIFF
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -17,6 +17,7 @@ import { ProjectHub } from './project-hub'
 import { Home } from './home'
 import { LaunchButton } from './launcher'
 import { installCopySession } from './mock-data/export-session'
+import { installVersionWatch } from './version-watch'
 
 import {
   sessions, connState, selected, selectedId, view, health, peers,
@@ -40,6 +41,11 @@ if (USE_MOCK) document.documentElement.classList.add('mock-mode')
 
 // Debug: __gmuxCopySession() in devtools console
 installCopySession()
+
+// Auto-reload when the bundle goes stale relative to the daemon.
+// Mock mode is offline-only and the daemon version is fixed, so the
+// watcher is pointless there and would risk masking real bugs.
+if (!USE_MOCK) installVersionWatch()
 
 // ── Components ──
 
--- a/apps/gmux-web/src/store.ts
+++ b/apps/gmux-web/src/store.ts
@@ -17,6 +17,7 @@ import { signal, computed, batch, effect } from '@preact/signals'
 import type { Session, ProjectItem, DiscoveredProject, PeerInfo, PeerProject, LauncherDef, Folder } from './types'
 import type { View } from './routing'
 import { resolveViewFromPath, viewToPath } from './routing'
+import { navigateWithReload } from './version-watch'
 import { buildProjectFolders, discoverProjects, countUnmatchedActive } from './projects'
 
 import { fetchFrontendConfig, buildTerminalOptions, resolveKeybinds, type ResolvedKeybind } from './config'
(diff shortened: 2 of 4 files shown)


mgabor3141:
Addressed Greptile P1 (`focusout` breaks busy-deferral) by changing the approach entirely. Rather than picking the reload moment ourselves and trying to detect "the user is idle now", we piggy-back on a moment the user already chose: the next in-app navigation. `store.navigate()` calls `navigateWithReload(url)` first; when the bundle is stale, it does `location.assign(url)` instead of a soft pushState. The browser's normal nav transition hides the reload entirely.

This drops the `isBusy()` heuristic and the `focusout` listener (the source of the bug), the timer-based evaluation, and most of the `decideVersionAction` complexity. PR body updated with the full rationale.

mgabor3141:
Addressed Greptile re-review (`replace` flag dropped): `navigateWithReload(url, replace, ...)` now takes the flag and routes to `location.replace` or `location.assign` accordingly; `store.navigate(url, replace)` forwards it. New test pins that `replace=true` propagates so auto-attach navigations (e.g. `navigateToSession`) don't grow the back stack just because the bundle happens to be stale at that moment.

mgabor3141:
Addressed Greptile re-review (loop guard erased on first mount):

`decideBundleState` now distinguishes `unknown` (daemon version not yet reported) from `fresh` (versions confirmed match). The effect only calls `removeItem(marker)` on `fresh`. So on first mount, before health arrives, the marker survives, and a subsequent stale verdict correctly checks the marker for the loop-guard condition.

New test pins both halves: `decideBundleState(undefined, '0.1.0', null)` returns `unknown`, and the marker-present variant still returns `unknown` (not anything that would override the guard's meaning).

PR #235, opened 2026-05-25
fix(web): hide session close buttons in mock mode (dangling comma)

mgabor3141:
## Summary

Tiny CSS fix. The mock-mode close-button rule had a stray trailing comma that chained it to the unrelated `.session-item.session-dragging` rule below:

```css
.mock-mode .session-close-btn,
/* Session drag-to-reorder */
.session-item.session-dragging {
  opacity: 0.4;
}
```

So in mock mode (landing-page screenshots), session close buttons rendered at `opacity: 0.4` instead of being hidden. Splits the rule into two and uses `display: none` for the mock-mode case, which matches the comment's stated intent.

Predates the activity-first stack; Greptile flagged it during the #232 re-review. Standalone PR because it has nothing to do with the dashboard rewrite.

### DIFF
--- a/apps/gmux-web/src/styles.css
+++ b/apps/gmux-web/src/styles.css
@@ -577,8 +577,17 @@ body {
   color: var(--text);
 }
 
-/* Mock mode (landing page screenshots): hide per-session close buttons */
+/* Mock mode (landing page screenshots): hide per-session close
+ * buttons. Previously chained to the .session-dragging rule below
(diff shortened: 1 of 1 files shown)


mgabor3141:
Addressed Greptile re-review: extended the mock-mode rule to cover the new wide-row close button (`.session-row-close`) in addition to the sidebar's `.session-close-btn`. Both variants are now hidden in mock-mode screenshots, matching the rule's stated intent.

PR #237, opened 2026-05-26
feat(web): tidy home/project navigation; fix terminal color bleed and replay width

mgabor3141:
A set of related UI tweaks plus two terminal-rendering bug fixes.

## Navigation tidy

- **Project page → back-to-home link.** The static `Project` eyebrow on the project hub header is replaced with a `← Home` link. The eyebrow conveyed nothing the page title didn't already; reusing the slot for navigation keeps the visual rhythm and adds a real affordance. Plain anchor so middle-click / right-click / keyboard / screen-reader all work.
- **Sidebar → home owns "+ Add project".** The add-project footer button is removed from the sidebar. Home already has the affordance; two entry points just split attention. The unused `unmatchedActiveCount` signal goes with it (no consumers).
- **Sidebar → home owns "Enable notifications".** The permission prompt moves from a cramped sidebar footer to a proper home-page block, where it has room to explain what notifications gate on and what `default` / `denied` / `granted` / `unavailable` mean. Sidebar `notifPermission` / `onRequestNotifPermission` props dropped; routing happens once at the top.

## Terminal fixes

- **Reset xterm on session switch.** Switching from a TUI session (btop, htop, vim) into a plain shell would carry the previous session's SGR state — colors, cursor mode, scroll regions — into the new buffer, most visibly as a tinted prompt. The xterm `Terminal` instance is intentionally reused across switches (creation is expensive), so the WS-reconnect path now calls `term.reset()` (RIS) right after the terminal-io epoch reset. RIS clears buffer, SGR, modes, cursor, scroll regions, character sets, and alt-buffer state — covers every bleed vector, not just colors.
- **Replay renders at recorded width.** The replay view used `FitAddon.fit()`, which re-flows scrollback to the current viewport width. Recordings made at a different width came out with corrupted cursor positions and broken box-drawing. The replay view now pins `cols` to `session.terminal_cols` (falling back to 80) and only fits rows; horizontal scroll appears when the viewport is narrower than the recording, courtesy of the existing `overflow: auto` on `.terminal-shell`.

## Verification

Mock-mode walkthrough plus live install:
- Back link renders in the eyebrow slot, hover brightens, click goes to `/`.
- Sidebar no longer has the add-project footer; "+ Add project" on home is the single entry point.
- Notification prompt renders on home as a regular page block in each of the four permission states.
- After exiting btop in one session and switching to a plain shell, the prompt renders in default colors immediately — no tinted bleed.
- Replay of a session recorded at 200×50 in a 120-column window shows horizontal scroll, rows fit vertically, no re-flow.

### DIFF
--- a/apps/gmux-web/src/home.tsx
+++ b/apps/gmux-web/src/home.tsx
@@ -17,6 +17,7 @@ import { HostSuffix } from './host-suffix'
 import { SessionRow } from './session-row'
 import { LaunchButton } from './launcher'
 import { sessionPath } from './routing'
+import { IconBell, type NotifPermission } from './sidebar'
 import type { Folder, Session, ProjectItem } from './types'
 
 /** Drag-to-reorder transient state. `from` is the index lifted off,
@@ -32,7 +33,15 @@ function reorder<T>(arr: readonly T[], from: number, to: number): T[] {
   return out
 }
 
-export function Home({ onManageProjects }: { onManageProjects: () => void }) {
+export function Home({
+  onManageProjects,
+  notifPermission,
+  onRequestNotifPermission,
+}: {
+  onManageProjects: () => void
+  notifPermission: NotifPermission
+  onRequestNotifPermission: () => void
+}) {
   const healthVal = health.value
   const foldersVal = folders.value
   const projectsVal = projects.value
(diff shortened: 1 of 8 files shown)


mgabor3141:
Visual reference for the three changes in this PR.

**1. Back-to-home link** (replaces the static `Project` eyebrow)

01-back-to-home-link

**2. Sidebar without the add-project footer button** (affordance lives on home)

02-sidebar-without-add-project

**3. Enable-notifications prompt on home** (between projects list and footer)

03-notif-on-home

PR #238, opened 2026-05-26
fix(daemon): preserve PTY size across /resume and /restart

mgabor3141:
When gmuxd forks a runner for /v1/resume or /v1/restart, the runner's stdin/stdout are nil, so `localterm.TerminalSize()` fails and the PTY falls back to 80×24. Between exec and the browser's first resize WS message, any program that captures `$COLUMNS` at startup (claude, prompt frameworks, vim, less invoked non-interactively) sticks at 80 columns.

**Fix.** The daemon now passes the last-known `sess.TerminalCols/Rows` to the runner via two new CLI flags, `--initial-cols=N` / `--initial-rows=N`. The runner uses them to override `ptyCfg.Cols/Rows` before `pty.StartWithSize`, so the child sees the right dimensions from its very first `ioctl(TIOCGWINSZ)`. `sess.TerminalCols/Rows` is reliably non-zero whenever the session was ever displayed (it's persisted in `meta.json` and round-trips across daemon restart).

## Also: deprecate `GMUX_RESUME_ID` env var

Replaces it with a `--resume-id=<id>` flag on the same path. The runner still reads the env var as a fallback when the flag is empty, so a brand-new runner paired with a pre-flag-aware daemon (rolling upgrade) still works. The daemon no longer sets it.

Why switch: the runner grew a real flag parser since the ADR was written. The flag form is greppable in `ps`, validates types via the existing parser (rejects negative sizes without ad-hoc code), shows up in `--help` (tagged "daemon-internal"), and doesn't need an env-strip step in `buildChildEnv` to prevent leakage. POSIX runner semantics in the parser (stop at first positional) already protect against a child command's argv colliding with the daemon's directives.

All three flags are rejected in management modes (`--list`, `--attach`, etc.) so misuse fails loud.

## Scope: not fixing /v1/launch in this PR

The same race exists for browser-initiated launches: the launch request from the browser has no protocol slot for client dimensions, so a fresh PTY starts at 80x24 too. Fixing it requires adding `cols/rows` to the launch request and threading them through. CLI launches (`gmux <cmd>` in a terminal) are not affected — they detect their own TTY size via `localterm.TerminalSize()`. The misleading comment in the old `/v1/launch` code path is updated to call out the gap and the follow-up.

## Docs

ADR 0003 is revised: alternative E ("CLI flag instead of env var") was originally rejected as too public-CLI; reconsidered and adopted now that the runner has a real flag parser. Status flipped to Accepted with a "Revised: 2026-05-26 (env var → CLI flag)" line.

## Tests

- `TestBuildLaunchArgs` (daemon): fresh launch is unmodified, restart prepends directives in stable order, zero dims omit size flags, child commands with colliding flag names survive intact.
- `TestParseCLIDaemonDirectives` (runner): flags captured into `flags`, rest contains only the child command verbatim, child argv with same flag names not absorbed.
- `TestParseCLIErrors` (runner): new cases for directives-with-management-mode and negative sizes.

### DIFF
--- a/cli/gmux/cmd/gmux/cli.go
+++ b/cli/gmux/cmd/gmux/cli.go
@@ -39,6 +39,15 @@ type flags struct {
 	host        string // --host=<peer>: target this peer instead of local
 	all         bool   // --list --all: include peer sessions in output
 	help        bool
+
+	// daemon→runner directives. Only set when gmuxd forks a runner
+	// for /v1/launch, /v1/resume, /v1/restart. End users have no
+	// reason to pass them, but they're plain CLI flags so the
+	// daemon↔runner contract is greppable and toolable; the legacy
+	// GMUX_RESUME_ID env var is still honored as a fallback.
+	resumeID    string // --resume-id=<id>: reuse this session id
+	initialCols int    // --initial-cols=N: pre-size PTY
+	initialRows int    // --initial-rows=N: pre-size PTY
 }
 
 // parseCLI parses argv (without program name) and decides which mode to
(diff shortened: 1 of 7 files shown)


mgabor3141:
Re Greptile #1 (uint16 truncation): not changing. Per AGENTS.md guidance against speculative defense-in-depth, this is a daemon-internal flag that only the daemon calls, and the daemon always emits a `uint16`. A user manually passing `--initial-cols=99999` would get a garbage-sized session, which is their problem, not a class of bug worth guarding against in a hot path.

Re Greptile #2 (daemon-before-runner upgrade): added a clarifying paragraph to ADR 0003. gmux and gmuxd ship as a single release (same goreleaser package, same `scripts/install.sh` step that replaces both binaries atomically). The fallback env var protects the one direction that can plausibly drift in practice — runner auto-updated faster than a containerised daemon. The reverse isn't a supported config; documenting it is the right fix, not adding compat code for an off-path scenario.

PR #240, opened 2026-05-28
fix(daemon): suppress no-op session broadcasts; deterministic snapshot order

mgabor3141:
## Summary

A user reported **~4 GB of mobile data burned in two days** with only minutes of active gmux use, plus the phone running hot and the browser tab crashing. Root cause: **peer-snapshot amplification**.

A spoke re-ships its entire `snapshot.sessions` on every change (and at its coalescer cadence). The hub's `applySessionsSnapshot` mirrored *every* session in *every* snapshot via `store.UpsertRemote`, which broadcast `session-upsert` **unconditionally**:

```
spoke emits snapshot.sessions @ ~20 Hz
  └─ hub applySessionsSnapshot → UpsertRemote(sess) for each
       └─ store.broadcast(session-upsert)  UNCONDITIONALLY
            ↳ ~15 sessions × 2 peers × 20/sec ≈ 600 redundant mutations/sec
                 └─ each fans out a full ~30 KB snapshot.sessions to every
                    browser/phone SSE subscriber → ~600 KB/sec/consumer
```

A backgrounded mobile tab burned ~2 GB/hour.

## The insight

The coalescer caps emission at ≤20 snapshots/sec/consumer — but that bound only *means* something if a mutation reflects a **real state change**. The amplification was redundant mutations the coalescer faithfully turned into full-size snapshots. A coalescer bounds *how often* you ship, not *whether you should have shipped at all*. That decision belongs at the mutation source.

## Fix (two invariants, both in the store/snapshot layer)

1. **`store.upsertCommon` suppresses the `session-upsert` broadcast when the write leaves the stored session byte-identical** — compared *after* path canonicalization and unique-slug renumbering. Two design points that matter:
   - Comparing the **normalized** value (not the caller's raw input) is what makes it correct. An earlier draft compared raw wire input in the peering layer and would have silently failed to dedup any slug-renamed or non-canonical-path session — re-broadcasting it at 20 Hz forever.
   - Doing it **in the store** (not the peering layer) means it covers every caller (local `Upsert` too) and is the only place that sees the post-transform bytes. Duplicate-slug evictions still broadcast their removes.

2. **`snapshot.ComposeSessions` sorts sessions by ID.** The store is map-backed, so without an explicit sort two snapshots of *identical* state serialize to *different* bytes (Go randomizes map iteration), defeating byte-level dedup downstream and making captures unreadable.

ADR 0001 updated to document both invariants.

## Measurements (production daemon, ~30 sessions, 3 peers)

| | Before | After |
|---|---:|---:|
| local `session-upsert`/sec | **~565** | <3 |
| `snapshot.sessions` emissions / 10s | 1,333+ | 1 (init) |
| steady-state SSE bytes/sec/client | **~600 KB** | **~10 B** |
| 1-hour idle backgrounded tab (est.) | **~2 GB** | **~80 KB** |

## Testing

`go test ./services/gmuxd/...` passes (also under `-race`). New tests:

- **store**: `TestUpsertIdenticalSuppressesBroadcast`, `TestUpsertRemoteIdenticalSuppressesBroadcast`, `TestUpsertChangedFieldStillBroadcasts` — assert broadcast suppressed on identical write, still emitted on a real field change.
- **peering**: `TestApplySessionsSnapshot_DedupsIdenticalState`, `TestApplySessionsSnapshot_BroadcastsOnRealChange` — the reported bug scenario end-to-end through the peer mirroring path.
- **snapshot**: `TestComposeSessions_SortsByID`, `TestComposeSessions_DeterministicAcrossCalls`.

## Out of scope (follow-ups)

- PTY WS keepalive (mobile NAT idle timeouts cause reconnect storms on chunky sessions; runner-side WS has no ping)
- WebSocket `permessage-deflate` (PTY redraws are ~200–500 KB uncompressed, 5–10× compressible)
- Delete the protocol-1 v1-compat `session-upsert` replay at SSE connect (~15 KB duplicate per connect)
- SSE write path doesn't notice a half-dead TCP socket (killed client stayed `ESTAB` for minutes in testing)
- Konyvtar peering: flat 30 s retry forever against an unreachable host

_(Diagnostic logging used to localize this was removed from the PR at review request; the fix stands on its own.)_

### DIFF
--- a/docs/adr/0001-snapshot-push-protocol.md
+++ b/docs/adr/0001-snapshot-push-protocol.md
@@ -115,6 +115,39 @@ worst case), no starvation, no unbounded backlog. World snapshots
 are typically much rarer than sessions snapshots, so the world
 coalescer is mostly a passthrough in practice.
 
+### Mutation hygiene (load-bearing)
+
+The ≤20 snapshots/sec/consumer bound only holds if a *mutation*
+means *state actually changed*. Two invariants make that true; both
+are easy to violate accidentally and were the cause of a production
+incident (a backgrounded mobile tab burned ~4 GB in two days):
+
+1. **No-op writes must not broadcast.** The session store suppresses
+   the `session-upsert` event when a write leaves the stored session
+   byte-identical (compared *after* normalization: path
+   canonicalization and unique-slug renumbering). Every write path
+   that can broadcast — `Upsert`, `UpsertRemote`, `Update`, and
+   `SetTerminalSize` — routes through this single guard, so the
+   property is structural rather than re-derived per call site. The
+   peer-mirroring path re-applies a spoke's entire snapshot on every
+   snapshot it receives; without this guard, N mirrored sessions ×
+   the spoke's emit rate × peer count became N×20×peers redundant
+   mutations/sec on the hub bus, each fanning out a full
+   `snapshot.sessions` to every browser — ~600 KB/sec/consumer of
+   pure churn. The same guard also absorbs the lower-frequency no-op
+   churn from the file monitor re-reading unchanged metadata and the
+   runner re-emitting identical `terminal_resize` events.
+2. **Snapshot composition is deterministically ordered.** The
+   sessions array is sorted by ID before emission. The store is
+   map-backed, so without an explicit sort two snapshots of
+   *identical* state serialize to *different* bytes (Go randomizes
+   map iteration), defeating any byte-level dedup downstream and
+   making the stream impossible to reason about in captures.
+
+Rule of thumb: a coalescer bounds *how often* you ship; it does
+nothing about *whether you should have shipped at all*. That
+decision belongs at the mutation source.
+
 ### Per-subscriber latest-only buffer
 
 Each subscriber has a bounded channel **per snapshot kind** with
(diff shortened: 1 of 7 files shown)


mgabor3141:
Good catch — addressed in b205fb71.

Rather than add a third copy of the `reflect.DeepEqual(prev, sess)` guard, I made the invariant **structural**: extracted the finalize+broadcast tail that `upsertCommon` and `Update` already nearly duplicated into shared `commitLocked` / `broadcastCommit` helpers, with the dedup baked in. Now **every** write path that can emit `session-upsert` — `Upsert`, `UpsertRemote`, `Update` — routes through one guard, so a future fourth path can't silently regress.

`SetTerminalSize` doesn't go through the slug/activity machinery (it only touches cols/rows), so it gets a cheap early-return when the dimensions are unchanged — relevant because the runner re-emits `terminal_resize` even when a follower connects at the same size.

You're right that these aren't 20 Hz like the peer path, but they're real no-op fan-outs: the file monitor re-reads identical metadata on inotify events, and status handlers re-stamp the same status. All now silent.

New tests: `TestUpdateNoOpSuppressesBroadcast`, `TestUpdateRealChangeBroadcasts`, `TestSetTerminalSizeNoOpSuppressesBroadcast` (plus the existing real-change tests still pass). ADR 0001 updated to note the guard is structural across all write paths.

Issue #244, opened 2026-05-29
Peering reconnect uses a flat 30s retry forever against an unreachable host


mgabor3141:
## Problem

A configured-but-unreachable peer (observed: `konyvtar` → `tail14d1ac.ts.net`) retries on a flat ~30s interval indefinitely. This produces steady log noise and a small but constant stream of failed-connection traffic, and never backs off even when the host has been down for hours.

## Proposal
- Apply exponential backoff with a ceiling (e.g. 30s → cap at a few minutes) to peer reconnect attempts in the peering manager (`services/gmuxd/internal/peering/`).
- Optionally collapse repeated identical failure logs (log on state change, not every attempt).

## Acceptance
- A long-down peer settles to an infrequent retry cadence.
- Logs don't repeat the same failure every 30s.

Follow-up from #240.

Issue #243, opened 2026-05-29
SSE write path doesn't detect a half-dead client; killed connections stay ESTAB for minutes


mgabor3141:
## Problem

During #240 testing, a killed `curl` SSE client stayed `ESTAB` server-side for minutes — the daemon never noticed the client was gone. The `/v1/events` write path (`fmt.Fprintf`-style writes) buffers locally without erroring promptly when the peer is gone, so the connection (and its per-subscriber goroutines/coalescer) leaks until something else forces a write error.

## Impact
- Resource leak: goroutines, coalescer subscriptions, and buffers held for dead clients.
- Skews any per-connection accounting and keeps composing snapshots for nobody.

## Proposal
- Apply a write deadline per SSE flush (`http.ResponseController.SetWriteDeadline`) and treat a deadline error as disconnect, OR
- Rely on the existing 30s heartbeat write to surface the error and tear down the subscriber on failure.
- Ensure teardown cancels the coalescer subscription and exits the per-kind goroutines.

## Acceptance
- A hard-killed SSE client is reaped within ~1 heartbeat interval.
- No leaked goroutines/subscriptions after the client disconnects (verify with a goroutine count before/after).

Follow-up from #240.

Issue #224, opened 2026-05-17
Drop the protocol-1 SSE and bulk GET surface in the next breaking release


mgabor3141:
## Context

PR #191 (rebase-merged) introduced protocol 2 — `snapshot.sessions` /
`snapshot.world` SSE events plus the `?as=peer` hub negotiation — and
kept the protocol-1 surface alive alongside it as deprecated for
backwards compatibility during a staggered fleet upgrade. ADR 0001's
"Backwards compatibility" section documents the dispatch matrix.

The compat surface is mechanical to remove once every gmuxd in the
wild is on v2+ and every consumer (web build, CLI, scripts) is too.
It is **not free**: every snapshot fan-out also fans out per-event
frames to non-asPeer subscribers, and every hub carries a second
event-decode path it never exercises against a v2 spoke.

## Scope of the deletion

Everything below is dispatched off the **same** internal broadcast
events (`session-upsert`, `session-remove`, `projects-update`,
`peer-status`); those stay on the bus as the canonical "something
changed" signal for the snapshot pump. Only the **wire emission**
and the **consumer-side decoders** for the legacy shape go away.

### Daemon (`services/gmuxd`)

- `cmd/gmuxd/main.go`, SSE handler:
  - Initial state dump loop (`for _, sess := range sessions.List()
    { sendSSE(w, \"session-upsert\", ...) }`).
  - `case \"session-upsert\", \"session-remove\"` branch of
    `activityCh` consumer.
  - `case \"projects-update\", \"peer-status\"` branch (same).
  - `isOwnedEvent` helper at file end (only caller is the SSE
    handler).
- `cmd/gmuxd/main_test.go`: `TestIsOwnedEvent`.
- `cmd/gmuxd/main.go`: `GET /v1/projects` handler. Confirm no
  remaining callers via `rg '/v1/projects'` across `apps/`,
  `services/`, `packages/`, `scripts/`.
- `internal/peering/peer.go`:
  - `sseEvent` type (per-event wire decoder; the
    `sseSnapshotSessions` and `sseActivity` types stay).
  - `case \"session-upsert\"` and `case \"session-remove\"` branches
    of `handleEvent`.
  - `case \"projects-update\", \"peer-status\"` branch (merge the
    \"hub ignores\" comment into the `snapshot.world` branch or just
    drop both — neither is reached in v2).
- `internal/peering/peering_test.go`: the two `TestHandleEvent_V1*`
  tests (regression-pinned the deletion target).

[...]

PR #245, opened 2026-05-29
feat(gmuxd): PTY WebSocket keepalive + compression (mobile data, #241/#242)

### LINKED ISSUE #241: PTY WebSocket has no keepalive: mobile NAT idle-drop triggers reconnect + full scrollback re-ship
## Problem

The runner-side PTY WebSocket (`cli/gmux/internal/ptyserver/ptyserver.go`, proxied through `services/gmuxd/internal/wsproxy/wsproxy.go`) sends **no ping/keepalive**. On cellular/NAT networks the idle TCP connection is silently dropped by the carrier's NAT after ~30–60s. The client doesn't notice until its next write, then reconnects (`apps/gmux-web/src/terminal.tsx:907`, exp backoff 500ms→8s) and the server **re-ships the full scrollback replay** — measured at 476 KB on open for one session, multi-MB for large scrollback.

On a flaky mobile link this becomes a reconnect storm: idle-drop → reconnect → multi-MB replay → idle-drop → … This is the likely remaining contributor to the reported mobile-data burn (the SSE half was fixed in #240) and the prime suspect for the mobile-browser OOM crashes.

## Evidence
- Greped ptyserver/wsproxy/main.go: no ping loop on the PTY WS. (SSE has a 30s heartbeat at `main.go:1763`; the PTY plane does not.)
- Open session = 476 KB replay; full redraw (resize) = 205–257 KB each.

## Proposal
- Add a ~25s server→client ping on the PTY WS in `ptyserver.go` (and/or `wsproxy.go`), with client pong handling, so dead connections are detected promptly and idle ones are kept alive under NAT.
- Consider a shorter client-side idle assumption so reconnects are deliberate rather than write-triggered.

## Acceptance
- Idle PTY WS survives >2min on a cellular link without reconnecting.
- A genuinely dead connection is detected within ~1 ping interval.

Follow-up from #240.

### LINKED ISSUE #242: Enable WebSocket compression (permessage-deflate) on the PTY data plane
## Problem

The PTY WebSocket runs with compression **disabled**. `nhooyr.io/websocket@v1.8.17` defaults to `CompressionDisabled`, and both `ptyserver.go:614` and `main.go:1537` use default `AcceptOptions`. PTY replay (476 KB on open) and full redraws (205–257 KB each) are highly compressible terminal text.

## Proposal
Set `CompressionMode: websocket.CompressionContextTakeover` (or `CompressionNoContextTakeover` to bound per-message memory) on the PTY WS accept options on both ends. Expect 5–10× reduction on replay/redraw traffic.

## Notes
- Pairs naturally with the PTY keepalive work (#241) — same subsystem, and you'll want to re-measure mobile data after both land.
- Context-takeover trades memory for ratio; pick based on per-connection memory budget on the runner.

## Acceptance
- PTY replay bytes-on-wire drop ~5×+ for a typical scrollback.
- No regression in latency for interactive keystroke→echo.

Follow-up from #240.

mgabor3141:
## Summary

Follow-up to #240, completing the mobile-data investigation. #240 fixed the **SSE** half (peer-snapshot amplification). This PR hardens the **PTY data plane** (browser ↔ gmuxd WebSocket) — the other half of the mobile-data burn and the prime suspect for the mobile-browser crashes.

Closes #241, closes #242. Two atomic commits.

## #241 — Keepalive ping on the browser-facing PTY WebSocket

The PTY WS carries no traffic while a terminal sits idle. Cellular carriers drop idle TCP flows behind their NAT after as little as 30–60s. Without a keepalive the browser only notices on its next write, reconnects, and the server re-ships the **full scrollback replay** (476 KB measured for one small session; multi-MB for large scrollback). On a flaky link this becomes a reconnect storm that burns data and can OOM the mobile browser.

- Adds a periodic ping (20s interval / 10s pong timeout) on the browser-facing hop of **both** proxy paths: `wsproxy.Handler` (local sessions) and `apiclient.pipeWS` (remote/peer sessions).
- New `internal/wskeepalive` package centralizes the interval/timeout/teardown semantics so the two paths can't drift, and makes the loop unit-testable.
- `websocket.Conn.Ping` requires a concurrent reader to receive the pong; both proxy paths always run a client read loop (asserted in a test that mirrors the real topology).
- No frontend change: browsers auto-reply to ping frames with pongs at the protocol level.

## #242 — permessage-deflate on the browser-facing PTY WebSocket

- Enables `CompressionContextTakeover` on the browser-facing accept in both proxy paths. Terminal replay/redraws are highly repetitive; context-takeover keeps the deflate window across frames for the best ratio.
- Only the browser-facing hop is compressed; the local Unix-socket backend hop and the hub→spoke hop don't benefit and would just burn CPU.
- **Caveat:** Safari does not implement permessage-deflate, so iOS/Safari clients fall back to plaintext (graceful). Compression helps Chrome/Firefox/Android/desktop; **#241 (keepalive) is the iOS-relevant half** of this workstream.

## Scope notes

- The runner-side `ptyserver.go` is intentionally untouched: its hop to gmuxd is a local Unix socket (no NAT, no wire).
- Compression + keepalive are independent (different concerns, independently revertable) → two commits.

## Testing

`go test ./services/gmuxd/...` passes (incl. `-race`). The existing `apiclient` ProxyWS tests (real WebSocket over httptest, large-snapshot + disconnect cases) exercise the combined keepalive path.

New tests:
- `wskeepalive.TestRun_PingsUntilContextCancelled`, `TestRun_FailedPingTriggersOnDead`, `TestRun_ShutdownRaceDoesNotReportDead`.

Verified end-to-end against the real `nhooyr.io/websocket` v1.8.17:
- A client offering `permessage-deflate` negotiates `Sec-WebSocket-Extensions: permessage-deflate`; a ~1 MB repetitive payload round-trips intact.
- Keepalive pings stay healthy with a concurrent reader (real proxy topology); `onDead` does not false-fire.

## Remaining follow-ups (tracked)
- #224 drop protocol-1 SSE/bulk-GET surface
- #243 SSE half-dead socket detection
- #244 peering reconnect backoff

### DIFF
--- a/services/gmuxd/internal/apiclient/client.go
+++ b/services/gmuxd/internal/apiclient/client.go
@@ -354,8 +354,12 @@ func (c *Client) DialWS(ctx context.Context, sessionID string) (*websocket.Conn,
 // sessionID is the ORIGINAL (unnamespaced) session ID on the spoke,
 // not the namespaced wire ID. Callers strip the suffix first.
 func (c *Client) ProxyWS(w http.ResponseWriter, r *http.Request, sessionID string) {
+	// CompressionContextTakeover compresses the browser-facing hop
+	// (terminal output is highly repetitive); the hub->spoke hop is
+	// left uncompressed. Safari falls back to uncompressed. See #242.
 	clientConn, err := websocket.Accept(w, r, &websocket.AcceptOptions{
 		InsecureSkipVerify: true,
+		CompressionMode:    websocket.CompressionContextTakeover,
 	})
 	if err != nil {
 		log.Printf("apiclient ProxyWS: accept: %v", err)
--- a/services/gmuxd/internal/apiclient/client.go
+++ b/services/gmuxd/internal/apiclient/client.go
@@ -31,6 +31,7 @@ import (
 	"nhooyr.io/websocket"
 
 	"github.com/gmuxapp/gmux/services/gmuxd/internal/sseclient"
+	"github.com/gmuxapp/gmux/services/gmuxd/internal/wskeepalive"
 )
 
 // Read limits for ProxyWS. These match the values wsproxy.go uses
@@ -385,7 +386,16 @@ func (c *Client) pipeWS(parent context.Context, clientConn, spokeConn *websocket
 	defer cancel()
 
 	var wg sync.WaitGroup
-	wg.Add(2)
+	wg.Add(3)
+
+	// Keepalive on the browser-facing hop, mirroring the local-session
+	// proxy. Runs alongside the client read loop below (which delivers
+	// the pongs) and cancels the pipe if the phone drops off. See #241.
+	go func() {
+		defer wg.Done()
+		wskeepalive.Run(ctx, clientConn,
+			wskeepalive.DefaultInterval, wskeepalive.DefaultTimeout, cancel)
+	}()
 
 	// Spoke → Client (terminal output + resize events).
 	go func() {
--- a/services/gmuxd/internal/wskeepalive/wskeepalive.go
+++ b/services/gmuxd/internal/wskeepalive/wskeepalive.go
@@ -1,0 +1,76 @@
+// Package wskeepalive provides a WebSocket ping loop that keeps an
+// otherwise-idle connection alive across NAT timeouts and promptly
+// detects a dead peer.
+//
+// Why this exists: the PTY data plane (browser ↔ gmuxd) carries no
+// application traffic while a terminal sits idle. Cellular carriers
+// drop idle TCP flows behind their NAT after as little as 30–60s.
+// Without a keepalive the browser only discovers the dead connection
+// on its next write, then reconnects and the server re-ships the full
+// multi-megabyte scrollback replay — a reconnect storm that burns
+// mobile data and can OOM the mobile browser. A small periodic ping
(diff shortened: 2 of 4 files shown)


mgabor3141:
Both test-quality points addressed in the latest push (rebased onto merged main).

**1. ShutdownRace didn't exercise the real concurrent race** — fixed. `fakePinger` now has a `block` mode where `Ping` blocks on `<-ctx.Done()`, simulating a ping in flight waiting for a pong. The renamed `TestRun_InFlightPingCancelledIsNotDead` starts `Run` with a *live* context, waits until a ping is actually blocked, then cancels concurrently from the test goroutine — so the cancellation genuinely interrupts the in-flight `Ping`. It asserts `onDead` does **not** fire (teardown, not a NAT drop). Uses a generous 5s per-ping timeout so the failure is unambiguously caused by the cancel, not the timeout.

**2. `time.Sleep`-based assertion was fragile** — fixed. `fakePinger` now signals each ping on a buffered channel; tests call `waitForPing` which blocks on that signal instead of sleeping. No `time.Sleep` in any assertion path now, so it's deterministic on a loaded runner.

Verified with `go test ./internal/wskeepalive/ -race -count=5` — all three tests pass every run.

PR #246, opened 2026-05-29
feat(env): source a fresh login environment when launching sessions

mgabor3141:
## Problem

Sessions launched by gmuxd inherit the daemon's environment, which is **frozen at daemon startup** (typically when `gmux` first auto-started it). The result:

- Editing `~/.zshrc`/`~/.profile` then clicking **Restart session** never picks up the change — the restarted runner reuses the same stale env.
- Even `gmuxd restart` only helps when run from a *pristine* login shell. Run from a terminal *inside* a gmux session (the common case), it re-freezes the already-stale env.

While investigating I also found two env **leaks**: the `gmux`→`gmuxd` auto-start path stripped nothing (so a daemon auto-started from inside a session inherited `GMUX_SESSION_ID`/`GMUX_SOCKET`/`GMUX_ADAPTER` and stamped them onto every future session), and the bare `GMUX=1` marker slipped through the existing `GMUX_`-prefix filter everywhere.

## Changes

Two commits:

**1. `fix(env)`: stop leaking session identity into spawned daemons**
New `packages/sessionenv.Strip` removes bare `GMUX` + all `GMUX_*` identity vars, while preserving `GMUX_SOCKET_DIR` (config — daemon and runner must agree on the socket dir) and `GMUXD_*`. Applied at all three spawn sites (`launchGmux`, `startBackground`, `startGmuxd`).

**2. `feat(env)`: source a fresh login environment when launching sessions**
`launchGmux` now sources a fresh **interactive-login** environment — `$SHELL -l -i`, run in the session's cwd — via a hidden `gmux --dump-env` probe that writes `os.Environ()` NUL-delimited to **fd 3** (keeps the payload clean of rc-file banner noise; NUL handles bash function exports with embedded newlines).

- **Merge** semantics: the probe inherits the daemon env and layers dotfiles on top, so `DISPLAY`/`SSH_AUTH_SOCK`/`XDG_RUNTIME_DIR` survive while `.zshrc` changes apply. Still run through `sessionenv.Strip`.
- **Robust:** synchronous, 5s timeout, **process-group kill** on expiry (via `cmd.Cancel`). Any failure (`$SHELL` unset → headless daemons, probe error, timeout) falls back to today's behavior — never a worse env, never blocks forever.
- **Scope:** daemon-initiated launch/resume/restart only. Terminal-initiated `gmux <cmd>` is unchanged (already fresh).

Design rationale and alternatives are captured in **ADR 0006**.

## Verification

- Unit tests: `sessionenv.Strip`; capture/fallback/timeout/NUL-parse/shell-quote.
- **e2e** (`TestFreshLoginEnvOnLaunch`): proves a dotfile edit reaches a **restarted** session with no daemon restart.
- Manually reproduced before/after.

## Notes

- Drive-by: `TestEndToEnd` invoked `gmuxd` without the `run` subcommand (pre-existing breakage) — fixed.
- Changelog is auto-generated from conventional commits, so it's left to the release workflow.

### DIFF
--- a/apps/website/src/content/docs/reference/environment.md
+++ b/apps/website/src/content/docs/reference/environment.md
@@ -84,3 +84,13 @@ These are available inside every session launched by `gmux`. Use them to detect 
 | `GMUX_RUNNER_VERSION` | Version of the gmux runner hosting the session. | `0.4.0` |
 
 See [Adapter Architecture](/develop/adapter-architecture) for how to use the child-to-runner API.
+
+## How a session's environment is sourced
+
+When the daemon starts, resumes, or **restarts** a session (the launch buttons in the UI), it sources a fresh environment from an interactive login shell — roughly `$SHELL -l -i` run in the session's working directory — and hands that to the session. This means edits to your `~/.zshrc` / `~/.bashrc` / `~/.profile` (and per-directory hooks like `direnv`) take effect on the next launch or restart, **without** needing a `gmuxd restart`. Clicking **Restart session** behaves like opening a fresh terminal.
+
+The captured environment is merged onto the daemon's own environment, so session/desktop variables that your dotfiles never set — `DISPLAY`, `SSH_AUTH_SOCK`, `XDG_RUNTIME_DIR`, and similar — are preserved. (One consequence of the merge: a variable you *remove* from your dotfiles may linger until the daemon itself is restarted.)
+
+If `$SHELL` is unset (typical for systemd- or Docker-managed daemons), this step is skipped and the session inherits the daemon's environment unchanged. The same fallback applies if the login shell fails or takes longer than 5 seconds. Sessions you start directly from a terminal (`gmux <cmd>`) always use that terminal's live environment.
+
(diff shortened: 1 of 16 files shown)


mgabor3141:
Addressed both Greptile findings (folded into the feature commit, force-pushed):

**1. `<-readCh` could block past the 5s deadline (P1).** Correct — once `cmd.Wait()` returns, Go's context watcher is gone and `cmd.Cancel` won't fire again, so a background process spawned by an rc file (`&`) holding fd 3 open would wedge `io.ReadAll` indefinitely. Replaced the unconditional `rr := <-readCh` with a `select` on `ctx.Done()`: on timeout we `pr.Close()` to unblock the reader and `kill(-pgid)` the lingering group, then return the timeout error (→ fallback). Added `TestCaptureLoginEnv_BackgroundHoldsFD3` (shell writes env, backgrounds `sleep 30 &` holding fd 3, exits 0) which returns at the 300ms timeout instead of hanging 30s.

**2. Stale doubled doc header on `launchGmux`.** Removed the duplicate lead-in lines.

PR #259, opened 2026-05-31
feat(web): replace colored host pills in settings with hostname suffix

mgabor3141:
The Settings modal used colored `PeerLabel` chips to mark host ownership, which differed from the muted " · hostname" suffix (`HostSuffix`) the sidebar, home, and project pages already use. This unifies on the suffix everywhere in Settings.

## Changes
- **Configured projects ("Your projects")**: dropped the leading colored chip on reference rows; ownership is already conveyed by the existing `HostSuffix` after the project name.
- **Discovered projects**: replaced the leading chip with a `HostSuffix` after the suggested slug (e.g. `my-project · laptop`).
- **References from other hosts**: replaced the leading chip with a `HostSuffix` after the slug.
- **Hosts tab**: removed the redundant colored chip on peer rows — the hostname is already the row's primary text. The neutral "this host" / "local" tags stay.

`PeerLabel` is no longer imported in `settings.tsx` (still used elsewhere in the app). Local single-host projects show no suffix, matching the established `HostSuffix` behavior (the suffix only flags cross-host ownership / multi-host setups).

## Verification
- `tsc --noEmit` clean; 426 web tests pass.
- Verified via the mock dev server: Projects and Hosts tabs render the muted suffix instead of chips, wide and narrow.

### DIFF
--- a/apps/gmux-web/src/settings.tsx
+++ b/apps/gmux-web/src/settings.tsx
@@ -21,7 +21,6 @@ import {
   removeProject, removePeerReference, localHostLabel,
   health, peers, sessions,
 } from './store'
-import { PeerLabel } from './peer-label'
 import { HostSuffix } from './host-suffix'
 import type { ProjectItem, DiscoveredProject, MatchRule, Folder } from './types'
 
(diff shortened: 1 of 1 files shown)


mgabor3141:
Superseded by #265, which folds in this commit (host pills → hostname suffix) and builds the source-grouped Hosts tab on top of it.

PR #262, opened 2026-05-31
feat(gmuxd): stable opaque node_id + ADR 0007 (host identity, naming, peer URLs)

mgabor3141:
Adds ADR 0007 capturing the host-identity / peer-URL design, and implements its first concrete piece — the stable opaque `node_id`.

## ADR 0007 (docs/adr/0007-host-identity-and-peer-urls.md)
The design we converged on for 2.0 (breaking):
- **One `identity()`**: live tailscale name if tailscale on, else `os.Hostname()`. No generated name, no name writeback, no `host.toml` writes — tailscale is the authority + state.
- Remove `tailscale.hostname`; seed the requested name at `gmuxd remote` onboarding (default `os.Hostname()`).
- Devcontainers gated on the `devcontainer.local_folder` label → ID fallback gone by construction.
- **Opaque `node_id`** for sameness, persisted next to `auth-token`.
- Peers as state (`peers.json`) via a "Connect to host" UI; name fetched from the peer, not assigned. `[[peers]]` removed.
- Short-name URLs (`@<name>`), uniform — no method prefix, no FQDN (FQDN is address metadata).
- Add-peer: dedup by `node_id` first; only a different host with a taken name is a collision, resolved viewer-side.
- No backwards compatibility (unknown `@<name>` → home).

## Implementation: `node_id` (ADR §4)
- New `internal/nodeid` package mirroring `authtoken`: `LoadOrCreate(stateDir)` generates `node_` + 32 random bytes (hex, `crypto/rand`) once and persists to `node-id` (`0600`). Present-but-corrupted file is a hard error (silent regen would split identity from peers).
- Wired into `/v1/health`, replacing the dormant hardcoded `"node-local"` placeholder (no consumer referenced it — verified across daemon, protocol, and frontend).
- Opaque by design: never shown in the UI, never part of a URL.

### Why this is the right shape
`node_id` is an equality-only sameness token, so a long `crypto/rand` value beats a UUID (no stdlib UUID; UUIDv4 is only 122 bits; structure unused). 32 bytes matches `auth-token` so both share one generate-and-persist path. No new persistence *requirement*: a peer's `auth-token` and tailscale node key already must survive on disk for it to be a stable peer; `node-id` rides along.

## Verification
- `go build ./...`, `go vet`, full `go test ./...` pass; new `nodeid` package tests cover generate/stability/distinctness/corruption/existing-file.
- Installed locally: `node-id` file generated, and `/v1/health` returns the matching `node_…` value.

### DIFF
--- a/docs/adr/0007-host-identity-and-peer-urls.md
+++ b/docs/adr/0007-host-identity-and-peer-urls.md
@@ -1,0 +1,262 @@
+# ADR 0007: Host identity, naming, and peer URLs
+
+**Status:** Proposed
+**Date:** 2026-05-29
+**Related:** ADR 0002 (project ownership from session origin), ADR 0005 (references and the local-peer exception)
+**Target:** gmux 2.0 (breaking; no backwards compatibility)
+
+## Context
+
+A host has no single identity in gmux today — it has **four**, and
+they disagree:
+
+| Source | Where it comes from | Used for |
+|---|---|---|
+| `os.Hostname()` | OS/kernel (the **container ID** under Docker) | `/v1/health` `hostname` field; peering self-echo name; devcontainer name fallback |
+| `tailscale.hostname` (config, default `gmux`) | `host.toml` | tsnet machine name → FQDN; tailnet discovery *filter* (prefix `gmux-*`) |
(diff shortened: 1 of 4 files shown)


mgabor3141:

PR #263, opened 2026-05-31
feat(gmuxd): resolve node identity from live tailscale name (ADR 0007 §1-2)

mgabor3141:
Implements ADR 0007 §1–§2: a single node identity = the **live tailscale name** when connected, else `os.Hostname()`.

## Change
- New `internal/identity` package: `Resolve(tailscaleFQDN, osHostname)` returns the first FQDN label (e.g. `gmux-laptop` from `gmux-laptop.tailnet.ts.net`) when present, else the OS hostname. Empty/degenerate first label falls back.
- `/v1/health` `hostname` field now uses `identity.Resolve(tsListener.FQDN(), os.Hostname())`, evaluated per request so it converges once the tailscale listener becomes ready.

## Why
This is the real fix for the original `@ca75413aec31` bug: a tailscale-connected container was named by its `os.Hostname()` (the ephemeral container ID). It now reports its tailscale name, which tailscale owns and keeps sticky across restarts/recreation. Peers name this node by the health `hostname` field, so this directly stabilizes peer names and URLs.

## Verification
- Unit tests for `Resolve` (fqdn wins / no-dot / fallback / empty / leading-dot).
- `go build`, `go vet`, full `go test ./...` (26 pkgs) pass.
- **Live**: on a tailscale-enabled host, `/v1/health` now reports `hostname: gmux` (the tailscale name) where it previously reported the OS hostname.

Stacked on #262 (ADR + node_id).

### DIFF
--- a/services/gmuxd/cmd/gmuxd/main.go
+++ b/services/gmuxd/cmd/gmuxd/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/gmuxapp/gmux/packages/paths"
 	"github.com/gmuxapp/gmux/packages/sessionenv"
 	"github.com/gmuxapp/gmux/services/gmuxd/internal/authtoken"
+	"github.com/gmuxapp/gmux/services/gmuxd/internal/identity"
 	"github.com/gmuxapp/gmux/services/gmuxd/internal/nodeid"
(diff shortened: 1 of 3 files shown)


mgabor3141:

PR #264, opened 2026-05-31
feat(gmuxd): gate devcontainer discovery on the local_folder label (ADR 0007 §3)

mgabor3141:
Implements ADR 0007 §3: devcontainer autodiscovery now requires the `devcontainer.local_folder` label in addition to the `GMUXD_LISTEN` env marker.

## Change
- `isGmuxContainer` returns false for any container lacking the `devcontainer.local_folder` label.

## Why
`deriveName` falls back to the container ID when a container has neither the folder label nor a usable name — that ID churns on every container recreation, producing unstable peer names/URLs. Gating discovery on the label means a discovered container *always* has a stable, folder-derived name; a label-less container is simply not discovered, so the ID fallback can never fire. The churn becomes impossible by construction rather than patched.

Behavior change: hand-rolled `docker run` gmuxd containers without the label stop being auto-discovered. This is intended for 2.0 — properly-launched devcontainers (VS Code Dev Containers / `devcontainer` CLI) always set the label.

## Verification
- `TestIsGmuxContainer` updated to require both env + label (covers env-but-no-label, label-but-no-env, etc.).
- Full `go test ./...` (26 pkgs) pass.

Stacked on the identity-resolve PR.

### DIFF
--- a/services/gmuxd/internal/devcontainers/watcher.go
+++ b/services/gmuxd/internal/devcontainers/watcher.go
@@ -348,10 +348,20 @@ func dockerSocketPath() string {
 	return "/var/run/docker.sock"
 }
 
-// isGmuxContainer returns true if a container is running gmuxd via the
-// devcontainer feature. Detection: GMUXD_LISTEN env var is set by the
-// feature's containerEnv.
+// isGmuxContainer returns true if a container is a discoverable gmux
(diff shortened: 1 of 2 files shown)


mgabor3141:

PR #267, opened 2026-06-02
fix(web): black rectangles in WebGL terminal — atlas mipmap driver bug + context-loss recovery

### LINKED ISSUE #266: WebGL context loss leaves black rectangles in the terminal (no onContextLoss handling)
## Symptoms

Reported by a user:

Image

1. **Black rectangles** rendered in place of *certain* characters in the terminal.
2. Console error:
   ```
   [.WebGL-0xcf400188a00] GL_INVALID_OPERATION: Error: 0x00000502,
   in ../../chromium/third_party/angle/src/libANGLE/renderer/gl/TextureGL.cpp,
   allocateMipmapLevelsForGeneration:1609.
   Internal error: 0x00000502: Unexpected driver error.
   ```

## Root cause

The two symptoms are one bug. xterm.js's WebGL renderer rasterizes each glyph once into a GPU **texture atlas**, then blits cached glyphs every frame. The `0x0502` error fires at the GPU/ANGLE layer during texture-level allocation — the path xterm uses to grow/upload its atlas — and typically precedes or accompanies a **WebGL context loss** (driver reset, GPU suspend, OOM, flaky driver).

Once the context/atlas is wedged:

- Glyphs **already cached** in the atlas keep rendering fine.
- Any **uncached** glyph (a character not seen before this point) can't be uploaded, so the renderer draws the empty atlas cell → a **black rectangle**.

That "only *certain* characters" detail is the tell that it's an atlas/context failure, not a font problem.

## The defect on our side

`apps/gmux-web/src/terminal.tsx` loads the WebGL addon but never handles runtime context loss:

```ts
function loadPreferredRenderer(term: Terminal) {
  try {
    term.loadAddon(new WebglAddon())
  } catch {
    /* falls back to DOM renderer */
  }
}
```

- The `try/catch` only guards the **synchronous** `activate()` (context unavailable at load time). It does nothing for a context that dies **later, at runtime** — exactly this scenario.
- The addon exposes an `onContextLoss` event specifically so embedders can react; we never subscribe. The upstream `@xterm/addon-webgl` docs are explicit that browsers *will* drop WebGL contexts (OOM/suspend/driver) and that this event must be handled.
- Result: after the driver error the addon keeps rendering into a broken context, the atlas can't accept new glyphs, and black boxes persist until a manual page refresh.

Same gap exists in `apps/gmux-web/src/replay-view.tsx` (duplicate `loadPreferredRenderer`).

## Proposed fix

Subscribe to `onContextLoss` and **recreate** a fresh `WebglAddon` (or fall back to the DOM renderer deliberately) — do **not** just dispose.

> ⚠️ Dispose-only is a known footgun: it can leave a permanently blank terminal (see NousResearch/hermes-agent#27740). The "xterm auto-falls-back to DOM on dispose" assumption is not reliable.

Sketch:

```ts
function loadPreferredRenderer(term: Terminal) {
  try {
    const addon = new WebglAddon()
    addon.onContextLoss(() => {
      addon.dispose()
      // recreate to regain GPU accel, or explicitly attach the DOM renderer
    })
    term.loadAddon(addon)
  } catch {
    /* falls back to DOM renderer */
  }
}
```

Apply in **both**:
- `apps/gmux-web/src/terminal.tsx`
- `apps/gmux-web/src/replay-view.tsx`

Suggest extracting a shared helper so the two loaders don't drift.

## Notes / rejected mitigation

- The `0x0502 Unexpected driver error` originates in the user's GPU/driver/ANGLE stack (common on integrated GPUs, small-max-texture GPUs, Intel HD driver bugs, or GPU memory pressure). We can't fix the driver — but the app must **survive** context loss gracefully instead of degrading into black rectangles.
- A `{ customGlyphs: false }` tweak was considered but rejected as the primary fix: the failing call (`allocateMipmapLevelsForGeneration`) is generic texture-level allocation for the main glyph atlas, not specific to custom glyphs. It would at best marginally reduce atlas pressure, does nothing once the context is actually lost, and degrades box-drawing rendering. It treats a symptom.

mgabor3141:
Fixes black rectangles in the WebGL terminal renderer. ✅ Confirmed fixed by the affected user (AMD Radeon 680M / radeonsi via ANGLE).

## Root cause (the reported bug)

On an **AMD Radeon 680M (radeonsi via ANGLE)**, glyphs rendered as black rectangles with a repeating console error:

```
GL_INVALID_OPERATION: 0x00000502 ... allocateMipmapLevelsForGeneration ... Unexpected driver error.
```

- The context was **never lost** — this is a live-context failure.
- The error is `gl.generateMipmap()` in xterm's glyph-atlas upload, on a 512×512 atlas (far under the 16384 max) → a **driver bug**, not a size limit.
- xterm sets **no `TEXTURE_MIN_FILTER`**, so the atlas texture uses WebGL's default `NEAREST_MIPMAP_LINEAR`, which *requires* a mipmap chain — hence the `generateMipmap` call. On this driver it fails, leaving the texture mipmap-incomplete, so glyphs sample as black. (Selecting text forced a redraw that re-uploaded and sampled correctly — which is why selection appeared to "fix" it.)

**Fix** (`pnpm patch`, pinned to `@xterm/addon-webgl@0.20.0-beta.194`): set `TEXTURE_MIN_FILTER = LINEAR` and drop `generateMipmap`. The atlas is sampled at ~1:1, so mipmaps add no meaningful quality; `LINEAR` makes the texture complete with level 0 alone and sidesteps the driver bug. The patch touches the runtime ESM bundle (`lib/.mjs`, what Vite loads) and documents the change in `src/GlyphRenderer.ts`; the unused CJS build is left pristine to keep the patch small.

> This bug exists on all published/upstream `addon-webgl` versions (confirmed: upstream `GlyphRenderer.ts` still calls `generateMipmap` with no `TEXTURE_MIN_FILTER`), so bumping the addon would not fix it. Follow-up: upstream the min-filter fix to `xtermjs/xterm.js`, then drop this patch.

## Complementary: WebGL context-loss recovery

The renderer also had **no runtime context-loss handling** — a context dropped at runtime (laptop suspend/resume, GPU reset, OOM) left black boxes until a page reload. New shared `loadWebglRenderer(term)` (wired into `terminal.tsx` + `replay-view.tsx`, replacing the duplicated `loadPreferredRenderer`):

1. On `onContextLoss`, **disposes the dead addon** — which reinstalls a `DomRenderer` (verified in the pinned fork), so the terminal keeps working.
2. **Re-attaches a fresh WebglAddon** on the next frame to regain GPU acceleration (common case: suspend/resume).
3. **Windowed circuit breaker** (>3 losses within 60s) avoids thrash on a chronically broken GPU; windowed, not a lifetime count, so a days-long session survives occasional suspend/resume losses.

Guards: skips recreation if the terminal was disposed while waiting (keys off `element.isConnected`, since `dispose()` detaches the element without nulling it); silent DOM fallback if WebGL is unavailable at construct/activate.

## Tests

`webgl-renderer.test.ts` (7 tests): load; dispose + deferred recreate on loss; windowed breaker (recovers from spread-out losses, stops on rapid thrash); disposed-while-waiting guard; WebGL-unavailable fallback; synchronous activate-throws cleanup.

`tsc --noEmit` clean · vitest **433 passed** · `vite build` clean (bundle contains the patched atlas path) · biome clean.

## Commits

1. `recover from WebGL context loss …` — the recovery handler + tests.
2. `patch @xterm/addon-webgl to drop fragile glyph-atlas mipmaps` — the fix for the reported AMD/ANGLE bug.

Fixes #266


mgabor3141:
The atlas-mipmap patch in this PR mirrors the upstream fix now in flight: **xtermjs/xterm.js#5987** ("fix(webgl): avoid glyph atlas mipmaps", issue xtermjs/xterm.js#5986). The diff is functionally identical (`TEXTURE_MIN_FILTER=LINEAR` + remove `generateMipmap`; #5987 also sets `MAG_FILTER=LINEAR`, which is already the default), and it's approved by a core maintainer.

Follow-up after merge: once #5987 ships in a published `@xterm/addon-webgl`, bump the dependency and remove `patches/@xterm__addon-webgl@0.20.0-beta.194.patch` (and the `pnpm-workspace.yaml` entry).

Issue #270, opened 2026-06-03
Surface and recover unresolved peer references (rename-proof references)


mgabor3141:
## Problem Statement

When a host that I've referenced in my sidebar changes its name — which now happens routinely on upgrade, because ADR 0007 derives the node name from Tailscale/`os.Hostname()` instead of config — my references to that host's projects silently stop working. The reference folders still render, but **empty**, looking exactly like a reachable host that just has no sessions right now. Meanwhile the host's actual sessions are stranded as "unmatched" and don't appear anywhere obvious. There's no signal that anything is wrong and no guidance on how to fix it. (Concrete case: a box referenced as `hs`, auto-discovered after upgrade as `unraid` → settling to `gmux-hs`; references `apps@hs` / `home@hs` / `chezmoi@hs` go dark.)

## Solution

Make the "unresolved reference" state **visible** and **recoverable**, and make references **rename-proof going forward**.

- The sidebar/projects view distinguishes an unresolved reference (its host isn't on the tailnet or manually added) from a merely-empty project, with a clear explanation that the host may have been renamed or removed.
- The Settings → Hosts tab gains a "Referenced but not found" group listing those hosts, each with a one-click **remap to a known host** and a **remove** action.
- The settings gear shows a red pip when there's at least one unresolved reference, so I know where to go.
- References are anchored on the stable opaque `node_id`, opportunistically backfilled whenever a reference resolves to a reachable host — so once fixed, a host renaming again never breaks the reference.

## User Stories

1. As a user upgrading a referenced host, I want its renamed identity to keep resolving my references, so that nothing breaks on upgrade.
2. As a user whose reference can't be resolved, I want it visually distinct from an empty-but-healthy project, so that I can tell something is wrong.
3. As a user with an unresolved reference, I want a clear explanation (host not on tailnet / not manually added; may have been renamed or removed), so that I understand the cause.
4. As a user, I want a red pip on the settings gear when a reference is unresolved, so that I know where to look for a fix.
5. As a user, I want the Hosts tab to list hosts that are referenced but not found, so that the error state is explicit alongside my tailnet and manual hosts.
6. As a user, I want to remap an unresolved reference to a host I pick from my current roster, so that I can reconnect `apps@hs` to the live host in one click.
7. As a user, I want remap to apply to all references that share the unresolved host name, so that I fix `apps`, `home`, and `chezmoi` together.
8. As a user, I want to remove an unresolved reference, so that I can clean up references to hosts that are truly gone.
[...]

PR #271, opened 2026-06-03
feat(gmuxd): expose peer node_id on the peer roster

mgabor3141:
Slice 1 of #270 (rename-proof peer references).

Surfaces each peer's stable opaque `node_id` (ADR 0007) on the peer roster so the viewer can anchor references on it instead of the display name:

- `SpokeHealth` now parses `node_id` from a spoke's `/v1/health`.
- `PeerInfo.NodeID` is populated in `PeerStatus()` from the cached health (omitted for offline/never-probed peers and pre-ADR-0007 daemons).
- Frontend `PeerInfo` type gains `node_id?`.

No behavior change yet — this is the foundation the resolver (slice 3) builds on.

**Test:** `TestPeerStatus_IncludesHealthData` extended to assert `node_id` flows from the spoke health to the roster (prior art for the pattern; mirrors the existing version/launcher assertions).

### DIFF
--- a/apps/gmux-web/src/types.ts
+++ b/apps/gmux-web/src/types.ts
@@ -172,6 +172,13 @@ export interface PeerInfo {
    * POST /v1/peers). Absent on older daemons.
    */
   source?: 'tailscale' | 'devcontainer' | 'manual'
+  /**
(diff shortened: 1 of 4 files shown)


mgabor3141:

PR #274, opened 2026-06-03
feat(web): surface unresolved peer references in the sidebar

mgabor3141:
Slice 4 of #270. Stacked on #273.

An unresolved reference (its host matches no roster peer — renamed or removed) no longer looks like an empty-but-healthy project. The folder now renders muted with an amber `!` pip and a tooltip: *"Host '…' isn't on your tailnet or manually added — it may have been renamed or removed. Open Settings → Hosts to remap or remove it."* The launch `+` is suppressed (you can't launch into a host that isn't there).

Distinct from the existing `missing` state (peer connected, slug gone → grey `?`) and from a disconnected-but-known peer.

Mock data gains a resolved reference (`api@server`) and an unresolved one (`legacy-app@old-tower`) for visual coverage; verified in the mock dev server (screenshot in thread). Full suite 434 pass.

### DIFF
--- a/apps/gmux-web/src/mock-data/index.ts
+++ b/apps/gmux-web/src/mock-data/index.ts
@@ -70,4 +70,9 @@ export const MOCK_PROJECTS: ProjectItem[] = [
       { path: '~/dev/openclaw' },
     ],
   },
+  // A resolved reference to a roster peer (server).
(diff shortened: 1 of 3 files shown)


mgabor3141:

PR #276, opened 2026-06-03
feat(web): red pip on the settings gear when references are unresolved

mgabor3141:
Slice 6 of #270 — the beacon. Stacked on #275.

When at least one referenced host is unresolved (`unresolvedHosts` non-empty), the sidebar settings gear shows a small red attention pip and its tooltip/aria-label change to *"A referenced host was not found — open Settings → Hosts."* This is what tells the user *where* to go fix the silently-broken reference; it clears on its own once they remap/remove or the host reappears.

Driven entirely by the slice-3 `unresolvedHosts` signal — no new state. Verified in the mock dev server (screenshot in thread). Full suite 436 pass.

This completes the user-facing loop for #270: surfaced in the sidebar (#274) → beacon on the gear (this) → recover in the Hosts tab (#275).

PR #277, opened 2026-06-03
refactor+test+docs: harden and document unresolved-reference recovery

mgabor3141:
Review pass over the #270 stack (stacked on #276).

**Hardening**
- `addPeerReference` now stamps the peer's stable `node_id` at **creation time** (when the peer is in the roster), so references made from the UI are rename-proof from the start — not only after a remap. (The opportunistic backfill for *pre-existing* name-only references remains a separate follow-up.)

**Tests** (meaningful, not padding — they cover the integration seam the unit tests don't)
- `buildProjectFolders` buckets a **renamed** host's sessions under a reference resolved by `node_id` (folder carries the live name).
- An **unresolved** reference is flagged and takes precedence over the `missing` state.

**Docs**
- `multi-machine.md` gains a "When a referenced host is renamed" section explaining node_id anchoring and the *Settings → Hosts → Referenced but not found* remap/remove recovery.

438 web tests pass.

### DIFF
--- a/apps/gmux-web/src/projects.test.ts
+++ b/apps/gmux-web/src/projects.test.ts
@@ -301,6 +301,37 @@ describe('buildProjectFolders', () => {
     expect(folders[0].missing).toBeUndefined()
   })
 
+  it('buckets a renamed host\'s sessions under a reference resolved by node_id', () => {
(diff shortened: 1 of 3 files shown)


mgabor3141:

PR #289, opened 2026-06-07
security(daemon): cap browser connections per session in wsproxy

mgabor3141:
Implements review ticket T25.

The wsproxy held an unbounded number of browser WebSocket connections per session (`p.sessions[id]`). An authed-but-buggy client stuck in a reconnect loop (the #242 storm) could pile up connections without limit, each carrying proxy goroutines plus a 4 MiB backend read buffer — a self-DoS / runaway-client footgun with no shedding. This adds a generous-but-bounded per-session cap (`maxClientsPerSession = 8`): connections beyond the cap are refused with a clear WebSocket close reason (`StatusTryAgainLater`, "too many connections for this session") rather than evicting healthy viewers, so a storming client can't knock an existing tab offline. The check and registration happen under the same lock, so concurrent connects can't race past the cap. The cap is per-session, so legit multi-tab / phone+desktop use of other sessions is unaffected, and a freed slot (viewer disconnect) is immediately reusable by a reconnecting client.

Scope note: the ticket also mentions an optional per-remote-addr cap on `/v1/events` SSE, but that part overlaps T14 and is left out of this focused change. The runner-side `ptyserver.s.clients` map is auth-gated behind an owner-only Unix socket; the hub-facing wsproxy is the meaningful blast-radius boundary, so the cap lives there.

## Verification
- `cd services/gmuxd && go build ./...` — passes
- `go test ./internal/wsproxy/` — new tests cover: refusal beyond the cap, slot not stored on refusal, cap is per-session, and a freed slot is reusable
- `go test ./...` — full daemon suite passes

Source finding: T25 (P2) connection caps.

### DIFF
--- a/services/gmuxd/internal/wsproxy/wsproxy.go
+++ b/services/gmuxd/internal/wsproxy/wsproxy.go
@@ -21,6 +21,17 @@ import (
 	"nhooyr.io/websocket"
 )
 
+// maxClientsPerSession caps how many browser WebSocket connections the
+// proxy will hold open for a single session at once. The limit is
+// generous enough for legit multi-tab / multi-device use (desktop +
+// phone + a couple of spare tabs) but bounded so an authed-but-buggy
+// client stuck in a reconnect loop (the #242 storm) can't pile up
+// unbounded proxy connections — each one carries goroutines plus a
+// 4 MiB backend read buffer. Beyond the cap, new connections are
+// refused with a clear close reason rather than evicting existing
+// viewers, so a storming client can't knock a healthy tab offline.
+const maxClientsPerSession = 8
+
 // SessionStore provides terminal-size read/write for the proxy.
 type SessionStore interface {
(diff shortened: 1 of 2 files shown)


mgabor3141:
@greptile review

mgabor3141:
Closing as insurance-without-incident: this caps a pile-up shape (new connections while keeping old ones alive) that we have not actually observed — the reconnect storms we did hit (#242, and the stale-WS bug fixed in #347) close/abandon their previous sockets, so slots free up and connections hover at 1–2.

For the record, an independent review (June 11) verdicted MERGE on implementation quality: all handler exit paths leak-free, check-and-register race-free under one lock, cap placed before the backend dial, end-to-end tests with -race. So if we ever see actual connection accumulation, this branch (`review/T25-connection-caps`) is a known-good, ready-to-rebase fix — revisit then rather than carrying the magic number now.

PR #290, opened 2026-06-07
fix(daemon): disable websocket compression on remote-session proxy

mgabor3141:
Implements review ticket T18.

The two browser-facing WebSocket proxies made opposite permessage-deflate decisions: the local-session proxy (`wsproxy.go`) disabled compression with a thorough #242 comment explaining it triggers a mobile-browser reconnect storm (the browser negotiates deflate, can't decode gmuxd's frames, drops the socket after the first large frame, and the frontend re-ships the scrollback replay on every reconnect — costing far more than compression saves), while the remote/peer proxy (`apiclient.ProxyWS`) still enabled `CompressionContextTakeover`. Same browsers, same hop type, so leaving compression on the peer path was a latent reconnect-storm bug. This change disables compression on the apiclient path to match the local proxy, and aligns the comments on both sides so they reference each other.

## Verification
- Added `TestProxyWS_NoCompressionNegotiated` which performs a raw WebSocket handshake offering `permessage-deflate` and asserts the response does not negotiate it. Confirmed it fails when compression is re-enabled and passes after the fix.
- `cd services/gmuxd && go build ./... && go test -count=1 ./internal/apiclient/ ./internal/wsproxy/` — all pass.

Source finding: E1

### DIFF
--- a/services/gmuxd/internal/apiclient/client.go
+++ b/services/gmuxd/internal/apiclient/client.go
@@ -354,12 +354,19 @@ func (c *Client) DialWS(ctx context.Context, sessionID string) (*websocket.Conn,
 // sessionID is the ORIGINAL (unnamespaced) session ID on the spoke,
 // not the namespaced wire ID. Callers strip the suffix first.
 func (c *Client) ProxyWS(w http.ResponseWriter, r *http.Request, sessionID string) {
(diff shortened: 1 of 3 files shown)


mgabor3141:
duplicate of #285

PR #292, opened 2026-06-07
docs(security): explain why gmuxd skips origin and host checks

mgabor3141:
Implements review ticket T03 (S3) — **docs-only**, after concluding the proposed code change isn't worth adding.

## What changed
Documents in security.md why gmuxd intentionally skips both WebSocket Origin verification and Host-header validation: the mandatory bearer token (plus `HttpOnly` + `SameSite=Strict` cookie) is the actual DNS-rebinding defense, so a name allowlist would add no marginal security while being brittle and easy to misconfigure into a lockout. The S3 finding was essentially that this rationale only lived in a code comment about the Unix-socket case, which doesn't apply to the TCP listener — so the fix is to state it correctly and visibly.

## Why not the Host-header allowlist the ticket proposed
DNS rebinding only bites services that authorize by network position ("you reached me on localhost"). gmuxd requires the token on every TCP request and can't disable auth, so rebinding requests get 401 and the cross-origin cookie is never attached (HttpOnly/SameSite=Strict, scoped to the login origin not `evil.com`). The token already closes the hole. A Host/Origin allowlist would be pure defense-in-depth against a hypothetical future unauthenticated endpoint — and in practice it introduced a real regression (it 403'd the shipped Traefik example), which only an extra `GMUXD_TRUSTED_HOSTS` knob could paper over. That cost (new config surface, ongoing allowlist-completeness maintenance) isn't justified under this project's state-discipline / correctness principles. Discussed and agreed to drop the enforcement.

## Verification
- Docs-only diff (`jj diff --name-only` → just `security.md`); no code or test changes.

Source finding: review/03-security.md S3

### DIFF
--- a/apps/website/src/content/docs/security.md
+++ b/apps/website/src/content/docs/security.md
@@ -49,6 +49,22 @@ This is acceptable when:
 For remote access without managing certificates yourself, use [Tailscale](/remote-access/). For container deployments, see [Running in Docker](/running-in-docker/) for examples with WireGuard and Traefik.
 :::
 
+### Host-header allowlist (DNS-rebinding defense)
+
+The TCP listener validates the `Host` header on every request. A request is accepted only when its Host is one of:
+
+- a loopback name (`localhost`) or any IP literal (covers `127.0.0.1`, `::1`, and tailnet IPs)
+- the configured bind host (`GMUXD_LISTEN`, when it is a hostname)
+- the tailnet FQDN, once the optional Tailscale listener is up
+
+Everything else is rejected with `403 Forbidden`. This is defense-in-depth against [DNS rebinding](https://en.wikipedia.org/wiki/DNS_rebinding): a malicious web page whose domain resolves to your loopback address can make your browser issue requests to the daemon, but those requests carry the attacker's domain in the Host header, so the allowlist drops them. The bearer token / `SameSite=Strict` cookie already block this attack; the Host check is a cheap second layer at the listener level. The Unix-socket and tsnet listeners are exempt — they are reached over channels that DNS rebinding cannot target.
+
+If you front gmuxd with a reverse proxy that forwards a public hostname (e.g. Traefik routing `gmux.example.com` to the daemon), add that hostname to the `GMUXD_TRUSTED_HOSTS` environment variable (comma-separated) so it passes the check. It is empty by default.
+
+### Why WebSocket Origin checks are skipped
+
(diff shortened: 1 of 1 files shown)


mgabor3141:
@greptile review

mgabor3141:
@greptile review

mgabor3141:
One gap worth noting in the rationale here. The argument is sound for **cross-site** attackers and DNS rebinding, and dropping the brittle Host-name allowlist (`GMUXD_TRUSTED_HOSTS` enumeration) is the right call. But it conflates "cross-site" with "cross-origin" and misses the **same-site co-tenant** case:

`ts.net` is on the Public Suffix List (`ts.net` + `*.c.ts.net`, submitted by Tailscale), so `<tailnet>.ts.net` is the registrable domain = the site. That means **every other web service on your own tailnet is same-site with gmux** — and `SameSite=Strict` provides *no* protection between same-site origins. So the `gmux-token` cookie *is* attached on requests from `anything-else.<tailnet>.ts.net` to gmux. A compromised/XSS'd co-tenant tailnet service can therefore CSRF or CSWSH the `/v1/` + WebSocket control plane, which is effectively RCE (`/new`). `WhoIs` doesn't help — the request comes from the user's own authorized device.

This isn't the rejected allowlist resurfacing. The fix is a single same-origin check, not a name enumeration, so the brittleness objection doesn't apply:

> **Cookie-authed requests ⇒ enforce same-origin. Bearer-token requests ⇒ no origin constraint.**

Bearer headers can't be forged cross-origin, so the bearer reasoning in this doc stays fully intact. Concretely: re-enable WS `CheckOrigin` for cookie-authed upgrades (same-origin only — this is the highest-value one since WS→PTY is the RCE path and it's exactly what's disabled today), require `Sec-Fetch-Site: same-origin` on cookie-authed state-changing `/v1/` requests, and add `frame-ancestors 'none'`. All cheap, none brittle.

Separately, the accuracy bug already flagged stands: the "Host-header allowlist" section describes `GMUXD_TRUSTED_HOSTS` + a 403 path as if implemented, but the enforcement was dropped — so the docs should describe the actual posture (token + SameSite as the rebinding defense, plus the cookie-path same-origin check above) rather than the unbuilt allowlist.

mgabor3141:
Closing in favor of implementing the same-origin hardening discussed in the comment above (cookie-authed ⇒ same-origin enforcement on WS upgrades and mutating /v1/ calls; bearer clients exempt; frame-ancestors deny; Secure cookie). The superseding PR will carry corrected docs for the actual posture — this PR's text also described the dropped GMUXD_TRUSTED_HOSTS enforcement as if implemented. The Host-allowlist rejection rationale documented here remains valid and will be preserved in the new docs.

PR #293, opened 2026-06-07
fix(daemon): re-poll session store during restart so a dropped exit event doesn't time out

mgabor3141:
Implements review ticket T19 (rebased onto current main; extended with a review follow-up).

The `restart` handler subscribes to the session event bus, kills the runner, then waits up to 5 s for the session to become resumable. Because `store.broadcast` silently drops events when a subscriber's 64-slot buffer is full, an event burst could drop the exit upsert and produce a spurious `504 kill_timeout` even though the kill succeeded. `handleWait` already defends against this with a 500 ms re-poll ticker; `restart` did not.

This change mirrors that defense fully:

- **Re-poll fallback**: the wait loop re-checks `sessions.Get(sessionID)` every 500 ms, so a dropped exit event delays restart by at most one tick instead of timing out. Extracted into a `waitForSessionExit` helper (wait.go) to make it unit-testable; the 5 s overall deadline is preserved.
- **Fast-fail on session removal** (review follow-up): if the session is removed mid-restart (UI dismiss, retention prune), the helper returns `errExitWaitRemoved` immediately — mapped to `409 session_removed` — instead of hanging out the full deadline and reporting a misleading `kill_timeout`. Safe because a `session-remove` for an id is never followed by a transient re-upsert of that id (shadow eviction only removes *other* ids).
- **Single source of truth for "resumable"**: the helper checks the store-stamped `Session.Resumable` field instead of duplicating the `!Alive && len(Command) > 0` predicate.

## Verification
- `cd services/gmuxd && go build ./... && go test ./...` — all packages pass; `waitForSessionExit` tests also pass under `-race`.
- Tests in `cmd/gmuxd/wait_test.go`:
  - `TestWaitForSessionExitRepollsDroppedEvent` — saturates the subscriber buffer so the exit upsert is dropped, asserts the helper still returns the exited session via the re-poll fallback.
  - `TestWaitForSessionExitReturnsOnEvent` — primary path: the exit upsert arrives on the channel and the helper returns without waiting for a tick.
  - `TestWaitForSessionExitFailsFastOnRemove` — session removed mid-wait surfaces `errExitWaitRemoved` promptly, not a deadline timeout.
  - `TestWaitForSessionExitTimesOut` — the overall deadline still fires when the session never exits.
- Reconfirmed the gap exists on current main before rebasing: a verbatim copy of the pre-PR inline loop times out when the exit broadcast is dropped, while the store already holds the exit state.

Source finding: E2

### DIFF
--- a/services/gmuxd/cmd/gmuxd/main.go
+++ b/services/gmuxd/cmd/gmuxd/main.go
@@ -1576,9 +1576,13 @@ func serve(stderr io.Writer) int {
 					writeError(w, http.StatusInternalServerError, "kill_failed", err.Error())
 					return
 				}
-				exited, ok := waitForSessionExit(sessions, evCh, sessionID, 5*time.Second, 500*time.Millisecond)
-				if !ok {
-					writeError(w, http.StatusGatewayTimeout, "kill_timeout", "session did not exit in time")
+				exited, err := waitForSessionExit(sessions, evCh, sessionID, 5*time.Second, 500*time.Millisecond)
+				if errors.Is(err, errExitWaitRemoved) {
+					writeError(w, http.StatusConflict, "session_removed", err.Error())
+					return
+				}
+				if err != nil {
+					writeError(w, http.StatusGatewayTimeout, "kill_timeout", err.Error())
 					return
 				}
 				sess = exited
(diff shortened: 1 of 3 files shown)


mgabor3141:
@greptile review

PR #295, opened 2026-06-07
security: move per-session sockets off shared /tmp to a per-user dir

mgabor3141:
Implements review ticket T05.

Per-session Unix sockets defaulted to the world-shared, predictable `/tmp/gmux-sessions`. On a multi-user host the first user to create the directory owns it 0700, locking every other user's runner out (cross-user DoS / squatting), even though every other gmux artifact already lives under a per-user state dir. This moves the default to a per-user location: `$XDG_RUNTIME_DIR/gmux/sessions` when set, otherwise a per-uid subdir of the system temp dir (e.g. `/tmp/gmux-sessions-1000`). The resolution lives in the shared `packages/paths` (`SessionSocketDir()`), so the runner (`cli/gmux`) and the daemon's discovery scan (`services/gmuxd`) derive the identical directory from one source of truth. The `GMUX_SOCKET_DIR` override is preserved unchanged, so tests and devcontainer discovery are unaffected. The runner already creates the dir 0700 via `ptyserver.BindSocket`.

## Why runner and daemon still agree
The daemon is auto-started by the runner and inherits its environment; `sessionenv.Strip` only removes `GMUX_*` identity vars, so both `GMUX_SOCKET_DIR` (preserved explicitly) and `XDG_RUNTIME_DIR` (non-GMUX) pass through. Same env + same uid ⇒ identical `SessionSocketDir()` in all three branches (override / XDG / per-uid temp).

## Verification
- `packages/paths`: `go build ./... && go test ./...` — pass. New `TestSessionSocketDir` covers all three branches and asserts the default is never the shared `/tmp/gmux-sessions`.
- `cli/gmux`: `go build ./... && go test ./...` — pass.
- `services/gmuxd`: `go build ./... && go test ./...` — pass (full suite incl. discovery).
- Grepped both sides: runner (`run.go`) and daemon (`discovery.go`) both resolve via `paths.SessionSocketDir()`; no hardcoded `/tmp/gmux-sessions` default remains in non-test code.
- Discovery tests and `filemon_test` set `GMUX_SOCKET_DIR` explicitly, so the override path (and devcontainer discovery, which relies on it) is exercised and unaffected. Remaining `/tmp/gmux-sessions` literals in tests are arbitrary `SocketPath` fixtures, not default-resolution dependencies — no test update needed.

## Adjacent findings (out of scope, not fixed)
- The dead `cli/gmux/internal/metadata` package still hardcodes `/tmp/gmux-meta` (`MetaDir`). It has no callers and is slated for deletion in T13, so I left it untouched.
- Docs still cite the old default: `apps/website/src/content/docs/reference/environment.md` (GMUX_SOCKET_DIR default column), `file-paths.md`, `architecture.md`, `session-schema.md`. Worth a docs follow-up.

Source finding: review S7

### DIFF
--- a/cli/gmux/cmd/gmux/run.go
+++ b/cli/gmux/cmd/gmux/run.go
@@ -102,10 +102,7 @@ func runSession(args []string, attach bool, dir runDirectives) {
 	if sessionID == "" {
 		sessionID = naming.SessionID()
 	}
-	socketDir := os.Getenv("GMUX_SOCKET_DIR")
-	if socketDir == "" {
-		socketDir = "/tmp/gmux-sessions"
-	}
+	socketDir := paths.SessionSocketDir()
 	sockPath := filepath.Join(socketDir, sessionID+".sock")
 
 	// Bind the socket BEFORE any sessionID-dependent setup
--- a/packages/paths/paths.go
+++ b/packages/paths/paths.go
@@ -2,6 +2,7 @@
 package paths
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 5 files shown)


mgabor3141:
@greptile review

PR #297, opened 2026-06-07
security: validate session ids and slugs before filesystem and url use

mgabor3141:
Implements review ticket T24.

The daemon derived a session ID from the runner's `/meta` and fed it straight into `filepath.Join` for the metadata write path (`sessionmeta.Write`) without ever validating it — a malicious `socket_path` in `POST /v1/register` could steer the derived ID to carry `..` or path separators and escape the sessions dir. Separately, `ptyserver.handlePutSlug` accepted any non-empty ≤256-byte string as a session slug (newlines, `/`, `::`), bypassing the strict rules project slugs follow; those slugs flow into `/@<peer>/<slug>` URLs and the `${peer}::${slug}` folder key.

This change:
- Adds `paths.IsValidSessionID` (`^sess-[A-Za-z0-9_-]+$`) and enforces it both at the registration boundary (`discovery.Register`) and defensively inside `sessionmeta.Write`, so a crafted ID can never escape the sessions dir on the write path. The allowlist is a touch broader than pure hex to stay robust to ID-shape evolution while excluding every path-dangerous character.
- Adds `adapter.IsValidSlug` and normalizes the session slug in `handlePutSlug`: well-formed slugs pass through verbatim, anything else is slugified, and a slug that slugifies to empty is rejected with 400.

## Verification
- `go build ./...` + `go test ./...` in packages/paths, packages/adapter, cli/gmux, services/gmuxd — all pass.
- New tests: `paths.TestIsValidSessionID`, `sessionmeta.TestWriteRejectsUnsafeID` / `TestWriteAcceptsWellFormedID`, `ptyserver.TestPutSlugValidation` (covers `/`, `::`, uppercase normalization, and unslugifiable rejection).

Note (out of scope): the ticket flags potential overlap with T14 on the register path; this PR only adds the ID guard there and leaves any T14 work untouched.

Source finding: review/01-security.md P1

### DIFF
--- a/cli/gmux/internal/ptyserver/ptyserver.go
+++ b/cli/gmux/internal/ptyserver/ptyserver.go
@@ -526,6 +526,17 @@ func (s *Server) handlePutSlug(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "slug is empty", http.StatusBadRequest)
 		return
 	}
+	// Session slugs flow into /@<peer>/<slug> URLs and the
+	// ${peer}::${slug} folder key, so normalize to the same URL-safe
+	// shape as project slugs: reject "/", "::", newlines, and other
+	// separators by slugifying anything not already well-formed.
+	if !adapter.IsValidSlug(slug) {
+		slug = adapter.Slugify(slug)
+	}
+	if slug == "" {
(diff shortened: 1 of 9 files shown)


mgabor3141:
@greptile review

PR #300, opened 2026-06-11
fix(web): mobile toolbar respects armed ctrl/alt modifiers

mgabor3141:
## Problem

On mobile, arming **alt** via the soft toolbar and then tapping **send** sent a bare Enter (`\r`) instead of Alt+Enter. An audit found the same class of bug elsewhere (see below).

## Root cause

The toolbar is wired to `sendRawInput`, which deliberately bypasses the ctrl/alt arm logic in `sendInput` (so esc/arrow escape sequences aren't corrupted by an armed modifier). Toolbar taps therefore ignored armed modifiers entirely.

## Fix

New `applyArmedModifiers` in keyboard.ts — the single source of truth for armed-modifier encoding, shared by typed input (`sendInput`) and the toolbar:

- single chars: legacy ctrl codes / ESC-prefix alt; ctrl+alt now combine instead of leaving alt dangling
- CSI cursor keys: modifier-param injection (`\x1b[D` + ctrl → `\x1b[1;5D`)
- enter/tab/esc: ESC prefix for alt (`\x1b\r` = alt+enter); CSI-u when ctrl is involved, since these combos have **no legacy encoding** (consistent with the existing ctrl+uppercase CSI-u handling)

Bugs fixed:
- armed **alt + send** → now sends `\x1b\r`
- armed **alt + ←/→** → was ignored; alt stayed armed and silently hit the next typed char
- armed **ctrl + ←/→** → bytes were right (word-nav) but the arm was never consumed, so the next typed char became ctrl+char (typing `c` sent SIGINT)

Deliberately unchanged:
- ↑/↓ layer buttons stay plain — their purpose is plain-arrow access while a modifier is armed
- ctrl-armed send slot still shows paste, so ctrl+enter remains unreachable from the bar (it's only representable via CSI-u anyway)

## Testing

- 10 new unit tests for `applyArmedModifiers`
- `tsc --noEmit` clean, all 476 web tests pass

### DIFF
--- a/apps/gmux-web/src/keyboard.test.ts
+++ b/apps/gmux-web/src/keyboard.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest'
-import { ctrlSequenceFor, formatPasteText, pickBinaryDataTransferItem } from './keyboard'
+import { applyArmedModifiers, ctrlSequenceFor, formatPasteText, pickBinaryDataTransferItem } from './keyboard'
 
 // Build an array-like stand-in for DataTransferItemList. Vitest runs in
 // node by default, where the real DOM type isn't available; a plain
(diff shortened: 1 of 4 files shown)


mgabor3141:
Generalized in the follow-up commit: `applyArmedModifiers` in keyboard.ts is now the single source of truth for armed ctrl/alt encoding (chars, CSI cursor keys, enter/tab/esc via CSI-u), used by both typed input and the toolbar. Also fixes armed alt being ignored on ←/→ and armed ctrl never being consumed by word-nav arrows.

PR #301, opened 2026-06-11
fix(web): open links reliably on mobile by hit-testing taps against the buffer

mgabor3141:
## Problem

Tapping a URL in the terminal never worked on iOS and was flaky on Android.

Root cause: xterm activates links via a 3-event handshake on the screen element — `mousemove` resolves the link into `Linkifier._currentLink` (providers reply synchronously), `mousedown` snapshots it, `mouseup` activates if both match. Mobile browsers synthesize that cascade after `touchend`, but:

- **iOS** cancels the rest of the cascade if the DOM changes during mouseover/mousemove (the "content change rule"). Our tap handler focuses the hidden textarea on touchend → style mutations + keyboard opens → cascade dies. Links could never work.
- The keyboard-open viewport resize triggers a terminal resize, and the Linkifier clears `_currentLink` on resize — racing whatever events do arrive (**the Android flakiness**, only partially mitigated by the `setTimeout(0)` from #44).

## Approach

On a clean tap, `preventDefault()` on the touchend (per spec this suppresses the browser's synthesized cascade entirely) and dispatch the mousemove/mousedown/mouseup handshake ourselves — synchronously, before any focus/keyboard/resize can interfere (`terminal-link.ts`).

- **WebLinksAddon stays the single source of truth** for link detection: identical behavior to desktop, and OSC 8 hyperlinks work too via OscLinkProvider.
- Events are dispatched **non-bubbling** on the screen element: the Linkifier (listening there) receives them, but ancestor listeners never do — keeping the handshake invisible to xterm's MouseService (whose mousedown handler calls `focus()` and would open the on-screen keyboard), selection handling, and PTY mouse reporting.
- mousedown/mouseup only fire when a link actually resolved, so plain taps inject nothing; no link → existing keyboard-toggle behavior, unchanged.
- Stale links are xterm's problem, already solved: the Linkifier clears/re-resolves `currentLink` on `onRenderedViewportChange`.

Considered and rejected (documented in the module header): expressing the keyboard toggle as a catch-all link provider — providers answer "what's at this cell", not "what should a tap do".

## Testing

- Unit tests for the handshake: ordering, coordinates, no-link short-circuit, missing screen/linkifier guards
- Full suite passes, tsc clean
- Device-verified on iOS: tap URL opens it (keyboard stays closed), tap plain text toggles keyboard as before

## Follow-up

Long-press on a URL → copy/open action sheet (prototype on a side branch) slots into the same touch handler.

### DIFF
--- a/apps/gmux-web/src/terminal-link.test.ts
+++ b/apps/gmux-web/src/terminal-link.test.ts
@@ -1,0 +1,94 @@
+import { beforeAll, describe, expect, test } from 'vitest'
+import type { Terminal } from '@xterm/xterm'
+import { openLinkAtPoint } from './terminal-link'
+
+// Node has Event/EventTarget but not MouseEvent; polyfill enough for
+// the synthetic handshake the module dispatches.
+beforeAll(() => {
+  if (typeof globalThis.MouseEvent === 'undefined') {
+    class MouseEventPolyfill extends Event {
+      clientX: number
+      clientY: number
+      button: number
+      buttons: number
+      constructor(type: string, init: MouseEventInit = {}) {
+        super(type, init)
+        this.clientX = init.clientX ?? 0
+        this.clientY = init.clientY ?? 0
+        this.button = init.button ?? 0
+        this.buttons = init.buttons ?? 0
+      }
+    }
+    // @ts-expect-error assigning polyfill into the global scope
+    globalThis.MouseEvent = MouseEventPolyfill
+  }
+})
+
+interface FakeTermOptions {
+  /** Called on mousemove; return a truthy link to simulate resolution. */
+  resolveLink?: () => unknown
+  withScreen?: boolean
+  withLinkifier?: boolean
+}
+
+function makeFakeTerm(opts: FakeTermOptions = {}) {
+  const { resolveLink, withScreen = true, withLinkifier = true } = opts
+  const screen = new EventTarget()
+  const received: { type: string, clientX: number, clientY: number }[] = []
+  const linkifier: { currentLink?: unknown } = {}
+
+  for (const type of ['mousemove', 'mousedown', 'mouseup']) {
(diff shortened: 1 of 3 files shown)


mgabor3141:
Addressed both review points in 764cb01f: `window.open` now uses `noopener,noreferrer`, and `suppressNextClick` resets after 500ms in case the browser never synthesizes the click.

mgabor3141:
Reworked the double-open prevention: instead of a `suppressNextClick` flag + timeout, the touchend handler now calls `ev.preventDefault()` when it opens a link — per spec this stops the browser from synthesizing the click in the first place. No state, no timer, no extra click listener.

mgabor3141:
Reworked the approach after digging into the exact mechanisms (c8c7c6a):

**Why links never worked on iOS**: xterm's Linkifier needs a 3-event handshake on the screen element — `mousemove` (resolves the link into `_currentLink`; providers reply synchronously), `mousedown` (snapshots it), `mouseup` (activates if both match + position still hits). Mobile browsers synthesize that cascade after `touchend`, but iOS cancels the rest of the cascade if the DOM changes during mouseover/mousemove (the "content change rule"). Our tap handler focuses the hidden textarea on touchend → style mutations + keyboard opens → cascade dies. The keyboard-resize also clears `_currentLink` (Linkifier clears on resize), which explains the Android flakiness.

**New approach**: `preventDefault()` on touchend suppresses the browser's own cascade entirely, and we dispatch the mousemove/mousedown/mouseup handshake ourselves, synchronously, before any focus/keyboard/resize can interfere. The custom regex hit-test module is gone — WebLinksAddon stays the single source of truth for link detection (identical behavior to desktop), and OSC 8 hyperlinks now work too via OscLinkProvider. mousedown/mouseup are only dispatched when a link actually resolved, so plain taps feed nothing into selection handling or PTY mouse reporting.

PR #302, opened 2026-06-12
feat(web): long-press a terminal link on mobile to copy or open it

mgabor3141:
Follow-up to #301. Long-pressing a link in the terminal on a touch device opens a centered action sheet with the resolved target, **Copy URL**, and **Open**.

## Design (from the discussion)

- **Generic long-press recognizer** (`long-press.ts`): pure timer state machine wired into the existing touch handler; movement-cancel delegates to the handler's existing 6px tap-vs-pan threshold so there's a single threshold. A ≥500ms hold is a distinct intent: the release never opens a link or toggles the keyboard, even when nothing is under the finger. This is the hook the future "copy paragraph" long-press plugs into.
- **Query-only link resolution** (`terminal-link.ts` refactor): `linkAtPoint` dispatches just the hover step of the Linkifier handshake and reads `currentLink` — no activation. `openLinkAtPoint` (tap path) builds on it. WebLinksAddon/OscLinkProvider remain the single source of truth.
- **OSC 8 inspection**: `link.text` is the URI for both providers; the visible label is read from the link's buffer range. When they differ (OSC 8), the sheet shows both — the only way on mobile to see the real target before opening, since there's no hover status. xterm already rejects non-http(s) OSC 8 URIs.
- **Sheet** (`link-action-sheet.tsx`): preact, reuses the `modal-backdrop`/`modal-panel` conventions, center-center. Copy flips the button to "Copied ✓" and auto-closes (~600ms); failure shows "Copy failed" and stays. Backdrop tap dismisses (excluded from the shell's touch/click handling). Holds a snapshot taken at press time, so it stays open and truthful while output scrolls.

## Testing

- Recognizer: 7 unit tests (fake timers) — fire/cancel/end lifecycle, fired-stays-fired through movement, reset semantics
- `linkAtPoint`: label extraction (plain URL, OSC 8 label≠URI, wrapped rows, missing-row fallback), query-only dispatch assertions
- Full suite passes (492), tsc + biome clean
- Device verification pending (iOS): long-press link → sheet; copy; open; long-press plain text → no keyboard; tap unchanged

### DIFF
--- a/apps/gmux-web/index.html
+++ b/apps/gmux-web/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover" />
     <meta name="theme-color" content="#0a0e13" />
     <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
     <link rel="manifest" href="/manifest.json" />
--- a/apps/gmux-web/src/link-action-sheet.tsx
+++ b/apps/gmux-web/src/link-action-sheet.tsx
@@ -1,0 +1,99 @@
+/**
+ * Action sheet for long-pressing a link in the terminal.
+ *
+ * Centered modal (reuses the settings-modal backdrop conventions)
+ * rather than anchored to the touch point: consistent placement, no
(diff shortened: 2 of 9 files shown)


mgabor3141:
Addressed both review points (tidiness, neither affected correctness):
- `terminal.tsx`: `longPress.cancel()` in the effect cleanup so a pending hold timer can't outlive the effect.
- `link-action-sheet.tsx`: store the post-copy auto-close timer in a ref and clear it on unmount, so dismissing via the backdrop within the 600ms window doesn't leave a dangling `onClose`.

PR #304, opened 2026-06-13
fix(web): blip counts only sidebar sessions, not discovered

mgabor3141:
## What

The waiting-attention blip (gmux logo + mobile hamburger badge) counted **discovered** sessions — unstamped/unreferenced sessions that render nowhere in the sidebar. A session you've never adopted into a project would ping you for attention.

## Fix

`unreadCount` read the raw `sessions` set. It now derives from `folders`, which `buildProjectFolders` builds one-to-one from `projects.json` items — so "session is in some folder" == "session is in the sidebar" (owned project or resolved reference). Each session buckets into at most one folder, so summing alive+unread across folders needs no dedup. The selected-session exclusion and the `alive` requirement are preserved.

Both consumers (sidebar logo blip `sidebar.tsx`, mobile hamburger badge `main.tsx`) read the single `unreadCount` signal, so this fixes both.

## Tests

Added a `unreadCount` describe block in `store.test.ts`:
- discovered (unstamped) alive+unread session is excluded
- alive+unread session stamped into a folder is counted (dead/read excluded)
- currently-selected session is excluded

Full web suite (483 tests) + `tsc --noEmit` green.

### DIFF
--- a/apps/gmux-web/src/store.test.ts
+++ b/apps/gmux-web/src/store.test.ts
@@ -7,7 +7,7 @@ import {
   isSessionUnavailable, urlPath, urlSearch, filteredSessions, selectedId,
   navigateToSession, setNavigate,
   applyPending, _rawSessions, _rawWorld, _setRawWorld, _pendingMutations,
-  toUISession, localHostLabel, parseConnectURL,
+  toUISession, localHostLabel, parseConnectURL, unreadCount,
 } from './store'
 import type { PendingMutation } from './store'
 import type { Session } from './types'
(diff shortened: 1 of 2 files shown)


mgabor3141:
Good catch by Greptile on the `filteredSessions` coupling. Fixed: `unreadCount` now builds its folder set from the **unfiltered** `sessions` (via a shared `foldersFrom` helper) rather than the visible `folders` signal, so the global blip no longer undercounts when a `?project=`/`?cwd=` view filter is active. Added a regression test (`ignores the ?project=/?cwd= view filter`). The visible sidebar `folders` still uses `filteredSessions` as before.

PR #307, opened 2026-06-13
feat(web): collapse the session header while the keyboard is open on phones

mgabor3141:
On a phone, the on-screen keyboard already eats half the screen; the session header then costs another row of terminal. This collapses the header while the keyboard is open, reclaiming that row, and restores it when the keyboard closes.

## Detection — visual viewport, not focus

The keyboard is detected from the **visual viewport**, not textarea focus. Focus is a false signal here: the resize path deliberately re-focuses the hidden textarea after the keyboard closes (so iOS doesn't re-blur mid-transition), which would pin the header collapsed forever. Instead, `updateKeyboardPresence` compares `window.innerHeight` (layout viewport, stable on iOS) to `visualViewport.height` (shrinks under the keyboard); the difference is the occluded height. The browser URL bar moves both together so it never registers, letting a small threshold (`KEYBOARD_PRESENCE_PX = 60`) trip early in the keyboard's slide-up animation.

That presence drives a `keyboardOpen` signal (store), which `MainHeader` reflects as a `keyboard-collapsed` class. CSS owns the *when*: `@media (pointer: coarse) and (max-width: 640px)` — width is the reliable "this is a phone" proxy (the iOS layout viewport doesn't shrink with the keyboard, so a height gate would never match in portrait). Tablets keep the header; desktop is untouched.

## Single resize

Hard requirement: the header collapse must not add a second PTY resize on top of the keyboard's. It doesn't — the collapse/expand happens *during* the keyboard's continuous `visualViewport.resize` stream, so the existing 20ms height-only debounce coalesces both into one resize. Verified on-device with a SIGWINCH-echo session: exactly one resize per open and per close, no header flicker.

(An earlier iteration carried hysteresis + a widened "transition" debounce from a focus-based prototype; once detection moved to the viewport that machinery proved unnecessary — removing it was a net simplification *and* made the post-keyboard reflow ~180ms snappier.)

## Testing

- New `keyboardOpen` signal; detection gated on `pointer: coarse`.
- Full suite green (501), tsc + biome clean.
- Device-verified (iOS): header collapses early in the keyboard animation, returns and stays on close, no flicker, single resize each way; toolbar buttons don't disturb it.

### DIFF
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -22,7 +22,7 @@ import { installVersionWatch } from './version-watch'
 import {
   sessions, connState, selected, selectedId, view, health, peers,
   terminalOptions, keybinds, macCommandIsCtrl,
-  unreadCount,
+  unreadCount, keyboardOpen,
   urlPath, urlSearch,
   initStore, setNavigate, navigateToSession,
   dismissSession, resumeSession, restartSession,
@@ -35,6 +35,11 @@ const InputDiagnostics = lazy(() => import('./input-diagnostics'))
 // ── Config ──
 
 const USE_MOCK = import.meta.env.VITE_MOCK === '1' || location.search.includes('mock')
+
+/** Visual-viewport occlusion (px) above which the on-screen keyboard is
+ * considered open. Low enough to trip early in the slide-up animation,
+ * well above sub-pixel/URL-bar noise. */
+const KEYBOARD_PRESENCE_PX = 60
 
 // Mock mode: hide close buttons and other interactive chrome via CSS.
 if (USE_MOCK) document.documentElement.classList.add('mock-mode')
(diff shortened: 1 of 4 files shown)


mgabor3141:
Addressed both findings in 8404088:

**1. Stale JSDoc (store.ts)** — corrected to describe the visual-viewport occlusion detection instead of the abandoned focus-based mechanism.

**2. Session-switch header flash** — investigated and fixed at the root. The premise (cleanup re-running on session switch) doesn't actually hold: `TerminalView` has no `key` and the setup effect's deps are stable (`onCtrlConsumed`/`onInputReady` are `useCallback`s, `fontReady` only goes `false→true` once), so it isn't torn down when you switch sessions — which is why the flash wasn't reproducing.

But rather than rely on that, I moved keyboard detection out of `TerminalView` entirely and into App's existing `visualViewport` effect (the one already computing `--app-height`). Viewport occlusion is an app-global fact; App never unmounts, so there's no per-component cleanup, no `keyboardOpen.value = false` reset, and the flash class is impossible by construction — on session switch, navigation, or anything else. `terminal.tsx` no longer references `keyboardOpen` at all. Net simpler too (detection collapses into the effect that already runs on every vv resize).

501 tests green, tsc + biome clean.

mgabor3141:
@greptile-apps review

PR #321, opened 2026-06-21
feat(web): bundle Nerd Font symbols subset with local() fallback to full font

mgabor3141:
## What

Terminal icon glyphs (Powerline, file-type, git, devicons, codicons, octicons, …) now render in the browser **on any host**, regardless of which fonts the host or client machine has installed.

## How it works

`apps/gmux-web/src/fonts.css` declares `@font-face` for `Symbols Nerd Font Mono` with an ordered `src`:

1. `local(...)` — if the user has the **full** Symbols Nerd Font (or its mono variant) installed, the browser uses it: all ~10k glyphs, **zero download**.
2. `url(...)` — otherwise it loads the **bundled ~490 KB subset**.

The font is layered as a fallback behind `'Fira Code'` in the default terminal `fontFamily` (`settings-schema.ts`, `terminal.tsx`), so text still renders through Fira Code and only icon glyphs resolve to the Nerd Font.

A `unicode-range` descriptor pins the `@font-face` to icon code points, so the browser **never fetches the woff2 for plain text / CJK / emoji** output — it loads lazily, only when an icon glyph actually appears. There are no external network requests (the subset is bundled), so the `no-external-network` e2e contract still holds.

## Lazy loading vs. the WebGL glyph atlas

xterm's WebGL renderer rasterizes each glyph into a texture atlas on first draw and caches it; this build has no font-load hook to invalidate it. Combined with lazy loading, the *first* icon can be rasterized as tofu before the font bytes arrive and stay cached until a resize.

`nerd-font.ts` bridges the gap: it watches for the icon font finishing load and clears the atlas **once** (forcing re-rasterization), without triggering the download itself — preserving the lazy/unicode-range optimization. Scoped to the live terminal; the demo `MockTerminal` has no icon glyphs so it's left untouched.

## The subset

Reproducible via `scripts/build-nerd-font-subset.sh`, which downloads a **pinned** Symbols-Only release (`v3.4.0`) and trims it with `fonttools` — re-running reproduces the committed woff2 byte-for-byte. 2,724 glyphs / ~490 KB.

Excluded to keep size down (full-font installers via `local()` still cover these):
- Material Design Icons (plane-1, thousands of glyphs)
- Font Awesome v6 block `U+ED00-EFCE` (~720 glyphs, overlaps FA legacy)

Bundle-size comparison: full font ≈ 1.15 MB, this subset ≈ 490 KB, powerline-only ≈ 9 KB.

The Nerd Fonts project **MIT** license is shipped alongside the woff2 (`apps/gmux-web/src/assets/SymbolsNerdFont-LICENSE`).

## Tests

- `nerd-font.test.ts` — the atlas-refresh state machine: no-op when already loaded, clears only once the icon font is ready (not on unrelated font loads), clears exactly once, and detaches its listener.
- `fonts.test.ts` — cross-file invariants on the bundled asset: the woff2 referenced by `fonts.css` exists, has the `wOF2` signature, stays within a size budget (catches an accidental full-font commit), and that `unicode-range` is present.
- `settings-schema.test.ts` — updated default `fontFamily` assertions.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `vitest run` — 436 passing (one unrelated pre-existing failure: `store.test.ts` needs `@gmux/protocol` built)
- [x] `vite build` — subset emitted (~500 KB), build clean
- [x] verified `unicode-range` covers 100% of the shipped glyphs (no tofu) and that the subset reproduces from the pinned release
- [ ] manual: confirm a Powerline/Starship prompt + nvim file-tree icons render in a live session


mgabor3141:
Thanks @greptile-apps — addressed all 5 in `ba46404`:

1/3/4. **Size figures corrected** to ~490 KB in both `fonts.css` and the script header.

2. **Pinned the release** — `scripts/build-nerd-font-subset.sh` now downloads `v3.4.0` via `NERD_FONTS_VERSION` instead of `releases/latest`. Re-ran the script against the pin and it reproduces the committed woff2 byte-for-byte (same Vite content hash `0dvQlBVU`). Bumping the font is now an intentional, reviewable change.

5. **Added `unicode-range`** to the `@font-face`, restricted to the Nerd Font icon code points (kept in sync with the script's `RANGES`). The browser now skips the 490 KB download entirely on sessions that emit CJK/emoji but no icon glyphs. Good catch — that was a real regression risk for non-ASCII terminal output.

Also a correction the bot surfaced: the bundled license is **MIT** (the Nerd Fonts project license), not SIL OFL — fixed the PR description accordingly.

PR #326, opened 2026-06-23
fix(web): rework launch menu popup positioning + remove appear animations

mgabor3141:
## What

Fixes the launch ("+") menu popup UX in the sidebar folder rows.

### 1. No more fade-out when the cursor moves away
The open menu is a DOM descendant of `.launch-container.folder-launch-btn`, which is `opacity: 0` until the folder row is hovered. Moving the cursor into the menu items left the row's hover and faded the menu out. Now the container stays opaque while the menu is open (`.open` class + `:has(.launch-inline-menu)`), mirroring the existing `:has(.loading)` escape hatch. The "+ hidden until row hover" behavior is preserved when the menu is closed.

### 2 & 3. Left-anchored, viewport-clamped positioning
Previously the menu's **right** edge was anchored to the button, so it grew leftward and a long target-line path pushed the left edge off-screen. Now the menu's **left** edge sits slightly left of the button (6px) and grows rightward, clamped to the viewport on both edges so it never overflows.

### 4. Default item still under the cursor (preserved)
Vertical math unchanged: the menu is lifted by `4 + (showTarget ? 32 : 0)` so the default item (e.g. Shell) lands directly under the button — open + double-click still launches the default.

### 6. Removed appear animations
Dropped the staggered `launch-item-in` keyframes, the `opacity: 0` rule, and the inline `animationDelay`. Items just appear. Kept functional micro-transitions (hover-background, launch spinner).

## Testing
- Factored positioning into a pure `computeMenuPos(rect, viewport, showTarget, menuWidth?)` helper returning `{top, left}`, with new unit tests: left-anchor, right-edge clamp, left-margin clamp (menu wider than viewport), and the default-item-under-cursor vertical offset (with/without target line).
- `pnpm test launcher` (19 passing), `pnpm lint` (tsc), `pnpm build` all green.
- Outside-click (`mousedown`) and Escape handlers unchanged and still close the menu.

## Notes
File overlap with sibling branches (project-launch-cwd touches `resolveTarget`); these changes are in the positioning/animation/styles regions only — `resolveTarget` and target-derivation untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,30 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  js:
-    name: JS (lint, build, test)
+  checks:
+    name: Lint, build, test
     runs-on: ubuntu-latest
+    # One source of truth: run the same pnpm entrypoints developers run
+    # locally. The actual commands live once in moon.yml (moon discovers
+    # projects from the workspace globs), so CI can't drift from local.
+    # `pnpm lint` = biome (--error-on-warnings, matching the root script)
+    # + `moon run :lint`; build/test fan out the same way.
+    env:
+      GOWORK: ${{ github.workspace }}/go.work
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          # moon's VCS layer needs the default branch ref to compute a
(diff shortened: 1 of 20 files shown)


mgabor3141:
Rebased onto #328 (CI infra). This PR's base is now `ci-infra`, so the diff here is just the launch-menu UX change. Merge #328 first, then this.

PR #329, opened 2026-06-23
refactor: remove vestigial Status.label; derive status display in frontend

mgabor3141:
## What

\`Status.label\` was synthesized in exactly one place — gmuxd's \`exited (N)\` on a non-zero exit — and never written by any adapter or agent hook. Every reachable render was either dead or duplicable:

- The header pill was **label-gated**, so it never showed for live sessions (the runner only ever sends \`working\`/\`error\` booleans, never a label).
- The sidebar/row already derive \`exited (N)\` from \`exit_code\`.
- \`notify.go\`'s label fallback was effectively dead (the only label is set on a dead session; the unread transition fires on a live turn-complete with no label) → always \`"New output"\`.

So the field carried no information that isn't already available from \`working\`/\`error\` + \`exit_code\`.

## Change

- Drop \`label\` from \`Status\` across the stack: \`packages/adapter\`, gmuxd \`store\`, \`packages/protocol\` schema, web \`types.ts\`.
- On exit, gmuxd sets \`Status = nil\` (was \`{label: "exited (N)"}\`). The frontend derives all exit text from \`exit_code\` (ReplayView bottom bar + sidebar fallback, both already did).
- Header pill now derives from the booleans: shows **Working…** / **Error** while an agent is busy, nothing when idle — turning a no-op into a useful live-status indicator.
- \`notify.go\` always uses \`"New output"\` for the unread body.
- Docs updated (state-management, session-schema, using-the-ui).

## Status contract now

\`Status\` carries only \`working\`/\`error\` (live state). Display text is the frontend's concern; exit text comes from \`exit_code\`. On exit, \`Status\` is \`null\`.

## Tested

- Go: \`go test ./...\` green in \`packages/adapter\`, \`services/gmuxd\`, \`cli/gmux\`; \`go vet\` clean.
- Web: 521 tests pass; \`tsc --noEmit\` clean; \`vite build\` ok; \`biome lint --error-on-warnings\` clean.
- Protocol: schema tests pass.

### DIFF
--- a/apps/gmux-web/src/home-partition.test.ts
+++ b/apps/gmux-web/src/home-partition.test.ts
@@ -45,7 +45,7 @@ describe('partitionForHome', () => {
       // session lands in; that's covered by other tests.
       const s = makeSession({
         id: 's1', cwd: '/x', alive: true,
-        status: { label: 'failed', working: false, error: true },
+        status: { working: false, error: true },
       })
       const { needsAttention } = partitionForHome([s], NOW)
       expect(needsAttention).toEqual([])
@@ -54,7 +54,7 @@ describe('partitionForHome', () => {
     it('puts alive+working sessions in running', () => {
       const s = makeSession({
         id: 's1', cwd: '/x', alive: true,
-        status: { label: 'building', working: true, error: false },
+        status: { working: true, error: false },
       })
       const { needsAttention, running } = partitionForHome([s], NOW)
       expect(needsAttention).toEqual([])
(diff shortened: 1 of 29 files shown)


mgabor3141:
Also dropped `statusText` (the `exited (N)` segment) from `SessionRow` — the wide row used on the **home dashboard** and **project hub** (not the sidebar; that's `SessionItem`). Dead sessions there now rely on the dimmed/dot treatment rather than an exit-code text segment. Removed the now-orphaned `.session-row-status` CSS. Sidebar (`SessionItem`) and ReplayView still surface exit info.

mgabor3141:
Correction per review: kept the `exited (N)` segment in `SessionRow` (home dashboard + project hub) and ReplayView's bottom bar. Instead removed status/exit text from the **sidebar** `SessionItem` only — its line-2 meta now carries just the "⚠ open elsewhere" dup-warning when applicable.

mgabor3141:
Fixed the export-session `error`-field gap Greptile flagged while in here: the mock exporter now emits `error: true` when set, so captured fixtures round-trip the error flag faithfully.

PR #332, opened 2026-06-23
ci: prototype — let moon manage the JS toolchain + auto-install deps

mgabor3141:
**Draft / exploratory** — answering "could moon manage the JS toolchain?" Yes. Pushed so CI shows the real behavior + timing. Not for merge as-is.

## What this does
- `.moon/toolchains.yml` — moon v2 split the old monolithic `node` platform into three independent toolchains, so all three are enabled: `javascript` (selects pnpm), `node`, `pnpm`. (v2 also renamed the file `toolchain.yml` → `toolchains.yml`.)
- `.prototools` — pins `node`/`pnpm`/`moon` versions as the single source of truth.
- `ci.yml` — drops `setup-node` + `pnpm/action-setup` + `pnpm install`, replaced by `moonrepo/setup-toolchain` (installs proto+moon, caches `~/.proto`) and a direct `moon run :lint :build :test`.

## Verified locally
- `moon setup` provisions proto + Node 22 + pnpm 10.32.1.
- Removing `node_modules` and running a moon task **auto-reinstalls** — the action graph gained `SetupToolchain(node/pnpm)` + `InstallDependencies(javascript)`, which were absent before.
- Full `build` + `test` pass under moon-managed Node 22.

## Tradeoffs surfaced (why this is a draft, not a merge)
1. **Lockfile churn.** The `javascript` toolchain's default `dedupeOnLockfileChange` rewrites `pnpm-lock.yaml` on install. Disabled it here; worth knowing it's on by default.
2. **Bootstrap chicken-and-egg.** moon was a pnpm devDep; if you `rm -rf node_modules` the `moon` binary goes with it. The toolchain approach wants moon installed standalone (via `setup-toolchain` / proto), which is what the CI change does — but it's a real shift in how you bootstrap.
3. **biome sits outside moon.** `pnpm lint` runs biome *before* `moon run`, so it needs deps pre-installed; with moon-managed install I had to call moon directly and run biome as its own CI step. This breaks the tidy `pnpm lint` == CI property from #328. Cleanest fix if adopted: **make biome a moon task** so the whole pipeline is one `moon run`.
4. **Go is unaffected** — the Go projects shell out to `go` directly (`language: bash`), so they still need `setup-go`. moon only manages the JS side here.
5. **Supersedes the docs note** in #330 ("moon does not auto-install") if adopted.

## Recommendation
Genuinely nice for determinism (pinned Node/pnpm everywhere) and removing the manual `pnpm install`. But it's a real infra decision with the tradeoffs above — wants its own ADR + a decision on biome-as-moon-task before it's more than a prototype.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,8 @@ jobs:
           # head and makes `moon run` fail with git exit 128.
           fetch-depth: 0
 
+      # Go isn't moon-managed (the Go projects shell out to `go` directly),
+      # so it still needs its own setup + cache.
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
         with:
           go-version: "1.26"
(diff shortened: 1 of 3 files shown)


mgabor3141:
## CI finding: proto shims collide with the system Go toolchain

The CI run fails at `moon run :lint :build :test`:
```
gmuxd:lint | Error: proto-shim: Failed to execute proto for the shimmed command: Permission denied (os error 13)
```
Once `setup-toolchain` puts proto on PATH, proto shims **all** commands including `go` — but we only pinned node/pnpm in proto, so the `go` shim is broken. The JS half works; the Go projects (which share the consolidated #328 job and shell out to `go`) break.

### What this means for adoption
Adopting moon-managed JS toolchain re-introduces a Go/JS separation problem that #328 had just consolidated away. Options, none free:
1. **Split Go back into its own job** that doesn't touch proto/moon — but then Go commands get re-listed inline (un-does part of #328's dedup).
2. **Manage Go via proto too** (add `go` to `.prototools` + a moon go toolchain) — bigger scope; the Go projects are currently `language: bash`.
3. **Disable proto shims** / scope them to node/pnpm so `go` resolves to the system binary.

### Net recommendation
The local proof is solid — moon **can** manage the JS toolchain and auto-install deps (verified: `InstallDependencies(javascript)` in the graph, wiped `node_modules` auto-reinstalls). But end-to-end it's a real project with ≥4 tradeoffs (lockfile churn, bootstrap, biome-outside-moon, and now proto/Go shim collision). **Not worth adopting casually** — it deserves its own ADR and a deliberate decision on Go + biome, weighed against the modest payoff (dropping a one-line `pnpm install` step that the build cache already makes cheap). Leaving this as a documented spike.

PR #339, opened 2026-07-04
docs: finish Status.label removal in adapter-facing docs

mgabor3141:
Follow-up to #329, prompted by an independent re-review of that PR.

#329 removed `Status.label` from the runtime but missed the **public `PUT /status` API docs**, which still taught adapter authors to send `label` — a field the runner now silently drops:

- `develop/session-schema.md`: Status type definition, both schema tables, the label-usage guidance, the curl + OSC 7777 examples, and the `/meta` + `/v1/sessions` example payloads
- `adapters.md`: curl example
- `develop/adapter-architecture.md`: curl example

All aligned with the booleans-only contract (`working`/`error`; display text derived in the frontend).

Also noted during the re-review, deliberately **not** fixed here: `tests/e2e/e2e_test.go` (Go module, not run by CI — the CI "e2e" job is Playwright) asserts `Status.Label` and a `state` field that never existed, so it was already broken on main before #329. Candidate for deletion/rewrite in a separate change.

### DIFF
--- a/apps/website/src/content/docs/adapters.md
+++ b/apps/website/src/content/docs/adapters.md
@@ -12,7 +12,7 @@ An adapter watches the terminal output of your process and reports structured st
 - **Working** — the tool is busy (cyan pulsing dot)
 - **Idle** — the tool is waiting (no dot)
 
-Without an adapter, gmux still tracks whether the process is alive — but with one, you get meaningful at-a-glance status like "thinking" or "42/50 passed".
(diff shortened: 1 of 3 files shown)


mgabor3141:
Addressed the Greptile finding: the adapters.md overview sentence still promised arbitrary status text ("thinking", "42/50 passed"). Reworded to describe what adapters actually provide now (working/error state, unread markers, session titles). The remaining "thinking / waiting for input" mentions describe what adapters *detect* (mapped to the working boolean), not displayed label text, so those stay.

PR #340, opened 2026-07-04
fix(gmuxd): stop Sweep from destroying live runners' scrollback on daemon restart

mgabor3141:
## What

Fixes a pre-existing bug surfaced by the post-merge retention audit (#333/#338
context): **every gmuxd restart destroyed the on-disk scrollback of every
live, never-died session.**

## The bug

Three facts that only line up wrong:

1. `Sweep` (startup, before discovery) `RemoveAll`s any session dir lacking
   `meta.json` as an "orphan".
2. `meta.json` is written **only when a session lands dead** (`persistDead`);
   a resumed session has a leftover one, but a fresh-launched live session
   never does.
3. The runner creates and holds open `sessions/<id>/scrollback` from launch
   (`cli/gmux/cmd/gmux/run.go`).

So a live session's dir is exactly "scrollback, no meta.json" — Sweep's orphan
definition. On restart (routine on upgrades): the open file is unlinked, the
runner keeps appending to the unlinked inode, its next rotation's `os.Rename`
hits ENOENT and sticky-fails, and when the session eventually dies its
post-mortem replay is permanently empty. Live-attach was unaffected (runner's
in-memory ring), which is why this went unnoticed.

Notably, the old `TestSweepRemovesOrphanDirs` staged **precisely the
live-runner layout** ("a scrollback file the runner wrote") and asserted its
destruction — the bug was codified as the contract.

## The fix

Sweep spares dirs that contain scrollback files (`hasScrollback`). Such a dir
is either:
- a **live runner's dir** — must not be touched; its `meta.json` arrives when
  the session dies; or
- a **crash leftover** — bounded by `PruneScrollback`'s cache cap (its bytes
  are already counted and evicted oldest-first, and the then-empty dir is
  swept normally on a later startup).

Truly empty/junk orphan dirs (e.g. stale `meta.json.tmp-*`) are still removed.
No new unbounded surface: leftovers live under the existing 256 MiB cap.

## Tests

- New `TestSweepSparesLiveRunnerScrollbackDir`: scrollback-only dir survives
  Sweep while an empty orphan alongside it is still removed (red before the
  fix — confirms the bug empirically).
- `TestSweepRemovesOrphanDirs` rewritten to junk-only orphans, with a comment
  pointing at the new contract.
- `go test -race ./...` green in `services/gmuxd`.

### DIFF
--- a/services/gmuxd/internal/sessionmeta/sessionmeta.go
+++ b/services/gmuxd/internal/sessionmeta/sessionmeta.go
@@ -373,8 +373,13 @@ func (s *Store) WatchRemovals(events <-chan store.Event) {
 // register.
 //
 // Cleanup performed during the sweep:
-//   - directories with no meta.json (orphan scrollback or empty
+//   - directories with no meta.json AND no scrollback (empty or junk
 //     dirs) are logged and removed
+//   - directories with scrollback but no meta.json are left alone:
+//     they belong to live, never-died runners (meta.json is written
+//     only on dead landings) whose open scrollback file must not be
+//     unlinked; crash leftovers among them are reclaimed by
+//     PruneScrollback's cap and swept once empty
 //   - directories whose meta.json fails to parse are logged and
 //     left in place (operator may want to inspect)
 //   - non-directory entries under the base dir are ignored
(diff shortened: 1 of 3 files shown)


mgabor3141:
Re Greptile's P1 (`go test -race ./...` failing in `internal/notify`): confirmed real, but **pre-existing and unrelated to this PR** — this diff doesn't touch notify, and the race reproduces identically on clean main (~50% of `-race` runs of `TestSessionRemove_CleansUpPrevState`). Root cause is `Router.Run`'s `prevState` seeding loop writing without `r.mu` while every other access locks. Fixed in #342 (two-line lock around the seed; 12 consecutive `-race` package runs green there).

PR #341, opened 2026-07-04
fix(web): mobile launch menu clipped by the sidebar drawer

mgabor3141:
## What

Fixes the launch ("+") menu popup being clipped and misaligned in the sidebar on mobile. Desktop (fixed in #326) is unchanged.

## Why it broke on mobile only

The menu is `position: fixed` precisely so it can escape the sidebar's overflow. But on touch layouts (`@media (pointer: coarse)`) the sidebar is a slide-in drawer with `transform: translateX(...)` — and a transformed element becomes the **containing block for fixed-position descendants**. So on mobile the menu's viewport coordinates were reinterpreted as sidebar-relative and the menu was clipped at the drawer's edge.

A second, latent issue: `computeMenuPos` lifts the menu by `4 + 32px` so the default item lands under the button, but never clamped `top`. On mobile the first project row sits at the very top of the drawer (no header above the list), so the lift pushed the menu off-screen (`top: -25px`).

## Fix

- **Portal the menu to `<body>`** (same pattern as `SheetBackdrop` in `sheet.tsx`), so its fixed coords stay viewport-relative regardless of layout.
- **Clamp `top` to the viewport margin** in `computeMenuPos` (+ unit test).
- Outside-click dismissal now also checks the portaled menu's ref, so `mousedown` on a menu item no longer counts as "outside" (which would have unmounted the item before its `click` fired).
- Updated the CSS comments; dropped the now-dead `:has(.launch-inline-menu)` opacity selector (the `.open` class covers it).

## Before / after (390×844, touch emulation)

| Before | After |
|---|---|
| Menu clipped at drawer edge + pushed above viewport | Fully visible, aligned to the trigger |

(Verified with headless Chrome at 1280×800 and 390×844 w/ touch emulation; desktop position unchanged.)

## Testing

- `pnpm test` in `apps/gmux-web`: 522 passing (new test: top-clamp for first-row buttons)
- `pnpm lint` (tsc) + biome clean
- Browser-verified: menu-item mousedown keeps menu open; outside mousedown closes it; Escape unchanged.

### DIFF
--- a/apps/gmux-web/src/launcher.test.ts
+++ b/apps/gmux-web/src/launcher.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from 'vitest'
-import { launchersForPeer, formatTarget, computeMenuPos } from './launcher'
+import { launchersForPeer, formatTarget, computeMenuPos, clampMenuLeft } from './launcher'
 import type { LauncherDef, PeerInfo } from './types'
 
 const localLaunchers: LauncherDef[] = [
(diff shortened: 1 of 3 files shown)


mgabor3141:
Follow-up pushed: the previous horizontal clamp assumed the menu's min-width (180px), but nowrap items can render it wider, so the right edge could still overflow the screen on narrow viewports.

Now the clamp is a pure `clampMenuLeft` helper that's re-run with the menu's **measured** `offsetWidth` in a layout effect (same-paint correction) — an over-wide menu gets pulled left so its right edge sits 8px inside the screen, i.e. right-aligned to the viewport. A CSS `max-width: calc(100vw - 16px)` backstop keeps the menu itself narrower than the screen.

Verified at 390×844 with a forced 300px-wide menu: renders at `left: 82, right: 382` (8px margin), no overflow. 525 tests passing (3 new for the helper).

PR #343, opened 2026-07-04
refactor(web): unify alive vs resumable terminal chrome (Option A)

mgabor3141:
## What

Unifies the chrome around alive (`TerminalView`) and dead/resumable (`ReplayView`) sessions so they feel like one screen with different bodies:

- **Lifecycle actions in both intended homes, one label source.** ReplayView's bottom action bar carries the *primary* Resume/Rerun button; the same action is deliberately mirrored in the header `SessionMenu`, where Restart muscle memory lives for alive sessions. Both surfaces read from one pure function (`session-actions.ts` → `lifecycleAction`): alive → Restart, dead+resumable agent (claude/codex/pi) → Resume, dead+resumable non-agent → Rerun, dead+non-resumable → nothing. While a resume is in flight both show the busy label and disable.
- **Exit status is header chrome only.** `HeaderStatusChip` renders "Exited (N)" in the same slot as the alive Working…/Error chip; while a resume is pending it flips to "Resuming…"/"Rerunning…" so there's feedback after the menu closes.
- **☰ reachable everywhere on touch** (fixes the regression where a dead session stranded phone users with no sidebar toggle). The hamburger with its waiting-dot badge is extracted into `MenuButton` with three touch-only placements:
  - `bar` — cell in the mobile key bar (attached sessions, as before; the bar stays attachable-only rather than showing a full set of disabled dead keys),
  - `footer` — in ReplayView's action bar (dead sessions),
  - `floating` — bottom-left overlay on home / project hub / transient states, which previously had no ☰ at all.

## Decisions made by judgment (per handoff)

- **Action placement:** bottom bar = primary action, header menu = consistent muscle-memory mirror (per maintainer direction). Status is not duplicated — it lives only in the header chip.
- **Continuous resume** (same xterm across resume, ADR 0004's SessionStream): out of scope, nav-swap kept. No conflict with that proposal.
- **Retention contract:** verified — `fetchScrollback` + ReplayView already handle empty/absent/error scrollback; no changes needed.
- Loading copy left as-is ("Waiting for output…" vs "Loading scrollback…") — genuinely different waits, same visual treatment (`terminal-loading`).

## Testing

- `session-actions.test.ts`: action gating per state/kind, menu + button labels, busy/disabled during resume.
- `pnpm test`, `pnpm lint`, `pnpm build` all green.
- Manually verified the built bundle against a live gmuxd (mobile-emulated viewport): dead session → "Exited" chip, Resume in ⋮ menu **and** footer, footer ☰ opens the sidebar; home + project hub → floating ☰ opens the sidebar; alive session → "Working…" chip, Restart in menu, full key bar, no floating button; desktop → no ☰ anywhere (sidebar always visible).

### DIFF
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -272,8 +272,10 @@ function SessionMenu({ session, onRestart, onResume, resuming }: {
   // for the secondary+F keybind).
   const canFind = session.alive
 
-  // One lifecycle action per state (Restart alive, Resume/Rerun dead);
-  // the menu is the single home for it in both states.
+  // One lifecycle action per state (Restart alive, Resume/Rerun dead).
+  // The menu is where muscle memory expects it in both states; for dead
+  // sessions the same action is deliberately mirrored as the primary
+  // button in ReplayView's action bar.
   const action = lifecycleAction(session, resuming)
   const actionHandler = action?.id === 'restart' ? onRestart
     : action?.id === 'resume' && onResume ? () => onResume(session.id)
@@ -749,6 +751,8 @@ function App() {
           <ReplayView
             session={selectedVal}
             terminalOptions={termOpts}
+            onResume={handleResume}
+            resuming={resumingId === selectedVal.id}
           />
         ) : selectedVal ? (
           <div class="state-message">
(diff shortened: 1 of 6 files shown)


mgabor3141:
Re Greptile's P1 ("Footer still renders" / duplicate lifecycle chrome): the Resume button living in **both** the ReplayView action bar and the header menu is intentional — the bar is the primary affordance for a dead session, the menu entry sits where Restart muscle memory lives for alive sessions.

The genuinely redundant part was the status label (header "Exited (N)" chip + footer "Session ended"). Fixed in d1be432: the footer now carries only the action button, sourced from the same `lifecycleAction` as the menu, and is omitted entirely for non-resumable sessions (body-only in that case).

PR #354, opened 2026-07-06
feat(web): find-in-terminal (search) via xterm SearchAddon

mgabor3141:
Adds find-in-terminal to the gmux web terminal.

## What

- **SearchAddon**: integrates `@xterm/addon-search@0.17.0-beta.195` (same beta line as the other `@xterm/*` deps), loaded alongside fit/image/web-links in `TerminalView`.
- **Find bar** (`terminal-find.tsx`): floats top-right over the terminal (sticky anchor, same pattern/design language as the resize pill). Incremental search while typing, `n/total` match counter (red on no match), next/prev via Enter / Shift+Enter or ‹ › buttons, ✕ / Escape closes. Matches highlighted via the addon's decorations (dim yellow, active match orange) + overview-ruler marks. Closing clears decorations and hands focus back to the terminal (desktop only — on touch that would re-pop the OSK).
- **Keybind**: new `find` action, default `secondary+f` (Cmd+F on macOS, Ctrl+F elsewhere). Declared in the schema like the other actions, so it's overridable / disablable (`action: "none"`) from settings.jsonc.
- **Menu entry**: "Find in terminal" in the session **⋮** menu (the closest equivalent to a "…" overflow menu in this app), so the bar is reachable on touch devices without a hardware keyboard. Shown for alive sessions.
- Opening is wired through a `terminalFindOpen` signal in store.ts (same pattern as `terminalScrolledUp`), so both the key handler and the menu can open the bar without threading callbacks.

## Mobile

- Coarse-pointer styles: 38px tap targets, compact input so the bar doesn't cover half the terminal.
- The input focuses on mount (pops the OSK — typing the query is the immediate next step); search-shaped input attrs (`enterkeyhint=search`, autocorrect/autocapitalize off).
- The shell's touch focus-hold handler exempts the find bar so its input can actually gain focus.

## Verified

- `pnpm lint` (tsc + biome), `pnpm test` in `apps/gmux-web` (550 tests, incl. new keybind default/disable tests).
- Manual via dev stack + headless Chrome:
  - Desktop: Ctrl+F opens+focuses, incremental highlight, counter, Enter/Shift+Enter/buttons step with wraparound, Escape closes and refocuses terminal, browser-native find stays suppressed while the input is focused.
  - Off-screen matches scroll into view (verified jump from bottom to scrollback top).
  - Mobile emulation (390×844, coarse+touch): ⋮ → Find in terminal opens the bar, typing/stepping/closing all work with the OSK heuristics untouched.

## Notes / limitations

- When the terminal is in passive "sized for another device" mode on a small screen, the addon scrolls xterm's own viewport to the match, but the shell's pan offset (the overflow scroll used in that mode) is not adjusted — the match can sit outside the panned area. Normal owned-size sessions are unaffected.
- Decoration colors are fixed (dark-theme oriented) rather than theme-derived; the addon needs concrete colors and the app currently has no themed-highlight tokens.

**Screenshots** (from headless verification): desktop shows the find bar top-right with `needle` 1/3 — active match orange, other matches yellow, ruler marks on the scrollbar; mobile shows the bar under the header with enlarged tap targets and the toolbar unaffected. (Captured locally; not attached — ask if wanted.)

### DIFF
--- a/apps/gmux-web/package.json
+++ b/apps/gmux-web/package.json
@@ -19,6 +19,7 @@
     "@preact/signals": "^2.9.0",
     "@trpc/client": "^11.6.0",
     "@xterm/addon-fit": "0.12.0-beta.195",
+    "@xterm/addon-search": "0.17.0-beta.195",
     "@xterm/addon-image": "github:gmuxapp/xterm.js#v6.1.0-beta.195-gmux.2&path:addons/addon-image",
     "@xterm/addon-web-links": "0.13.0-beta.195",
     "@xterm/addon-webgl": "0.20.0-beta.194",
--- a/apps/gmux-web/src/keybinds.test.ts
+++ b/apps/gmux-web/src/keybinds.test.ts
@@ -135,6 +135,21 @@ describe('resolveKeybinds', () => {
     expect(keys).toContain('ctrl+alt+t')
   })
 
+  it('includes the find keybind on secondary+f', () => {
+    const resolved = resolveKeybinds(null)
+    const find = resolved.find(r => r.action === 'find')
+    expect(find).toBeDefined()
+    expect(find!.baseKey).toBe('f')
+    // "secondary" resolves to exactly one platform modifier.
+    expect(find!.ctrl || find!.meta).toBe(true)
+    expect(find!.ctrl && find!.meta).toBe(false)
+  })
+
+  it('allows disabling the find keybind', () => {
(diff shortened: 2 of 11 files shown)


mgabor3141:
The failing `Lint, build, test` check is pre-existing main breakage (`go vet`: `wait_test.go` still uses the pre-#353 `Kind` field, introduced by the #293 merge race) — fix in #355, which is fully green. Re-running this job after #355 merges should clear it; all other checks (incl. E2E) pass here.

PR #380, opened 2026-07-07
feat(acp): stream assistant text end-to-end (ACP session/update tracer)

mgabor3141:
## Tracer #1 — streaming assistant text, end-to-end (ADR 0021)

The foundational HITL tracer for **ADR 0021** (ACP `session/update` as gmux's internal conversation schema). It proves schema + transport + renderer on the narrowest slice — **streaming assistant text only** — and sets the conventions later slices build on. Wire contract documented in **`docs/acp-conversation-stream.md`**.

### Flow
```
pi extension (read-only, one-way, ordered)
  → POST /acp/ingest (token deltas)
  → runner acpHub: synthesizes ACP session/load snapshot + session/update stream
  → runner /acp WebSocket (snapshot-then-stream, mirrors PTY attach, ADR 0004)
  → gmuxd /acp/{id} proxy
  → gmux-web conversation panel (live assistant text)
```

### Content ownership (ADR 0011 / 0016 intact)
- Daemon holds **no** conversation content — it's a transparent WS proxy.
- Runner keeps only the **in-memory unwritten tail** (current partial assistant message) and **forgets it** on `message_end`, once pi has persisted it to JSONL.
- `session/load` = durable JSONL history **+** the in-mem tail; live stream = tokens.

## Conventions established — please review

1. **WS framing.** JSON-RPC 2.0 objects over a WebSocket. Endpoint: runner `/acp`, proxied at gmuxd `/acp/{sessionId}`. The server **pushes `session/load` as the first frame** (the history snapshot), then streams `session/update` notifications — no client→server `session/load` request this slice; the snapshot is server-initiated exactly like the PTY `renderScreen` frame.

2. **Schema home.** Go types in `cli/gmux/internal/acp` (runner-only; the daemon doesn't parse). TS types local to `gmux-web` (`conversation.ts`). Deliberately minimal: only `session/load` + `session/update: agent_message_chunk`.

3. **Token volume / ordering.** Per-token on the wire. The extension **serializes** its ACP posts through a promise chain so token order is preserved (independent fire-and-forget POSTs can otherwise arrive out of order and corrupt the text). Runner broadcast is non-blocking and **drops frames for a slow subscriber** (backpressure). Frontend **coalesces** deltas into one render per microtask burst.

[...]

PR #383, opened 2026-07-07
feat(acp): thinking (agent_thought_chunk) + markdown/code rendering (slice #2)

mgabor3141:
Slice #2, stacked on the tracer (#380, `feat/acp-tracer-streaming`) — **base this PR against that branch, not `main`.** Adds two things end to end, additively:

## A. Thinking — `agent_thought_chunk`
Streams pi's reasoning token deltas as the ACP `agent_thought_chunk` update variant, parallel to `agent_message_chunk`.

- **Verified pi's event shape** (not guessed): pi-ai's `AssistantMessageEvent` union (pi 0.80.3) emits reasoning as `thinking_start` / `thinking_delta` / `thinking_end`, incremental text in `.delta`; it persists to JSONL as a `{type:"thinking", thinking, thinkingSignature?}` content block.
- Extension forwards `thinking_delta` as `{op:"thinking_chunk"}` over `/acp/ingest`, on the same ordered promise-chain as text.
- `acpHub` now holds the unwritten tail as **ordered content blocks**, so interleaved thinking+text keep their arrival order; broadcasts the matching `session/update` variant per delta.
- `session/load` snapshot includes in-flight thinking; `history.go` surfaces thinking blocks from durable JSONL in message order.
- Schema additions are additive: new `UpdateAgentThoughtChunk` / `ContentTypeThinking` values, no change to existing shapes.

## B. Markdown + code rendering
Assistant text (and thinking) now render as **markdown with fenced-code syntax highlighting**, natively in Preact.

- `markdown.ts`: `markdown-it` + `highlight.js`. Picked `markdown-it` over `marked` because **`html:false` escapes raw HTML by default** — that's the injection boundary for model output, with no DOM sanitizer needed (and none would run in the node test env). markdown-it also validates link schemes (blocks `javascript:`).
- **Streaming-safe:** incomplete markdown (unterminated ``` fences, half links, dangling emphasis) degrades gracefully; `render()` is wrapped defensively regardless. Tests cover the mid-stream shapes.
- Thinking renders as a dimmed, collapsible `<details>` block; text as inline markdown.

## Tests
- Go: `agent_thought_chunk` shape; hub broadcasts thinking + keeps interleaved thinking/text ordered in the tail/snapshot; `LoadHistory` surfaces thinking blocks; extension forwards `thinking_delta`→`thinking_chunk` and ignores start/end/toolcall events.
- Web: markdown highlight/HTML-escape/dangerous-scheme/streaming-partial; store thinking accumulation; view markdown + thinking + injection + partial-fence.
- `pnpm test` (582 web tests), Go suites, `tsc`, `vite build` all clean.

## Watch for — native-vs-island judgment
**Native still clearly right.** Rendering markdown + thinking natively was genuinely cheap and not fiddly:

- **LOC:** `markdown.ts` ~40 lines; view changes ~30 lines; store change ~15 lines; CSS ~90 lines. Well within the "few hundred lines" estimate — no hidden complexity.
- **Streaming-partial markdown**, the thing I expected to bite, was a non-issue: markdown-it degrades gracefully on unterminated fences/half tags, so re-rendering on every delta "just works" with one defensive try/catch. A chat framework would have absorbed nothing here that markdown-it doesn't.
- **Interleaving** (thinking then text in one message) fell out naturally from ordered content blocks — same model on both the Go tail and the JS store.
- **Nothing steers us to adopt the `@assistant-ui/react` island sooner.** The island's value is rich/generative tool-call UIs (slice #3+), and none of that surfaced here. Reinforces the plan: stay native now, hold the island for when tool UIs land.
- **Dep weight — the one honest cost:** `markdown-it` ≈51KB gzip + `highlight.js/lib/common` ≈52KB gzip ≈ **104KB gzip**. highlight.js/common (≈35 languages) is the heavy half and could be trimmed to a curated language set later. Mitigated now by **lazy-loading the whole panel behind `?conv`**: the main bundle is unchanged (237KB gzip) and the renderer is a separate 102KB chunk loaded only when the conversation panel opens.

## Out of scope (unchanged)
`tool_call`/`tool_call_update`, `session/prompt`/`session/cancel`, peer proxying, the React island itself.


mgabor3141:
Consolidating the ACP conversation-UI stack into a single PR (#388) for easier iteration. All commits from this PR are included in #388, which now targets `main` and will be squash-merged in one go.

PR #388, opened 2026-07-07
feat(acp): conversation UI — streaming text/thinking/tool-calls in an assistant-ui island

mgabor3141:
## What

Adds a **conversation UI** for gmux sessions: a streaming, read-oriented view of
an agent's turn (assistant text, thinking, and tool calls) rendered from an ACP
`session/update` stream, plus a composer that sends prompts back as keystrokes.
Implements ADR 0021 tracer #1 end-to-end (extension → runner hub → daemon proxy
→ web island). This PR is self-contained against `main` (it absorbed the
now-closed stack PRs #380/#383).

## How it fits together

- **Extension** (`cli/gmux/internal/agentext/pi-ext.mjs`): read-only, one-way.
  Hooks pi events (`session_start`, `agent_thought_chunk`, `tool_execution_*`,
  message chunks) and posts facts to the runner's unix-socket `/acp/ingest` in
  order (promise-chained). Translates pi tool names → ACP `ToolKind`.
- **Runner hub** (`internal/ptyserver/acp.go`): holds only the unwritten
  in-flight tail, forgets it on `message_end`. `attach()` delivers an ADR
  0004-style snapshot-then-stream under one lock; broadcast is non-blocking
  (drop-on-full).
- **Schema** (`internal/acp/*`): additive ACP `session/update` variants —
  `agent_message_chunk`, `agent_thought_chunk`, `tool_call`, `tool_call_update`
  — with `ToolKind` mirrored from the ACP spec. Daemon stays content-free (a
  transparent WS pipe).
- **Web island** (`apps/gmux-web/src/conversation-*`): a lazily-loaded, isolated
  **React 18** subtree (assistant-ui) mounted from the Preact app. Renders
  markdown + Prism code, collapsible thinking, kind-keyed tool rows
  (Executing/Executed…), a shimmer on the active verb, and a "Working…" pill.
  The composer writes via the existing daemon `/input/{id}` keystroke actuator
  (no `session/prompt` duplex), so it works with no terminal mounted.

## Composer features

- **Send** composes text and posts it as keystrokes + Enter.
- **File attachments**: drag/drop or `+`, uploaded via the existing
  `/v1/sessions/{id}/clipboard` endpoint to a `/tmp` path on the owning gmuxd;
  the path is spliced (bare, no bracketed-paste escapes) into the outgoing text.
- **Optimistic echo** of the sent user turn, anchored by store position so it
  renders above the reply; dropped once the authoritative snapshot arrives.
  Failed sends roll back the echo, restore the composer text, and surface an
  inline error.

## Bundle isolation

The island is a lazy chunk (~178 kB gzip: react-dom + @assistant-ui/* + Prism),
loaded only behind `?conv` on pi sessions. The main bundle contains **zero**
react/assistant-ui references.

## Notable behavior change

The header "Working…" chip is gone — but `main` (96059b64) independently removed
it for the same reason (redundant with the session dot indicator); this PR keeps
that and adds the "Working…" pill above the composer inside the conversation view.

## Testing

- Go: `go build/vet/test ./...` (incl. `-race`) in cli/gmux + services/gmuxd.
- Web: `tsc --noEmit` clean; 624 vitest tests (adapter/store/attachments,
  including a regression test pinning the echo-ordering fix); prod build.

- Slash-command autocomplete (extension `getCommands()` + `BUILTIN_SLASH_COMMANDS`
  → ACP `available_commands` → assistant-ui slash adapter) — scoped, next slice.
- Playwright e2e for the island (adapter/store are unit-tested).
- Typed ACP `diff`/`terminal` tool content (render edits as real diffs).
- Runner tail: split/expire on messageId change over the lossy ingest channel
  (review P2 #7).

Reviewed independently (REQUEST-CHANGES); the four P1s (echo ordering, session
isolation, send-failure handling, rebase) and the cheap P2 doc/comment nits are
addressed here.

## 2026-08-02 rebase note

Rebased onto current `main` (`34930b07`) and reconciled with the post-v2 architecture:

- moved gmuxd `/acp/{id}` and `/input/{id}` registration into `serveCentral`, using the central registry/session identity resolver;
- adopted `ConversationRefSnapshot` and current `status.active` semantics;
- preserved main's ordered hook-delivery/turn-frame machinery and merged ACP streaming into the same pi message callbacks (pi permits one callback per event name);
- kept current routing/home semantics rather than restoring the removed project-view branch;
- regenerated the current pnpm lock and pinned the mutually compatible assistant-ui 0.14 package set.

The branch remains a pi-backed tracer: `adapter === pi`, PTY input, and the custom snapshot stream are intentionally still present. The pi/ACP fit investigation and recommended follow-up seams are recorded in `.memory/report-388-pi-acp-fit.md` in the rebase workspace; no ACP-host functionality was added here.

Local evidence after rebase: web TypeScript clean, 658 Vitest tests pass, production build succeeds; `go test ./...` passes in both `cli/gmux` and `services/gmuxd`.


mgabor3141 (commented):
## Independent review (read-only) — verdict: **REQUEST-CHANGES** (consistent with the draft status)

The architecture is sound and tracks ADR 0021 faithfully — daemon stays content-free (transparent WS pipe), runner holds only the unwritten tail and forgets it on `message_end`, write path stays keystrokes via `/input`, extension stays read-only/one-way with ordered posts. Go side is solid: `go vet` clean, all tests pass incl. `-race`. Web: tsc clean, 604 tests pass, prod build OK.

**Runtime-verified** (prod build against a live gmuxd on an isolated port, real pi session): toggle gating, terminal↔conversation mount-swap, composer → `POST /input/{id}` → PTY (works with no terminal mounted), `/acp/{id}` snapshot-then-stream (checked via websocat + browser), streaming thinking/markdown/Prism-code/tool rows, reconnect tail restore with messageId continuity. Bundle claim verified: island chunk **177.79 kB gzip**, main `index-*.js` contains zero react/assistant-ui references.

### P1 — should fix before undrafting

1. **User echo renders after the assistant's reply.** `[...store.getMessages(), ...echoes]` means once assistant deltas arrive, the just-sent user turn displays *below* the reply. Reproduced live. TODO #1 covers de-dupe but not this ordering defect — it hits the normal send flow.
2. **Echoes + store leak across session switches.** `ConversationView` isn't keyed by session; the store is created once and `echoes` are never cleared — switching sessions with the panel open shows session A's messages (and A's echoes, indefinitely) in session B. Suggest `key={sessionId}` + reconciling echoes on `session/load`.
3. **`onSend` failure = unhandled rejection, no user feedback.** `void sendSessionInput(...)` — on `!res.ok` the composer still clears and the echo still appends. Catch + toast.
4. **Rebase needed:** confirmed content conflict in `apps/gmux-web/src/main.tsx` vs current main (`96059b64`).

### P2 — nits

5. **Description ≠ diff** (it describes the delta vs closed #383): there is no `WebSpeechDictationAdapter`/mic in the code (only orphaned `.conversation-composer-mic` CSS); nothing is removed vs main; the composer uses `sendSessionInput` → daemon `/input` proxy, **not** a `handleMobileInput` bridge — which also means TODO #7 (detached send path) appears already solved. Please rewrite against main.
6. Stale references: main.tsx lazy comment ("markdown-it + highlight.js"), `docs/acp-conversation-stream.md` → `src/markdown.ts` (doesn't exist), styles.css "floating overlay / decision deferred" header.
7. **acpHub tail ignores messageId changes** (repro'd by injection): chunks with different messageIds and no `message_start`/`message_end` coalesce into one tail block that then haunts every snapshot until the next `message_end`. Frontend store splits on messageId change; hub should too (lossy ingest channel).
8. Header "Working…" chip removed for *all* surfaces, not just when the panel is open — undocumented behavior change for terminal-view users.
9. `supportsConversation` hardcodes `adapter === 'pi'` (acknowledged inline; fine for the tracer).
10. `/input` proxy silently truncates >1 MiB bodies rather than 413.
11. One-off dev-mode crash on first toggle after HMR churn (`… reading 'some'` in preact diff via the inline-lazy island — first inline `lazy()` outside a `<Route>` in the app). Not reproducible on a clean dev session or prod build; noting for the record.
12. No island/view component tests remain; the echo-ordering bug is exactly what the planned e2e (TODO #6) should pin.

### Interactions with open PRs

- **#396/#397**: guaranteed `pnpm-lock.yaml` conflicts (+~90 packages here) + adjacent `package.json` lines — second-lander regenerates.
- **#394**: same files (`pi-ext.mjs`, `ptyserver.go`) — expect textual conflicts; coordinate ordering.
- **#393**: touches gmuxd `main.go` `serve()` region — low conflict risk.
- **#223**: don't gate the release on this; when it lands, the new dependency tree (react/react-dom/@assistant-ui/radix) is worth a release-notes callout.

Security checked: `/input` & `/acp` 401 unauthenticated, POST/WS same-origin-enforced for cookie auth (ADR 0020 intact), `/acp/ingest` never exposed by the daemon. Full findings in the reviewer's local report.

PR #390, opened 2026-07-07
fix(gmuxd): bump last_activity_at on raw terminal output

mgabor3141:
## Problem

The "last activity N ago" shown for a session was stale: an actively-used session (producing output *today*) displayed **"last activity 3 days ago."** The timestamp reflected roughly session-creation time, not recent activity.

## Root cause

The displayed field is `last_activity_at` (`Session.LastActivityAt` in `services/gmuxd`, stamped by the daemon; rendered by `session-row.tsx` via `formatAge`). It's bumped by `shouldBumpActivity` only on four *edges*: exit (`alive` true→false), `unread` false→true, `status.working` false→true, `status.error` false→true.

An **interactively-used session produces none of these edges** — a shell you're typing in, or an agent streaming output while its tab is focused (so `unread` stays false and `working` doesn't re-toggle). Its only signal is the runner's transient `activity` SSE event, which the discovery handler re-broadcast as the live activity dot **without mutating state**. So `last_activity_at` froze at the first post-creation transition (~creation time).

## Fix

New `Store.BumpActivity(id)`, called from the `"activity"` case in `discovery/subscribe.go`, freshens `LastActivityAt` on real terminal output and broadcasts a `session-upsert` so the UI/CLI observe it.

**Write cadence** (no fsync-per-keystroke): the runner already throttles `activity` to ≤1/2s; `BumpActivity` adds a coarse 60s gate (`activityBumpInterval`), so a busy session emits at most one `session-upsert` per minute. Since the UI renders `<60s` as "now", worst-case staleness is imperceptible.

**Edge cases:**
- **Peer sessions** never bump here — peer activity is relayed as a separate namespaced `session-activity` broadcast (`peering/peer.go`); the owning daemon stamps the timestamp and `UpsertRemote` preserves it as-received.
- **Dead sessions** are guarded: a final buffered `activity` event can race the `exit` event over the same SSE stream, so `BumpActivity` ignores non-alive sessions — a dead session freezes at its last real activity. Persisted via sessionmeta, survives restart.
- **Unknown/removed session**: safe no-op (activity can race removal).

Consistent with the existing design that brand-new sessions don't bump (avoids stamping every rehydrated dead session at startup).

## Tests

- `store_test.go`: `TestBumpActivity_FreshensInteractiveSession` (the regression — a 72h-stale alive session with no transition gets freshened), `_ThrottledWithinInterval`, `_BroadcastsOnStamp`, `_IgnoresDeadSession`, `_UnknownSessionNoPanic`.
- `subscribe_test.go`: `TestActivityEventBumpsLastActivity` — wire-level proof that an `activity` event from a live runner bumps the persisted timestamp (the actual gap).

Each regression test verified to fail without its corresponding fix.

## Checks

`go test ./...` + `go vet ./...` (gmuxd) ✅ · `pnpm lint` ✅ · `pnpm -C apps/gmux-web test` ✅ (580)

### DIFF
--- a/services/gmuxd/internal/discovery/subscribe.go
+++ b/services/gmuxd/internal/discovery/subscribe.go
@@ -322,8 +322,12 @@ func (sub *Subscriptions) handleEvent(sessionID, socketPath, eventType string, d
 		})
 
 	case "activity":
-		// Transient signal: terminal produced output with no attached clients.
-		// Forward to the store's subscribers without mutating state.
+		// Raw terminal output. Bump the persisted last-activity timestamp
+		// (coarsely throttled inside the store) so interactively-used
+		// sessions surface as recently active even when they trigger no
+		// alive/unread/status transition — then forward the transient
+		// signal to the store's subscribers for the live activity dot.
+		sub.store.BumpActivity(sessionID)
 		sub.store.Broadcast(store.Event{
 			Type: "session-activity",
 			ID:   sessionID,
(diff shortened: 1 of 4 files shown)


mgabor3141:
This was misdiagnosed, closing

PR #393, opened 2026-07-08
fix(gmuxd): reseed last_activity_at from file mtime on daemon restart

mgabor3141:
## Problem

When the daemon restarts and re-adopts a **surviving** runner, the session's `last_activity_at` is lost and the UI falls back to `created_at` → "3 days ago".

Root cause: an **alive** session's `LastActivityAt` is never persisted — only *dead* sessions land in sessionmeta. On restart the alive session isn't in the sweep, so `discovery.Register` takes the fresh-upsert path and stores it with an empty stamp (`/meta` doesn't carry `last_activity_at`, and a brand-new session doesn't bump). Dead sessions were already fine (their real stamp is persisted and swept back).

## Fix

Seed `LastActivityAt` from the newest durable on-disk activity proxy whenever a locally-owned session is upserted **without** a stamp:

1. **`Session.ConversationFile` mtime** (primary) — the conversation file is appended on every message, so its mtime is an accurate, restart-surviving proxy. The path is adapter-supplied (agent hook, ADR 0011); the daemon only stats it, so there's **no adapter-specific reseed logic** (resolves the adapter-vs-generic fork).
2. **scrollback tee mtime** (generic fallback) — every session has one, incl. plain shells.

Applied only as a **floor** for empty stamps (never clobbers a live or sessionmeta-restored value), capped at `now` (clock-skew defense), and skipped for peer payloads (owning daemon stamps its own).

## Notes

- Does **not** build on #390 (`BumpActivity`) — that PR is closed/unmerged ("misdiagnosed"). This targets the real restart data-loss.
- No wire/web change: `conversation_file` and `last_activity_at` already exist.

## Tests

- store: reseed-empty (regression), floor-not-override, peer-skipped.
- seed helper: conv-file mtime, scrollback fallback, newest-wins, no-files→empty, future-mtime capped.
- `go build/vet/test ./...` in `services/gmuxd` all green.

### DIFF
--- a/services/gmuxd/cmd/gmuxd/activityseed.go
+++ b/services/gmuxd/cmd/gmuxd/activityseed.go
@@ -32,6 +32,17 @@ import (
 // empty and the UI would fall back to created_at ("3 days ago"). The
 // store applies this value only as a floor for empty stamps, so it can
 // never overwrite a live or persisted timestamp.
+//
+// Note: the store can't tell a restart-rehydrate from a first-ever
+// registration — both arrive with an empty stamp — so a brand-new
+// session is seeded too. That is harmless: its files' mtimes are ~now,
+// which is the same instant the UI would show via the created_at
+// fallback anyway, and the first real state transition bumps the stamp
+// forward. The one visibly-odd case is resuming an *old* conversation
+// when the scrollback tee failed to open: the seed would then be the old
+// conversation file's mtime (days ago) until the first transition
+// corrects it. Rare and self-healing, so we accept rather than guard —
+// there is no reliable restart-vs-fresh signal at this layer.
 func activitySeedFor(sess store.Session, sessionDir func(string) string) string {
 	var newest time.Time
 	consider := func(path string) {
--- a/services/gmuxd/cmd/gmuxd/activityseed.go
+++ b/services/gmuxd/cmd/gmuxd/activityseed.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/gmuxapp/gmux/packages/scrollback"
+	"github.com/gmuxapp/gmux/services/gmuxd/internal/store"
+)
+
+// activitySeedFor returns an RFC3339 timestamp to seed a rehydrated
+// session's LastActivityAt from the newest durable on-disk activity
+// proxy. Two sources, newest wins:
+//
+//   - the adapter's conversation file (primary): appended on every
+//     message, so its mtime is an accurate, restart-surviving proxy for
+//     when the session last did something. The path is adapter-supplied
+//     (session.ConversationFile, reported by the agent hook per ADR
+//     0011); the daemon only stats it, so no adapter-specific reseed
+//     logic is needed.
+//   - the runner's scrollback tee (generic fallback): every session has
+//     one under the per-session dir, written live by the runner. Covers
+//     plain shells and agent sessions whose conversation file hasn't
+//     appeared yet.
+//
+// Returns "" when neither file yields a usable mtime.
+//
+// This recovers "last activity" across a daemon restart. An alive
+// session's LastActivityAt is never persisted (only dead sessions land
+// in sessionmeta), so on re-register the store would otherwise leave it
+// empty and the UI would fall back to created_at ("3 days ago"). The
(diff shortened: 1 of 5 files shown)


mgabor3141 (commented):
## Independent review — verdict: **APPROVE-WITH-NITS**

Correct, narrowly scoped fix; tests pin the actual claims. All findings are P2 nits — nothing blocks merge.

### Verified
- **Root cause is real**: the runner's `/meta` never reports `last_activity_at` (no `LastActivityAt` anywhere under `cli/`), so a surviving runner re-registering after a daemon restart lands as a fresh `Upsert` with an empty stamp, exactly as described.
- **Gating is right**: the seed runs only when `bumpLocally && LastActivityAt == ""` *after* bump/carry-forward — a pure floor. Peers (`UpsertRemote`) skipped; sessionmeta-restored stamps untouched.
- **Paths agree**: runner tee at `paths.SessionDir(id)/scrollback` vs seed's `metaStore.SessionDir(id)` (`DefaultDir() == paths.SessionsDir()`) — same file. `SessionDir` contains malformed IDs, and the seed only stats, never writes.
- **Tests would fail without the fix** (`TestActivitySeed_ReseedsEmptyStamp`); floor/peer gating and all mtime-selection edges pinned. `go build/vet/test ./...` green locally; CI green.
- No conflicts with #397/#396/#394/#388/#223 (#388 touches `main.go` ~1300 lines away, no semantic overlap).

### Findings

**P2 — filesystem stat under the store write lock.** `upsertCommon` calls `s.activitySeed(sess)` while holding `s.mu`. `ConversationFile` is an arbitrary adapter-supplied path (the comment itself mentions remote filesystems); a hung NFS/FUSE stat would stall the whole store (all reads + SSE fan-out), not just this upsert. Cheap to avoid: compute the candidate seed before locking when the incoming stamp is empty, apply under the lock only if still empty after carry-forward.

**P2 — the hook also fires for brand-new sessions.** The mechanism can't distinguish restart-rehydrate from first registration: a new session registers with an empty stamp and its tee already open, so it gets `last_activity_at ≈ now`. Harmless (UI falls back to `created_at` anyway), but it quietly bypasses the `TestLastActivityAt_NewSessionDoesNotBump` invariant in production, and in the resume-old-conversation + tee-open-failure corner a just-created live session could show a days-old stamp. One sentence in the `activitySeedFor` doc comment acknowledging this would do.

**P2 — no wiring test.** Store hook and helper are unit-tested, but nothing exercises the end-to-end claim (restart → `discovery.Register` re-register → seeded stamp). The main.go wiring is 3 lines, so acceptable — an integration test would make the pin airtight.

### Pre-existing, not from this PR
`go test -race ./cmd/gmuxd/` fails (`TestBrokerTailParam*`, data race in `charmbracelet/x/vt` Emulator Close vs Read via `scrollback.RenderTail`) — reproduced identically on main. Worth filing separately.

mgabor3141:
Thanks for the review. Addressed all three P2s in `f7e10ca`:

**(1) stat under the store write lock — fixed.** The seed is now computed *before* `s.mu.Lock()` and applied under the lock only if the stamp is still empty after carry-forward, so it stays a pure floor while the (possibly remote/FUSE) `os.Stat` happens outside the lock. A hung stat can no longer stall reads / SSE fan-out.

**(2) hook fires for brand-new sessions — verified harmless, documented.** Confirmed it can't be distinguished from rehydrate at this layer, and the effect is benign: file mtimes are ~now, which is what the UI already shows via the `created_at` fallback, and the first real transition bumps it forward. I documented this in the `activitySeedFor` doc comment, including the one visibly-odd corner you flagged (resume-old-conversation + tee-open-failure → stale-until-first-transition), and the rationale for accepting rather than guarding (no reliable restart-vs-fresh signal here). Declining the guard on purpose.

**(3) no wiring test — added.** `TestActivitySeedWiring_ReseedsRehydratedSession` mirrors `serve()`'s exact wiring (`SetActivitySeed` + `activitySeedFor` + `SessionDir`) and drives a real store `Upsert` against a real on-disk scrollback tee — the rehydrate path both Sweep-restore and Register re-register funnel through. Ties together the two halves the isolated unit tests each covered separately.

`go build/vet/test ./...` green; `go test -race ./internal/store/` green. Noted the pre-existing `vt.Emulator` race in `TestBrokerTailParam*` is unrelated (reproduces on main) — will file separately.

PR #399, opened 2026-07-12
feat: unified turn model — every session gets active/idle/waiting-on-you and works with gmux wait

### LINKED ISSUE #373: send --wait for shell sessions (idle via OSC prompt events)
## Summary

Extend `gmux send --wait` to support **shell** sessions. Today `--wait` only works for agent sessions (which report a turn boundary via hooks); shells have no notion of "done", so `--wait` on a shell is unsupported.

## Semantics

`wait` (and the `--wait` flag on `send`) always means: **block until the session is idle.** Idle is a single, first-class runner-owned session state.

- **Idle includes process exit** — a session whose process has exited is idle.
- **For shells, idle is determined via OSC prompt events.** The shell integration emits OSC markers (e.g. OSC 133 prompt start/end) on each prompt cycle; `wait` considers a shell idle once the prompt has returned (command finished and the shell is back at a fresh prompt). This gives shells the same crisp busy→idle signal that agents already get from the turn-end hook.

## Why `send --wait` rather than `send X; wait X`

`--wait` must remain a single call, not sugar for two. Splitting into `send X` then `wait X` introduces a **race**: the command can start and finish (or the prompt can return) in the window between the two calls, so `wait` may observe an already-idle session and return immediately, having missed the busy period it was meant to bracket. `send --wait` brackets the send and the wait atomically on the runner, with no such gap.

## Scope

- Give shell sessions a busy/idle signal sourced from OSC prompt markers (busy on prompt-end/command-start, idle on next prompt-start).
- Wire that state into the same idle primitive `wait` / `send --wait` already use for agents.
- `send --wait` on a shell blocks until the prompt returns (or the process exits).

## Related

Part of the CLI framing where each session verb has a universal surface with a per-adapter-sourced meaning: `tail` (clean vs raw output), `send` (raw input), `wait` (busy/idle). Agents source idle from the turn-end hook; shells will source it from OSC prompt events.

mgabor3141:
## What

Closes #373 — and generalizes it. Instead of a shell-specific idle signal bolted onto the agent wait machinery, every session now runs **one turn state machine** — *active / idle / waiting-on-you* — with the turn **source** chosen by adapter kind. Design record: **`docs/adr/0023-unified-turn-model.md`**.

| Session kind | turn opens | turn closes |
|---|---|---|
| hook-driven adapter (`SessionExtender` / `SessionHookCommand`) | hook turn-start | hook turn-end |
| default model, **upgraded** by observed OSC 133 prompt marks | `133;C` (command starts) | `133;A`/`133;D` (prompt returns) |
| default model, no marks ever (**lifetime-as-turn**) | process launch | process exit |

What that buys concretely:

- `gmux wait` / `gmux send --wait` work on **every session**: agents (unchanged), integrated shells (per-command), and one-shots — `gmux -d -- pnpm test && gmux wait $id` blocks until the test run exits, rc 0.
- Shells and one-shots get full **agent parity in the sidebar**: blue active dot while a command/process runs, "waiting on you" on completion, error flag data on a failed one-shot (`Status.Error` when exit ≠ 0).
- The adapter surface **shrinks**: `Monitor(output []byte) *Event` (stubbed `return nil` by all five built-ins — dead code), the `Event` type, `BoolPtr`, and the `PromptSignaler` capability are deleted. Base interface is now `Name/Discover/Match/Env`; there is no status capability to implement at all.

## Design decisions (each argued in ADR 0023)

- **Evidence over declaration.** The upgrade to prompt-cycle turns happens on the first *observed* mark, not by argv classification. `bash -c script` (shell binary, really a one-shot) and `ssh host` with remote shell integration (non-shell argv, marks flow through the PTY) both classify correctly only by evidence. Sessions never switch adapters mid-life, so there's no conflict window; hook-driven sessions ignore marks entirely (a tool the agent runs may print them).
- **Turn-state-at-death.** The exit event no longer clears `Status`; it records whether the turn was closed at exit. `wait` resolves death by it — closed turn → `idle`/rc 0 (one-shot completed, shell exited at its prompt, agent exited after its turn), open or never-demonstrated turn → `died`/rc 2. **Timing-independent**: a wait issued after the exit answers the same as one that watched it live. Agent contract change (deliberate): a *clean post-turn* agent exit is now rc 0; mid-turn deaths stay rc 2.
- **Every session is waitable — the `no_idle_signal` 422 is deleted.** A markless interactive shell is Working for its whole life, so a wait on it *blocks honestly* until exit or `--timeout` ("never provably idle" is answered by not answering). The CLI keeps its 422 branch for skew against older daemons.
- **Unread parity.** Every genuine working→idle transition sets "waiting on you", exactly like `applyTurnEnd` for agents. The shell's *initial* prompt closes only the launch phase and does not. Notifications can't double-fire: the notify router already coalesces per-session (finished preferred over unread).
- **Status snapshot replay.** The runner's `/events` now replays the current `Status` on subscribe — previously the launch-time `Working=true` (and any agent turn spanning a daemon restart) never reached a late subscriber.

## Bugs found & fixed on the way

1. **Activity pulses misread as removals.** The wait loops dispatched on `ev.Session == nil` to detect session-removal — but `session-activity` pulses also carry no `Session`. Any unwatched session that produced output while a wait was pending resolved it as a spurious `died`; the new model made it deterministic (a detached one-shot emits its final output pulse in the same breath as its exit). Fixed by dispatching on named event-type constants (now used at every producer/consumer); regression test `TestWaitIgnoresActivityPulse`. Found by live e2e — the unit suite never emitted activity pulses.
2. **Resume derivation destroyed turn state.** gmuxd's `OnExit` hook nil'd `Status` when setting a resume command ("clean resumable display"), which would have silently split the wait verdict by timing for agents: live wait on a cleanly-exited agent → `idle`, wait a moment later → `died`. The hook now touches only `Command` (regression test at the exit-event seam); `OnExit` drops its unused bool return.
3. **Dead sessions couldn't clear "waiting on you".** Every one-shot now dies unread, and dead sessions have no runner WS (the live path for clearing the flag) — the dot would persist until dismiss. The presence `OnSessionSelected` callback now clears unread + error dot on dead sessions when viewed; `Working` (turn state at death) survives. Live sessions stay runner-owned.
4. **Shared-pointer Status mutations** (subscribe.go's unread-clear, and the new view-clear) corrupted the store's previous snapshot and defeated its no-op change detection; `Status` is now replaced, not mutated in place.
5. **Pre-existing `scrollback.RenderTail` data race** (vt `Emulator.Close`'s unsynchronized flag vs the drain goroutine's `Read`) — tripped by every `--for-text` test under `-race`. Fixed by closing the response pipe's write end directly (everything `Close` does minus the racy bool); `-race` regression test replaying a DSR query.

## Tests

All new behavioral tests verified red without their change:

- **runner** (`ptyserver/promptmarks_test.go`, real bash children): mark-driven busy/idle pulse (fast command's C+A in one PTY read emits both edges); marks ignored for hook-driven adapters; unread set on a completed prompt cycle but not on the initial prompt; markless session keeps its lifetime turn open. `finalizeSessionState` extracted from `run()` so the load-bearing ordering (turn-close status + unread **before** the exit event) is pinned by unit tests; `/events` status replay covered at the HTTP seam.
- **daemon** (`wait_test.go`, `sendwait_test.go`, `subscribe_test.go`, `unread_test.go`): dead-session resolution by turn state (closed→idle / open→died / none→died, table-driven); markless live shell blocks to 408 instead of 422; any adapter with turn state is waitable (gate removal); one-shot exit resolves send--wait as idle even when only the dead upsert is observed; `send --wait 'exit'` on a prompt-cycle shell → died; activity-pulse regression.
- **manual e2e** (isolated daemon): one-shot success (live rc 0 @ exit, post-mortem rc 0, unread+idle persisted), one-shot failure (`error:true`, exit 3), markless interactive bash (timeout rc 3), synthetic prompt-cycle (wait blocks mid-command, idle at prompt, session alive), real `bash` with PS0/PROMPT_COMMAND integration (`send --wait 'sleep 2'` → 2.0s rc 0; `send --wait 'exit'` → rc 2 with `working:true` persisted at death).
- `go build` / `go vet` / `go test ./...` green across `packages/adapter`, `cli/gmux`, `services/gmuxd`.

## Compatibility

- No wire or on-disk format changes. `Status` is the same object; it's now also present on dead sessions (the web UI keys working/error dots on `alive` and renders "exited (N)" from `exit_code`, so nothing breaks — verified).
- Contract changes, all intentional and documented: `wait` on shells/one-shots (was 422), rc 0 for closed-turn exits (was rc 2 `died`), unread/error state for non-agent sessions.
- Adapters are compiled-in Go, so deleting `Monitor`/`Event`/`PromptSignaler` breaks no external plugin ABI; contributor docs rewritten.

## Docs

ADR 0023 (+ this PR's history: the first six commits are the original shell-specific implementation, superseded in-place after design review — kept for the record). UBIQUITOUS_LANGUAGE gains the turn-model terms. `reference/cli.md`, `skills/gmux/SKILL.md`, `develop/writing-adapters.md`, `develop/adapter-architecture.md`, `develop/terminal-data-pipeline.mdx`, `planned/opencode-adapter.md` updated.

- Web UI: single blue "active" indicator (gray activity dot is now redundant for default-model sessions); red error dot for dead failed one-shots (data is already there, `sessionDotState` currently gates on `alive`).

### DIFF
--- a/UBIQUITOUS_LANGUAGE.md
+++ b/UBIQUITOUS_LANGUAGE.md
@@ -23,6 +23,17 @@
 | **Attributed**          | Tool-backed session whose adapter file exists, giving it a real slug                                | Named                  |
 | **Fast-exit**           | A runner whose child finished before gmuxd's `queryMeta` reaches it; lands in the store as Alive=false directly, never as Alive=true | Quick-exit |
 
+## Turn model (ADR 0023)
+
+| Term                    | Definition                                                                                          | Aliases to avoid       |
+| ----------------------- | --------------------------------------------------------------------------------------------------- | ---------------------- |
+| **Turn**                | One unit of "the session is doing something": an agent turn (hook-delimited), a shell command (prompt-mark-delimited), or — default — the whole process lifetime. `Status.Working` is its open/closed state | Task, job              |
+| **Turn source**         | Where a session's turn boundaries come from: agent hook for `HookDriven` adapters, otherwise the runner's default model (lifetime, upgraded by observed OSC 133 prompt marks) | —                      |
+| **Active**              | Turn open (`Status.Working == true`); the blue dot. Distinct from *Alive* (process exists) and from the transient output-activity pulse | Busy, working (in prose) |
+| **Idle**                | Turn closed; what `gmux wait` waits for. Includes a closed turn at process exit (one-shot completed, shell exited at its prompt) | Done, finished         |
+| **Waiting on you**      | Unread flag set by a turn close; cleared when the session is viewed                                 | Unread (in UI prose)   |
+| **Upgrade**             | A default-model session's switch from lifetime-as-turn to per-command turns on the first observed OSC 133 prompt mark; permanent for the session's life | Promotion              |
+
 ## Components
 
 | Term            | Definition                                                                                       | Aliases to avoid            |
--- a/apps/website/src/content/docs/develop/adapter-architecture.md
+++ b/apps/website/src/content/docs/develop/adapter-architecture.md
@@ -25,7 +25,7 @@ That split is why the adapter system is a set of small interfaces instead of one
 | Adapter availability detection | `gmuxd` | `Adapter.Discover()` |
 | Command matching | `gmux` | `Adapter.Match()` |
 | Child env injection | `gmux` | `Adapter.Env()` |
-| PTY output monitoring | `gmux` | `Adapter.Monitor()` |
+| Turn state (default model) | `gmux` | launch/exit lifecycle + OSC 133 prompt marks (non-hook-driven adapters) |
 | Child self-report API | `gmux` | Unix socket HTTP endpoints |
 | Launch menu discovery | `gmuxd` | `Launchable` + compiled adapter set |
 | Conversation metadata resolution | `gmuxd` | `ConversationDescriber` |
@@ -43,7 +43,7 @@ When you run a command through `gmux`:
    - otherwise shell fallback
 2. `gmux` starts the child under a PTY
 3. `gmux` injects the standard `GMUX_*` environment variables
-4. `gmux` feeds PTY output into `Adapter.Monitor()`
+4. `gmux` applies the default turn model for non-hook-driven adapters (active from launch; OSC 133 prompt marks upgrade to per-command turns)
 5. `gmux` serves the session on its Unix socket (`/meta`, `/events`, terminal attach, child callbacks)
 6. `gmuxd` discovers the runner socket (the runner registers itself; a periodic scan is the fallback), queries `/meta`, and subscribes to `/events`
 
@@ -71,7 +71,6 @@ type Adapter interface {
     Discover() bool
     Match(command []string) bool
     Env(ctx EnvContext) []string
-    Monitor(output []byte) *Event // partial update: title, status, unread, cwd
 }
 ```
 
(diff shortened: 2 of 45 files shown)


mgabor3141:
Addressed the Greptile finding in 6414c46: the status-evidence gate is now scoped to shell sessions only (`sess.Adapter == "shell" && sess.Status != nil`) — other adapter kinds must be allowlisted deliberately, pinned by `TestWaitRejectsNonShellStatusEvidence`. Writing that regression test also surfaced a second bug in the dead-session resolution branch (it used `terminalReason`, whose idle arm would let a live no-signal session with a stale one-off Status return "idle" silently); it now resolves strictly on death.

mgabor3141:
Superseded the shell-specific design in place after a design review pass (commits from `refactor(adapter): drop the dead Monitor/Event surface` onward): every session now runs one turn state machine with per-adapter-kind turn sources — see the rewritten PR description and **docs/adr/0023-unified-turn-model.md**. The earlier no_idle_signal 422 / evidence gate is gone; `Monitor`/`Event`/`PromptSignaler` are deleted. Also fixes a latent wait bug (session-activity pulses misread as removals → spurious "died").

PR #404, opened 2026-07-12
feat(web): sidebar view menu, tab-scope filters, and activity view

mgabor3141:
Brings the home dashboard's usefulness into the sidebar and makes each browser tab a first-class, bookmarkable workspace. Stacked on #402.

## View menu

One arrange icon in the sidebar header opens a compact popover (inspired by t3-code's list controls: one entry point, coherent presets, instant switching):

- **View** — *Projects* (default: grouped by project, manual order) or *Activity* (flat, partitioned exactly like home: Waiting / Active / recency buckets, rendered at sidebar density via a compact `SessionRow`).
- **Host** — narrow the tab to one host.
- **Alive only** — hide dead-but-resumable sessions.

A dot on the icon marks any non-default state.

## Tab-scope filters (`?filter=`)

Comma-separated selector list, union semantics:

| Selector | Matches |
|---|---|
| `gmux` | project on every host |
| `*@server` | everything on one host |
| `gmux@server` | exact |

Supersedes the old `?project=`/`?cwd=` params. Scopes the tab's browsing surfaces — sidebar (both views), home dashboard, waiting badge — while Settings/discovery stay global. Each selector renders as a removable chip above the list; a filter matching nothing shows a hint rather than silently un-narrowing (bookmarks must not lie). Local host matches by hostname or the `local` alias.

## Persistence model

- `?filter=` and `?sidebar=activity` are **tab identity**: they live in the URL, are omitted in the default state, and are carried across all in-app navigation (`withTabParams`; sidebar, home, and project-hub links, `navigateToSession`, URL normalization). Pinned tabs survive reload/restore and are bookmarkable — the browser's tab management becomes gmux's window manager.
- **Alive only** is sessionStorage (per tab, dies with it): after a host reboot every session becomes resumable, so a durable toggle would greet the user with an empty sidebar.

## Also

- Mock mode fixed: mock sessions lacked `project_slug` stamps (folders route purely by stamps post-#228), so `?mock` rendered empty folders. Stamped, plus demo sessions covering the activity spectrum and `window.__store` for console debugging.
- User docs: new "List options" and "Filtering a tab" sections in *Using the UI*.
- In-scope empty folders stay visible under `?filter=`: a tab pinned to a project keeps that project's folder and launch button even with zero sessions (out-of-scope folders stay hidden; the no-match hint fires only when nothing is in scope).
- Tests pin the decisions: selector parsing/matching, folder scope matching, URL round-trips (default omits params, setters preserve siblings), badge + home scoping. 609 total, tsc clean.

## Filtering vs. routing (folds in #402)

This PR now also **supersedes #402** and closes it. #402 fixed the mobile "scroll to selected on open" behavior, and its review surfaced a race (selected row possibly absent when the drawer opens) that it worked around with a bounded `requestAnimationFrame` retry loop. Digging into *why* the row could be absent exposed a deeper layering bug, fixed here structurally:

- **Routing is now filter-blind.** `view`/`selectedId` resolved the addressed session against `filteredSessions`, so narrowing a tab to a project that excluded the currently-open session **evicted the terminal back to the hub**. They now resolve against the full `sessions` list — a `?filter=` can never change which session the URL addresses.
- **The selected session is pinned into the sidebar.** New `sidebarSessions` = `filteredSessions` + the selected session, so you always see what you're looking at even when the filter (or the alive-only toggle) would hide it. Its folder renders; the empty-state hint doesn't fire while a session is open.
- **The reveal effect drops the retry loop.** With the selected row now a structural guarantee (sidebar only mounts with data loaded; selected pinned past the filter), the effect reads once on `[open, selId]`. #402's one durable idea — the testable `needsReveal` geometry helper (`sidebar-reveal.ts`) — is kept.

Tests: filter-never-evicts-selected (filter-blind routing, sidebar pins past filter, no pin when nothing selected) + `needsReveal` geometry. 617 total, tsc clean.


mgabor3141:
@greptile

PR #412, opened 2026-07-23
fix(centralstore): invalidate sessions after local placement

mgabor3141:
Stacked fix 2/6 for #410. Depends on the preceding socket-test PR.

Marks the sessions projection dirty when a local session placement changes; peer-only placements remain world-only. Adds regression coverage.


mgabor3141:
Closing as superseded by #421. The underlying placement-republication defect still matters, but #421 addresses the complete placement/order invalidation class and adds composed-seam coverage, rather than only the narrower local-placement mutation covered here.

PR #421, opened 2026-07-26
fix(gmuxd): republish session stamps when placement changes

mgabor3141:
## Symptom

In the sidebar's **Projects view**, dragging a session to a new position
"often does nothing" — silently, with no console or UI error.

## Root cause: backend, not the drag code

The drag handlers and the `PATCH /v1/projects/{slug}/sessions` call are
correct, and the daemon **persisted every reorder**. It just never
re-published it.

Placement (membership *and* order) is world state, but on the wire it rides
the session rows as `project_slug` / `project_index` (the FD-1 flatten).
`wire.Cache` does re-emit a `snapshot.sessions` frame for a projects-only
batch — but it rebuilds that frame from the composer's last cached
`central.SessionsPayload`, whose per-row `Placement` join is exactly the state
the reorder invalidated. Because `ReorderSiblingScopes` reported `WorldDirty`
only, every subscriber (including fresh ones, whose initial snapshot comes
from the same cache) kept the pre-drag order until an unrelated session event
happened to dirty the sessions kind.

Deterministic repro against an isolated daemon — authoritative read vs. what
the UI actually receives:

```
PATCH /v1/projects/repro/sessions  -> {"ok":true}
REST  GET /v1/sessions   : [('ccad', 0), ('5b9b', 1), ('2f07', 2)]   # new order
SSE   snapshot.sessions  : [('c322', 1), ('5b9b', 2), ('2f07', 0)]   # pre-drag order
```

Why it read as intermittent: a later session event (spawn/exit/activity)
recomposes and the reorder "suddenly works"; and Local-peer (devcontainer)
rows were never affected, since their positions ride the projects payload.

## Fix

Every `centralstore` mutation that writes the placement table now reports
`SessionsDirty` alongside `WorldDirty`, with the rationale recorded at
`place()`:

- `ReorderSiblingScopes` — the drag path;
- `place()` (`PlaceLocalSession` / `PlaceLocalPeerSession`) — a session moving
  project had the same latent staleness;
- `SetPromotion` — an order-only change was world-only;
- `PruneLocalPeer` — pruning renumbers surviving local siblings.

No-op mutations stay clean, so this adds no spurious frames.

After the fix the SSE frame matches the authoritative read immediately, and a
real HTML5 drag in Chrome moves the row, persists, and survives reload;
repeated drags in both directions keep working.

## Tests

- `centralstore.TestPlacementMutationsDirtyBothKinds` (new) — place local,
  place Local-peer, reorder, no-op reorder, prune. Fails on `main`
  (`place a=…SessionsDirty:false…`).
- `TestPromotionNoopVersionAndExplicitOrder` — the assertion that codified the
  bug is inverted, with a note on why the row version still doesn't move.
- `cmd/gmuxd.TestHarnessReorderRepublishesSessionStamps` (new) — the
  regression pinned at the **composed seam**: a reorder through the
  coordinator must reach subscribers as new `project_index` stamps on
  `snapshot.sessions`. Verified to fail when `SessionsDirty` is dropped from
  `ReorderSiblingScopes` (`reorder did not republish snapshot.sessions with
  the new project_index stamps`).
- `store.test.ts` → "reorder failures are surfaced, never silent" (new) — toast
  plus no lingering overlay on a local rejection, toast on a peer rejection,
  silence on a network reject.
- `store.test.ts` → "reorder adds no optimistic overlay, local or peer" (new) —
  codifies the removal below; the obsolete auto-clear test for the deleted
  `reorder` pending-mutation kind is gone with it.

`go test ./...` green, all 609 `gmux-web` vitest tests green, biome lint and
`tsc --noEmit` clean.

## Also here: the reorder optimistic overlay is removed

The local overlay rewrote `projects[].sessions`, but the sidebar derives order
purely from server-stamped `project_index` — so it produced no visible
pre-echo and its rollback had no visible effect. Rather than leave dead weight
pretending to be optimistic UI, the `reorder` pending-mutation kind is dropped
(commit 2); order now reaches the UI only via the owning daemon's echo, which
is immediate after the fix above.

Supersedes #412 with the complete placement/order invalidation fix.

### DIFF
--- a/apps/gmux-web/src/store.test.ts
+++ b/apps/gmux-web/src/store.test.ts
@@ -7,7 +7,7 @@ import {
   sessionStaleness, peers, peerAppearance, peerStatusByName,
   isSessionUnavailable, urlPath, urlSearch, filteredSessions, sidebarSessions, selectedId, folders,
   navigateToSession, setNavigate,
-  applyPending, _rawSessions, _rawWorld, _setRawWorld, _pendingMutations,
+  applyPending, _rawSessions, _setRawWorld, _pendingMutations,
   applySessionsSnapshot,
   toUISession, localHostLabel, parseConnectURL, unreadCount, discovered,
   view, duplicateConversationFiles,
@@ -144,7 +144,7 @@ describe('reorder failures are surfaced, never silent', () => {
   beforeEach(() => { toasts.value = []; _pendingMutations.value = [] })
   afterEach(() => { vi.unstubAllGlobals(); toasts.value = []; _pendingMutations.value = [] })
 
-  it('toasts and retracts the overlay when the server rejects a local reorder', async () => {
+  it('toasts when the server rejects a local reorder', async () => {
     vi.stubGlobal('fetch', vi.fn().mockResolvedValue({
       ok: false, status: 500, statusText: 'Internal Server Error',
       text: () => Promise.resolve(JSON.stringify({ error: { message: 'failed to save projects' } })),
@@ -153,10 +153,12 @@ describe('reorder failures are surfaced, never silent', () => {
     expect(toasts.value[0]).toMatchObject({
       kind: 'error', message: 'Reorder failed: failed to save projects',
     })
+    // No overlay to retract: the UI still shows the server's order, which
+    // after a rejection is still the truth.
     expect(_pendingMutations.value).toHaveLength(0)
   })
 
-  it('toasts when a peer reorder is rejected (no overlay to retract)', async () => {
+  it('toasts when a peer reorder is rejected', async () => {
     vi.stubGlobal('fetch', vi.fn().mockResolvedValue({
       ok: false, status: 502, statusText: 'Bad Gateway',
       text: () => Promise.resolve(''),
@@ -1187,10 +1189,7 @@ describe('pending mutations overlay', () => {
   describe('applyPending (pure function)', () => {
     it('returns raw unchanged when there are no mutations', () => {
       const sess = [makeSession({ id: 'a' })]
-      const projs: ProjectItem[] = [{ slug: 'p', match: [] }]
-      const out = applyPending(sess, projs, [])
-      expect(out.sessions).toBe(sess)
-      expect(out.projects).toBe(projs)
+      expect(applyPending(sess, [])).toBe(sess)
     })
 
     it('mark-read clears unread and status.error on the targeted session', () => {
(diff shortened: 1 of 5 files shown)


mgabor3141:
Added the follow-up as a second commit (`refactor(web): drop the reorder optimistic overlay`), so the PR now covers both halves of the story.

The `'reorder'` pending mutation rewrote `projects[].sessions`, which nothing has read for ordering since stamps became the sole authority: no pre-echo, a retract-on-failure that undid nothing visible, and an `isResolved` comparing a list the UI ignores. Deleted, leaving `reorderSessions` as what it already effectively was — fire the PATCH, toast a rejection, stay silent on a network reject — which also collapses the local and peer paths into one instead of differing by a no-op branch. The in-drag preview is unaffected: the sidebar renders that from its own drag state.

Falls out of the union collapsing to session-only mutations: `applyPending` takes and returns just the sessions array, `isResolved` loses its projects argument, and the public `projects` projection is now a plain read of raw.

Verified against an isolated daemon with real drags in Chrome — up and down, repeated, UI and `GET /v1/sessions` agreeing after each, and the order surviving a reload:

```
drag S1 down  -> UI: 0a15 5135 fc38 | server: ['0a15', '5135', 'fc38']
reload        -> UI: 0a15 5135 fc38
```

612 web tests, protocol tests, `go test ./...` and biome all green.

mgabor3141:
One more thing I wanted before calling this done — a test that pins the *behaviour*, not just the flags.

`TestHarnessReorderRepublishesSessionStamps` (third commit) drives the composed graph — real store → coordinator → composer → `wire.Cache` → `Frames` — places two sessions, reorders them through `Coordinator.ReorderSiblingScopes` (the route the PATCH handler takes), and asserts `snapshot.sessions` comes back with the swapped `project_index`.

Worth the composed setup because this was a **seam** bug, not a flag bug: `wire.Cache` does re-emit a sessions frame for a projects-only batch, but rebuilds it from the last cached sessions payload, whose placement join is exactly what a reorder invalidates. A flag-only assertion would keep passing if anyone re-derived dirtiness anywhere between the store and the cache. On `main` it fails with `reorder did not republish snapshot.sessions with the new project_index stamps`; with the fix it passes in ~17ms.

Two notes from the same self-review pass, both deliberate non-changes:

- **Filtered drags.** With an active `?filter=`, a drag sends only the visible keys and `DecomposeReorder` appends unmentioned siblings at the tail of their scope, so hidden rows can get reshuffled. That's documented production parity (see the `decompose.go` docstring), predates this PR, and drag is already disabled for the two modes where the visible subset is most misleading (collapsed, alive-only). Out of scope here rather than quietly changed.
- **`gofmt -l` flags `internal/centralstore/peers.go`** (trailing blank line). Pre-existing on `main`, untouched by this branch — left alone so the diff stays about one thing.

PR #425, opened 2026-07-27
test(ptyserver): synchronize readiness cancellation

mgabor3141:
## Summary

- synchronize the readiness-cancellation test on the server-side request context
- wait for handler completion before reporting readiness
- assert cancellation writes neither an HTTP response nor semantic transport bytes

## Testing

- `go test ./cli/gmux/internal/ptyserver -run '^TestCancelWhileWaitingForReadinessDeliversNothing$' -count=1000`
- `go test -race ./cli/gmux/internal/ptyserver -run '^TestCancelWhileWaitingForReadinessDeliversNothing$' -count=100`
- `go test ./cli/gmux/internal/ptyserver`

### DIFF
--- a/cli/gmux/internal/ptyserver/actions_test.go
+++ b/cli/gmux/internal/ptyserver/actions_test.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


mgabor3141:
@greptileai Please re-review the latest commit; the response assertion now uses a write-tracking ResponseWriter that distinguishes no response from an explicit empty 200.

PR #437, opened 2026-08-01
docs: correct post-2.0 behavior and migration guidance

mgabor3141:
## Summary

- correct the 1.x migration story in the migration guide: v2 starts with clean SQLite state and does not import hosts, projects, references, or dead history
- document explicit peer-project setup, offline projection behavior, and recursive dismissal in current user guides
- update architecture and contributor terminology for centralstore/sessioncoord, recency activity, CLI I/O/signals, logs, and remote access
- mark ADR 0012 as superseded and remove the stale project-model/API pages rather than maintaining duplicate implementation references
- remove ineffective retention settings from the published host configuration reference
- replace invalid copy-paste examples while retaining the devcontainer feature's current `1.0.0` pin

## Product gaps deliberately not fixed here

Follow-up code work remains for SQLite retention, the v1.6 socket path, peer reference double-qualification, shell resume, parent-session decoding, the development task, and tested Docker examples.

## Validation

- `pnpm --filter @gmux/website build`
- `go test ./cli/gmux/cmd/gmux`


mgabor3141 (commented):
**Advisory review of 68e39be0 (second accuracy pass) — verified against source, v1.6.0 tag, and a full site build + link/anchor check.**

**Verdict: nearly merge-ready.** The upgrade-importer fiction is gone, the removed pages leave no dangling links, and nearly every new behavioral claim checks out against code. Four findings should be fixed first — all one-or-two-line text edits.

### P1 — incorrect

1. **migrating-to-2.md contradicts itself about running 1.x sessions.** Step 4 says "Running 1.x sessions can re-register", but the sockets section says "Sessions that were running under 1.6 are not carried into the 2.0 daemon". Code sides with the latter: v1.6.0 runners bind in unsuffixed `/tmp/gmux-sessions` (v1.6.0 `run.go:96`), while 2.0's legacy scan only covers `$XDG_RUNTIME_DIR/gmux/sessions` and `/tmp/gmux-sessions-<uid>` (`packages/paths/paths.go:65-75`). Fix step 4 to say running sessions must be restarted.

2. **reference/cli.md run table: auto-detach does not print a session ID.** The new TTY/TTY/`GMUX=1` row claims "print the session ID on stdout", but the nested path (`spawnDetached(..., waitForRegistration=false)` → `startDetachedNoWait`, `run.go:218-221, 851-866`) prints only a stderr message and never an ID. Only explicit `-d` prints the id on stdout. A script trusting this row captures nothing.

### P2 — broken/confusing

3. **security.md anchor broken by this PR.** `#reverse-proxy-with-oidc-traefik--pocketid` was "fixed" to a single dash, but the built heading id keeps the double dash (verified in `dist/running-in-docker/index.html`). Revert. This is the only bad internal link in the whole built site.

4. **using-the-ui.md "Discovered" is not local-only.** The Discovered list still merges peer-advertised discovered rows from connected network hosts (`store.ts` `discovered` computed appends `peerDiscovered`; production populates `WorldProjection.PeerDiscovered`). The deleted clause "including directories advertised by peer hosts" was correct — "From other hosts" is a different list (the peer's *configured* projects).

### P3 — polish, non-blocking

- changelog.mdx: `--new --no-wait` **prints** the id as soon as the session exists, *before* delivery/admission (`agent.go:472-481` explicitly disclaims this); it's the process exit that waits for admission. Say "returns", not "prints … as soon as … admission complete".
- ADR 0012: no blank line between the new historical-record blockquote and `**Date:**` — markdown lazy continuation folds Date/Related into the quote.
- README "supported adapter conversations remain resumable" over-narrows: conversation-less shells stay rerunnable too (`render.go:87`: `everRan && command && verdict != Gone`).

### Verified correct (samples)

kill = SIGHUP → 2 s → SIGKILL to process group; 30 s fallback socket scan; `last_output_at` stamped only on unread false→true; home = alive-only day buckets with dated tail omitted; 10k-line replay buffer over ~2 MiB on-disk scrollback; `retention_days`/`retention_max` inert in production (only `Sweep` enforces them and serveCentral never calls it) so their removal from host.toml docs is right; recursive dismiss stamps `dismissed_at` and retains rows/scrollback; no 1.x importer exists; `/v1/state/*` + `/v1/shutdown` unix-only; bounded `gmuxd.log` + `.1` rotation incl. autostart; OSC 0/2 only; scrollback GET / read POST; pi/claude/codex all hook-driven; sync `--new` id on stderr; peer disconnect prunes projection; match precedence path>remote, longest path; "From other hosts" UI section exists; pi URL redirect target correct; no remaining links to the deleted pages.

Fix 1–4 and this is good to merge. (Advisory only — not a formal approval/blocking review.)

PR #438, opened 2026-08-01
feat(adapter): add Claude semantic agent parity

mgabor3141:
## Summary
- enable Claude Code semantic launch, prompt, follow-up, steer, and cancel through the common readiness-gated runner contract
- add Claude transcript/exchange rendering for result-bearing waits and `gmux agent logs`, excluding thinking, tool results, metadata, and sidechains
- emit adapter readiness on `SessionStart`, including fresh sessions without a transcript

## Parity / implementation notes
- launch uses native `claude --model` and `--name`; prompts never ride argv
- submit uses Claude's Enter binding; cancel uses reserved/non-rebindable Ctrl+C rather than context-sensitive Escape
- existing authoritative `UserPromptSubmit` / `Stop` / `StopFailure` / `SessionEnd` lifecycle continues to classify completion, failure, and interruption
- existing resume UUID lineage and transparent runner restart remain unchanged
- malformed live transcript tails are ignored; persisted visible assistant iterations provide active partial reports
- Claude has no stable hook field distinguishing permission/user-answer waits from other open-turn activity, so gmux truthfully leaves these active rather than reporting a false settled state

## Shared-contract impact
One shared runner test removes Claude from the deliberately-unsupported adapter list. No protocol, endpoint, CLI, status schema, or shared adapter interface changed. This should integrate independently of the parallel Codex adapter work; the likely conflict is only that unsupported-adapter test list.

## Test evidence
Passed:
- `go test ./packages/adapter/adapters ./cli/gmux/internal/ptyserver ./cli/gmux/cmd/gmux`
- all packages in `cli/gmux`, `packages/{adapter,paths,scrollback,socklease,sessionenv,workspace}`, and `services/gmuxd`

The repository-wide module loop reached existing environment-sensitive e2e failures: daemon startup timeout and restart HTTP 500 in `tests/e2e`; all preceding module suites passed.

### DIFF
--- a/cli/gmux/cmd/gmux/agent.go
+++ b/cli/gmux/cmd/gmux/agent.go
@@ -437,9 +437,9 @@ func cmdAgentPrompt(ref, mode string, noWait bool, timeoutSecs int, text *string
 // so tests can drive `--new`'s output contract without forking a real agent.
 var agentLaunchSession = launchDetachedSession
 
-// agentLaunchAdapter is the adapter `--new` launches through. The launch is
-// pi-only for now (there is no --adapter flag), exactly like the rest of this
-// namespace; a variable so tests can substitute a non-launcher adapter.
+// agentLaunchAdapter is the default adapter `--new` launches when no
+// --adapter selector is supplied; a variable so tests can substitute a
+// non-launcher adapter.
 var agentLaunchAdapter adapter.Adapter = adapters.NewPi()
 
 // cmdAgentPromptNew implements `gmux agent prompt --new`: launch a session and
--- a/cli/gmux/cmd/gmux/agent.go
+++ b/cli/gmux/cmd/gmux/agent.go
@@ -10,6 +10,7 @@ package main
 // and wait share one stdout exchange report; logs reads adapter storage only.
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -503,7 +504,7 @@ func cmdAgentPromptNew(adapterName, model, name string, noWait bool, timeoutSecs
 		fmt.Fprintln(os.Stderr, "gmux:", err)
 		return waitExitError
 	}
-	return deliverPrompt(cliSession{ID: id}, agentModePrompt, noWait, timeoutSecs, prompt)
+	return deliverPrompt(cliSession{ID: id, Adapter: launchAdapter.Name()}, agentModePrompt, noWait, timeoutSecs, prompt)
 }
 
 // agentLaunchArgv translates launch options into an argv, or reports that this
@@ -648,6 +649,13 @@ func reportAgentPromptSuccess(sess cliSession, status int, body []byte, noWait b
 	}
 	switch res.Outcome {
 	case waitOutcomeCompleted, waitOutcomeInterrupted, waitOutcomeError, outcomeTimeout:
+		// Claude's terminal hook and gmuxd's outcome projection travel on
+		// independent streams. If the fused prompt response won that race before
+		// carrying any exchange data, the already-settled observational wait is
+		// the authoritative read and completes immediately.
+		if sess.Adapter == "claude" && len(res.Exchanges) == 0 && res.Output == "" {
+			return waitSession(context.Background(), sess, 0, 0, "", "", false, os.Stdout)
+		}
 		text := ""
 		if len(submitted) > 0 {
 			text = submitted[0]
--- a/cli/gmux/cmd/gmux/agent.go
+++ b/cli/gmux/cmd/gmux/agent.go
@@ -155,7 +155,7 @@ func parseAgentPrompt(args []string) (*command, error) {
 	c := &command{mode: modeAgent, agentSub: "prompt", agentMode: agentModePrompt}
 	modeSet := ""
 	noWaitSet, timeoutSet := false, false
-	newSet, modelSet, nameSet := false, false, false
+	newSet, adapterSet, modelSet, nameSet := false, false, false, false
 	// A leading help token asks about the verb. Only the leading position:
 	// after the ref, every token is verbatim prompt text, so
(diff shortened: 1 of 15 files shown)


mgabor3141 (commented):
Blocking findings at `4a963fa7` (advisory review):

- **High — abandoned branch content is rendered/leaked.** `packages/adapter/adapters/claude_conversation.go:19-46` drops Claude's `uuid`/`parentUuid` and linearizes every non-sidechain record; both renderers then consume that list (`:85-130`). A parent-linked fork probe rendered `root`, abandoned user text, then the active follow-up. ADR 0030 requires only the final persisted active branch; this also risks exposing abandoned content.
- **High — launch support is unreachable.** Claude declares `AgentLauncher` at `packages/adapter/adapters/claude.go:97-107`, but `cli/gmux/cmd/gmux/agent.go:452-465` always launches the Pi global. There is no Claude selection path. #439's new selector permits only pi/codex, so joint integration still needs Claude routing.
- **Medium — image-only turns are not exchange boundaries.** `renderClaudeContent` sets `userBoundary` for text only (`claude_conversation.go:70-79`); the user filters at `:95-97` and `:116-121` therefore drop image-only turns and misattribute/drop their assistant response. A focused probe returned 0 exchanges for image-only user + assistant. Greptile's issue remains unfixed.
- **Medium — invalid user settings are destructively swallowed.** `claude_hook.go:59-81` removes `--settings`; `:84-92` silently ignores unreadable/malformed values. A missing value is dropped too. Instead of preserving Claude's normal error/argv as the comments promise, gmux launches with only its own settings.

Evidence: `go test ./packages/adapter/adapters ./cli/gmux/internal/ptyserver` passes. Temporary probes demonstrated the image and fork failures and were removed. No live Claude CLI lifecycle run was performed, so current hook firing/Ctrl+C behavior remains unverified. Head is based on v2.0 (`3768c4de`) rather than current main (`dc61d57b`): rebase and joint testing with #439 are required.

**Verdict: block** pending these fixes and rebase.

mgabor3141 (commented):
Re-review at `ec6d0d32` against current main `4138078e`:

All four original findings are now genuinely closed with regression coverage:

- parent-linked active branch reconstruction: `claudeActiveBranch` plus `TestClaudeConversationUsesActiveParentBranch`;
- reachable launch: `--adapter claude` routes to `NewClaude`, and `TestAgentPromptNewClaudeSelection` verifies the Claude argv;
- image-only turns: the image case now establishes a boundary, covered by `TestClaudeImageOnlyUserIsExchangeBoundary`;
- invalid settings: extraction/parsing errors return original argv with `ok=false`, covered for missing/empty/unreadable/malformed inputs by `TestClaudeHookCommand_InvalidSettingsPreserved`.

The rebase is clean and focused tests pass:
`go test ./packages/adapter/adapters ./cli/gmux/cmd/gmux ./cli/gmux/internal/ptyserver`.

I found one new **medium, blocking semantic-read regression** in the branch fix. `readClaudeEntries` calls branch selection before hiding sidechains (`packages/adapter/adapters/claude_conversation.go:31-47`); `claudeActiveBranch` chooses the final UUID-bearing record without excluding `IsSidechain` (`:50-91`), while filtering happens only later (`:132-133`, `:153-154`). A UUID/parentUUID-bearing sidechain can be the latest persisted record while Claude appends subagent work. It then becomes the traversal leaf; filtering hides its private text but the main branch tail has already been discarded. A focused probe with `u1 -> main answer` plus a trailing sidechain child returned the root exchange with zero iterations/empty terminal instead of `main answer`. Please select the latest eligible main-conversation leaf (while retaining invisible nodes needed as ancestors) and add this trailing-sidechain regression case.

Integration note: #439 independently makes the same selector `pi|codex`, whereas this branch makes it `pi|claude`; conflict resolution must preserve `pi|claude|codex`, not choose one side. The default-Pi and non-`--new` validation behavior here otherwise look sound.

**Final verdict: block** for the sidechain-leaf regression; the four original review gaps are resolved. No install/live Claude/credentials were used.

mgabor3141 (commented):
Final verification at `14034503`: **merge**.

The remaining branch issue is closed correctly. `claudeActiveBranch` now excludes UUID-bearing `IsSidechain`/`IsMeta` records only when choosing the leaf, while retaining them in `byID` and the parent walk. That prevents a trailing hidden record from erasing the visible main tail without breaking a visible continuation whose parent is hidden.

Regression coverage is appropriately narrow and passes:
- `TestClaudeConversationIgnoresTrailingHiddenLeaves` covers sidechain and meta tails;
- `TestClaudeConversationRetainsHiddenTraversalAncestors` proves hidden-parent traversal remains intact;
- `TestClaudeConversationUsesActiveParentBranch` still passes for abandoned branches;
- the image-only boundary test remains passing.

Evidence:
- relevant tests with `-count=20`: pass;
- `go test ./packages/adapter/adapters ./cli/gmux/cmd/gmux ./cli/gmux/internal/ptyserver`: pass;
- all reported CI checks: green.

Current main is one website-only commit beyond the PR merge base, with no overlap found. #439's selector conflict must eventually preserve all adapter values, but it is not a blocker for #438. No install, credentials, or live Claude invocation used.

mgabor3141:
Closing after a live manual gate rather than merging. The code passed review, but live testing showed Claude's interactive TUI is not a sound automation transport: readiness required arbitrary delays, submission depends on focus/modal/keybinding state, and results needed a compensating observational wait. Decision: capabilities attach to the session backend, not the adapter — Claude semantic control (gmux agent prompt/steer/wait/…) will be delivered through the upcoming ACP backend; `gmux -- claude` remains a plain interactive terminal session. An ADR documenting session backends and capability boundaries is being drafted. The review-hardened observational pieces from this branch (active-branch transcript reconstruction, privacy filtering, safe settings handling, resume lineage) will be extracted into a follow-up PR without semantic-control claims.

PR #456, opened 2026-08-04
fix(snapshot): redact git remote userinfo on session wire

mgabor3141:
## Summary
- redact URL userinfo from local session remotes at the sole consumer-facing wire conversion boundary
- apply the same redaction to relayed peer session rows before they are emitted again
- preserve source maps so durable and peer projections remain available for project matching

## Evidence
The local data flow was `centralstore.Session.Remotes` → `snapshot/central.SessionRow` → `snapshot/wire.Converter.session` → `wire.Session.Remotes` → GET `/v1/sessions`, SSE `snapshot.sessions`, and peer subscribers. `Converter.session` previously assigned the map verbatim. Peer rows joined the same payload via `Converter.Sessions` and were also copied verbatim. Both paths now call the existing `statetool.RedactURLUserinfo` helper at that shared serialization boundary.

The state export path already redacts each session remote in `statetool.exportSession`; its behavior is unchanged and remains covered by `TestExportIsRedactedAndDeterministic`. Other remotes-bearing structures are internal matching/durable projections or runner input, not consumer-facing session serialization boundaries.

## Tests
- `pnpm moon run gmuxd:test`
- `pnpm moon run gmuxd:lint`
- regression covers credential-bearing local and relayed peer remotes, userinfo-free SSH remotes, and source-map non-mutation

### DIFF
--- a/services/gmuxd/internal/snapshot/wire/convert.go
+++ b/services/gmuxd/internal/snapshot/wire/convert.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gmuxapp/gmux/services/gmuxd/internal/centralstore"
 	"github.com/gmuxapp/gmux/services/gmuxd/internal/peering"
 	central "github.com/gmuxapp/gmux/services/gmuxd/internal/snapshot/central"
+	"github.com/gmuxapp/gmux/services/gmuxd/internal/statetool"
 )
 
 // Converter holds the pure derivation policy injected at construction:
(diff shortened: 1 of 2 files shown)


mgabor3141 (commented):

I traced both local and relayed peer rows through the shared session-wire conversion, read every changed line, and verified that `redactRemotes` copies rather than mutates either source map. The change is limited to session remotes and reuses the established `statetool.RedactURLUserinfo` contract. The regression exercises credential-bearing URL and schemeless forms, an unchanged SCP-style SSH remote, and source-map immutability. `go test -race ./internal/snapshot/wire` passes, and all required CI checks pass. No remaining findings.

PR #457, opened 2026-08-04
fix(adapter): re-watch removed-then-recreated directories

mgabor3141:
## Summary

- prune removed or renamed directories (and descendants) from filewatch's watch bookkeeping so recreated paths get fresh fsnotify watches
- keep a watch on the configured root's parent so removing and recreating the root itself is recoverable
- ignore unrelated events observed through that parent watch
- add integration regressions for both recreated subdirectories and the configured root

## Evidence

A live Linux/fsnotify probe confirmed that `rmdir` emits `REMOVE` for the watched directory and removes it from `Watcher.WatchList`; recreating the path emits only `CREATE` through the parent watch. `rename` similarly emits `RENAME` and drops the original watch. The old bookkeeping retained those paths, so `addWatch` skipped re-adding them.

Both new tests were also run against the pre-fix implementation and failed after their 5-second polling deadlines:

- `TestWatchRewatchesRecreatedSubdir`
- `TestWatchRewatchesRecreatedRoot`

No file-level watches exist in this package; file rename/remove event reporting remains unchanged. Directory `Remove` and `Rename` now share the same pruning path.

## Test plan

- `go test ./filewatch -count=20`
- `go test -race ./filewatch`
- `pnpm moon run adapter:test`
- `pnpm moon run adapter:lint`

### DIFF
--- a/packages/adapter/filewatch/filewatch.go
+++ b/packages/adapter/filewatch/filewatch.go
@@ -63,7 +63,16 @@ func Watch(ctx context.Context, root, suffix string, emit func(Event)) error {
 		}
 	}
 
-	tw := &treeWatcher{w: w, suffix: suffix, emit: emit, watched: map[string]bool{}}
+	root = filepath.Clean(root)
+	tw := &treeWatcher{w: w, root: root, suffix: suffix, emit: emit, watched: map[string]bool{}}
+
+	// Keep watching the root's parent so removal of root itself does not leave
+	// us with no watch from which to observe its recreation.
+	if parent := filepath.Dir(root); parent != root {
+		if err := w.Add(parent); err != nil {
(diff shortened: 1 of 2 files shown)


mgabor3141 (commented):

I read every changed line and traced Linux/fsnotify removal, rename, recreation, descendant pruning, and root recreation. The bookkeeping now drops exactly the kernel watches invalidated by directory removal/rename; the root-parent watch supplies the otherwise-missing recreation event, while unrelated parent events are filtered. File remove/rename reporting remains intact. Both regressions fail on the old mechanism per the PR evidence, and locally `go test ./filewatch -count=20` plus `go test -race ./filewatch` pass. All required CI checks pass. No remaining findings.

PR #458, opened 2026-08-04
perf(gmuxd): cache resume commands for session lists

mgabor3141:
## Summary

- cache adapter-derived resume commands in the conversation index while sources already describe each transcript
- render session-list commands from that in-memory index instead of re-reading every dead conversation JSONL
- clear cached commands with conversation removals and return defensive copies
- add a 1,000-distinct-session regression test proving list-time lookups do not invoke `DescribeConversation`

## Profiling / measurements

Fixture: this host's production state, ~994 sessions (~947 dead local sessions). The dead pi sessions referenced 798 distinct JSONL transcripts totaling **249 MiB**.

Before (warm production daemon):

- direct `GET /v1/sessions` over the Unix socket: **3.665s**
- `gmux ls --json`: **5.768s** in the sampled run (the daemon endpoint alone accounts for the scaling defect)

Root cause: wire conversion called `discovery.ResolveResumeCommandFor` for every dead session with a conversation ref. For pi, that calls `DescribeConversation`, whose `os.ReadFile` + full JSONL parse re-read all 798 transcripts on every list request.

After (scratch daemon built from this branch, isolated copied state, 997 sessions), five direct endpoint runs:

- **34ms, 28ms, 27ms, 30ms, 33ms**

This leaves `gmux ls` / `gmux ls --all` comfortably subsecond at ~1,000 sessions; both consume the same endpoint and their client-side filtering/sorting is in-memory. For the 947 sessions that were dead in both snapshots, the before/after rendered commands had **0 differences**.

The web UI's composed session snapshots use the same wire converter, so it shared the transcript re-read during snapshot recomposition. This change improves that path too; its HTTP/SSE delivery itself is cache-backed.

## Validation

- `pnpm moon run gmuxd:test`
- `pnpm moon run gmuxd:lint`
- scratch-daemon runtime measurement above (no binaries installed over the live daemon)

### DIFF
--- a/services/gmuxd/cmd/gmuxd/serve_central.go
+++ b/services/gmuxd/cmd/gmuxd/serve_central.go
@@ -234,10 +234,7 @@ func serveCentral(stderr io.Writer, replace bool) int {
 		return h
 	}
 
-	converter := &wire.Converter{Titlers: make(map[string]func([]string) string), SemanticAgents: make(map[string]bool), ResumeCommand: func(adapterName, ref string) []string {
-		legacy := &compatSession{Adapter: adapterName, ConversationRef: ref}
-		return discovery.ResolveResumeCommandFor(legacy.Adapter, legacy.ConversationRef)
-	}, IsLocalPeer: func(name string) bool { return peerManager != nil && peerManager.IsLocalPeer(name) }}
+	converter := &wire.Converter{Titlers: make(map[string]func([]string) string), SemanticAgents: make(map[string]bool), ResumeCommand: convIndex.LookupResumeCommand, IsLocalPeer: func(name string) bool { return peerManager != nil && peerManager.IsLocalPeer(name) }}
 	for _, a := range adapters.All {
 		if titler, ok := a.(adapter.CommandTitler); ok {
 			converter.Titlers[a.Name()] = titler.CommandTitle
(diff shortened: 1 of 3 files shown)


mgabor3141 (commented):

One correctness regression remains:

- `services/gmuxd/internal/conversations/index.go:215-218`: on a transient `DescribeConversation` error, `Scan` now overwrites an already-good cached resume command with nil. Existing index behavior deliberately retains the previous `Info` when a refresh cannot be described; this new side cache breaks that stale-good behavior. A file-change upsert can race a partial/transiently unreadable transcript and receive no later event, after which dead sessions render with an empty command and become non-resumable until restart. The prior list path retried description on each render, so this is a semantic change despite the unchanged JSON shape. Preserve an existing cached command on describe failure (while still leaving first-time failures absent), and add a regression: successful scan, mutate the fake describer to fail, rescan the same ref, then assert the prior command remains cached. Explicit removal should continue clearing it.

I otherwise confirmed the profile identifies repeated transcript parsing, the hot list path is converted to in-memory lookup, defensive copies prevent cache mutation, the 1,000-ref test bounds render-time descriptor I/O, output comparison is reported as unchanged, and CI passes.

mgabor3141:
Fixed in c73d66f7. `Scan` now preserves stale-good cached resume commands when `DescribeConversation` fails transiently; only an explicit removal clears them. Added the requested successful scan → descriptor failure → rescan → cache preserved → `RemoveByRef` clears regression test. Focused conversations package tests pass.

mgabor3141 (commented):

The delta removes the cache overwrite on descriptor failure, preserving the main index's stale-good semantics. The new regression proves successful scan → transient failed rescan retains the prior resume command, while authoritative `RemoveByRef` still clears it. First-time failures remain naturally absent. `go test -race ./internal/conversations` and all required CI checks pass. No remaining findings.

PR #459, opened 2026-08-04
fix(peering): harden reconciliation and surface adapter errors

mgabor3141:
## Summary

- tolerate a peer disappearing between `PeerStatus` and `GetPeer` during concurrent manual-peer reconciliation, then restore it from durable state
- read the project-refresh status callback value under `Peer.mu`
- log peer projection conversion, local-placement prune/read, and placement-upsert failures instead of silently dropping them

## Evidence trail

This addresses the aggregator-verified peering findings from `.memory/AGGREGATE.md`:

- Tier A #1 / reports 312 and 143: `reconcileManualPeers` could nil-deref after concurrent `RemovePeer`
- Tier B #9 / report 130: `fetchProjects` read `p.status` outside `p.mu`
- Tier B #17 / reports 143 and 312: adapter errors were silently ignored

The reconciliation regression test queues `RemovePeer` behind the read lock held by `PeerStatus`, forcing it to win before the subsequent `GetPeer`. Against the old implementation it fails with:

```
reconcile panicked: runtime error: invalid memory address or nil pointer dereference
```

The status callback test continuously updates peer status while project refresh callbacks run and is included in the race suite.

## Verification

- `go test ./services/gmuxd/cmd/gmuxd ./services/gmuxd/internal/peering`
- `go test -race ./services/gmuxd/cmd/gmuxd ./services/gmuxd/internal/peering`
- `go test ./services/gmuxd/cmd/gmuxd -run TestReconcileManualPeersToleratesConcurrentRemoval -count=20`
- `pnpm test`

### DIFF
--- a/services/gmuxd/cmd/gmuxd/bootstrap_peering_adapters.go
+++ b/services/gmuxd/cmd/gmuxd/bootstrap_peering_adapters.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
 
 	"github.com/gmuxapp/gmux/services/gmuxd/internal/centralstore"
 	"github.com/gmuxapp/gmux/services/gmuxd/internal/config"
@@ -54,11 +55,17 @@ func (a *centralPeerAdapter) PeerSessions() []wire.Session {
 	rows := a.manager.SessionProjections()
 	out := make([]wire.Session, 0, len(rows))
 	for _, r := range rows {
-		b, _ := json.Marshal(r)
-		var s wire.Session
-		if json.Unmarshal(b, &s) == nil {
-			out = append(out, s)
+		b, err := json.Marshal(r)
+		if err != nil {
+			log.Printf("peering: %s: marshal session projection %s: %v", r.Peer, r.ID, err)
+			continue
 		}
+		var s wire.Session
+		if err := json.Unmarshal(b, &s); err != nil {
+			log.Printf("peering: %s: convert session projection %s: %v", r.Peer, r.ID, err)
+			continue
+		}
+		out = append(out, s)
 	}
 	return out
 }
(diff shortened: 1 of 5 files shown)


mgabor3141 (commented):

The panic is guarded, but the demonstrated concurrent-reconciliation race is only partially closed:

- `bootstrap_peering_adapters.go:168-180`: `GetPeer` returns a raw pointer after dropping `Manager.mu`. A competing reconciliation can remove that peer after this lookup but before this loop deletes it from `desired`. This invocation then observes/config-compares the stale pointer, deletes the desired entry, and does not restore the now-absent runtime peer in the add pass. The new test forces only the earlier ordering (remove before `GetPeer`), so it misses this equally reachable ordering from the same two concurrent HTTP handlers. Reconciliation needs a manager-level atomic operation (under `Manager.mu`) that ensures the desired config is installed/replaced, rather than a `GetPeer`/`RemovePeer`/`AddPeer` check sequence. Add a deterministic regression for removal after lookup/before desired deletion, or a concurrency test against the atomic API that proves the desired peer remains present.

The locked `p.status` snapshot is correct, error paths now emit useful context, focused/race tests are reported green, and all CI checks pass.

mgabor3141:
Addressed in `65a66a8`.

- Added `Manager.EnsurePeer`, which serializes comparison/replacement with `AddPeer` and `RemovePeer` through a lifecycle mutation lock. The lock remains held while an old connection is cancelled, joined, and cleaned up, so replacement startup cannot race old-generation cleanup.
- Reconciliation no longer uses `GetPeer` or deletes desired entries based on a raw pointer. It removes runtime-only peers first, then calls `EnsurePeer` for every durable desired row.
- Updated the concurrency regression to force removal after `PeerStatus` has begun the runtime snapshot and before the desired-config ensure pass, then assert the durable peer is installed. It passes 50 consecutive runs.

Verification rerun:
- `go test -race ./services/gmuxd/cmd/gmuxd ./services/gmuxd/internal/peering`
- `go test ./services/gmuxd/cmd/gmuxd -run TestReconcileManualPeersToleratesConcurrentRemoval -count=50`
- `pnpm test`

mgabor3141 (commented):

The delta removes the stale-pointer decision entirely: reconciliation now ensures every durable row, and `Manager.EnsurePeer` serializes comparison/replacement with all add/remove lifecycle mutations via `mutationMu` while retaining the narrower state lock around map access. This closes both removal-before-lookup and removal-after-lookup schedules without holding `mu` while waiting for connection shutdown. The regression exercises snapshot→queued removal→ensure restoration; focused `go test -race ./services/gmuxd/cmd/gmuxd ./services/gmuxd/internal/peering` passes, as do all required CI checks. No remaining findings.

PR #460, opened 2026-08-04
fix(daemon): prevent slow incumbents from triggering takeover

mgabor3141:
## Summary

- treat a connected or ambiguous Unix socket as owned independently of health-response latency
- serialize concurrent CLI autostart with an owner-only flock, jittered recheck, and candidate readiness/exit wait
- move the daemon incumbent check ahead of adapter discovery, conversation indexing, and state/database work
- make shutdown timeout fail, reject malformed health identities, and refuse to unlink live/ambiguous socket owners

## Investigation

A hermetic SIGSTOP reproduction with isolated HOME/XDG/socket state and five concurrent clients produced five candidate `gmuxd run` processes from the old code. Each performed launcher discovery and conversation indexing before probing the incumbent. The decision path and worst-case interleaving are documented locally in `.memory/report-daemon-takeover.md`.

Repeating the same scratch reproduction after this change left exactly one process (the stopped incumbent) and produced no autostart log/candidate bootstrap.

## Tests

- `pnpm moon run gmux:test gmuxd:test`
- `pnpm moon run gmux:lint gmuxd:lint`
- `go test -race ./services/gmuxd/internal/unixipc ./cli/gmux/cmd/gmux`

### DIFF
--- a/cli/gmux/cmd/gmux/daemon.go
+++ b/cli/gmux/cmd/gmux/daemon.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"log"
 	"net"
@@ -11,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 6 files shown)


mgabor3141 (commented):

Two ownership-safety gaps remain, both tracing directly to the investigation report:

1. `unixipc.HealthIdentity` treats any syntactically valid JSON (including `{}` or `{"data":{}}`) as a valid identity with empty version/PID. `implicitIncumbentCheck` then treats that empty version as a known version mismatch and permits implicit takeover of a live/ambiguous owner, contradicting the report's “unavailable/malformed identity => conservative yield” rule. The release CLI has the same issue in `gmuxdNeedsStartContext`: a valid JSON body missing `data.version` triggers replacement. Require a non-empty version (and whatever PID validity the actual health contract guarantees) before identity is considered usable; add missing-field tests on both daemon and CLI paths, not just invalid JSON syntax.

2. The report identifies old-owner cleanup unlinking a successor pathname as the final socket-steal step, but `unixipc.Cleanup` remains unconditional `os.Remove(sockPath)`. Make cleanup identity-bound/pinned so it removes only the socket inode created by that listener, and add a regression that replaces the pathname before old-owner cleanup and proves the successor remains named. This likely requires returning/retaining a pin or listener lease from bind rather than looking up the path only at cleanup time.

The measured slow-response/autostart cause, conservative connect classification, cross-process spawn serialization, startup-first precheck, shutdown timeout correction, and pinned stale-socket removal otherwise trace to the report. Focused race suites and CI pass.

mgabor3141:
Review round 1 remediated and force-pushed in `833dbfd`:

1. Identity is now contract-valid only when `unixipc.HealthIdentity` receives a non-empty version and positive PID. The release CLI likewise treats a missing/blank version as unavailable identity and conservatively preserves the connected owner. Added table coverage for `{}`, `{"data":{}}`, missing version/PID, blank version, invalid PID, and malformed JSON, plus a release-CLI missing-version regression.
2. Removed pathname-only `unixipc.Cleanup`. `Listen` now disables `UnixListener` auto-unlink, pins the just-bound socket inode, and returns an owned listener whose idempotent `Close` removes the pathname only while it still names that pin. Production cleanup now closes this listener. Added a replacement-before-old-close regression proving the successor remains reachable.

Revalidated:
- `go test ./services/gmuxd/internal/unixipc ./cli/gmux/cmd/gmux ./services/gmuxd/cmd/gmuxd`
- `go test -race ./services/gmuxd/internal/unixipc ./cli/gmux/cmd/gmux`
- `go vet ./cli/gmux/... ./services/gmuxd/...`
- moon test/lint targets remain green

mgabor3141 (commented):

The delta closes both findings. Health identity now requires a non-blank version and positive PID, while the release CLI independently refuses takeover when version is missing/blank; the table-driven daemon test and release-path test cover syntactically valid but contract-invalid bodies. Socket lifetime is now represented by a listener-held inode pin with Go auto-unlink disabled; `Close` removes only if the pathname still names that listener, and the successor-rebind regression proves old-owner cleanup cannot unlink the successor. Server shutdown now routes through that listener close. Focused race suites for unixipc, CLI, and gmuxd pass, as do all required CI checks. No remaining findings.

PR #461, opened 2026-08-04
perf(gmuxd): skip indexing in autostart contenders

mgabor3141:
## Summary

- defer the full conversation-history snapshot until after daemon state/socket ownership is settled
- let same-version autostart contenders yield before reading and parsing 2,366 conversations
- add an integration guard whose challenger would block on a FIFO conversation if snapshotting happened before the incumbent precheck

This is stacked on #458, which owns the pre-agreed list-endpoint scope and removes the investigation's primary recurring cost (session rendering reparsing every dead transcript). This PR addresses the separate startup/autostart amplification found by the daemon-wide profile.

## Investigation evidence

The live daemon had 1,430 durable local rows (1,415 exited; 964 non-dismissed) and had performed 2.9 TB of logical reads in about 25 hours. On a copied-state scratch daemon, 30 health renders in 30 seconds consumed 52.95 CPU-seconds; 45.63 seconds (86.2%) was `Converter.session -> ResolveResumeCommandFor -> Pi.DescribeConversation`. #458 fixes that shared converter path.

Independently, `convIndex.Snapshot()` ran before `bootstrapOwnership`'s healthy-incumbent precheck. The live log contained 216 `conversations: indexed` lines, including 4-8 identical lines in the same second: every autostart contender paid the complete history scan before yielding. Scratch startup attributed 3.77 GB of allocations to the initial conversation snapshot.

## Before / after (~950 visible sessions)

Fixture: SQLite online backup of the live state, 1,426 durable rows / about 960 visible rows, 2,366 indexed conversations; isolated state/config/socket/port. A healthy fixed incumbent was running, then a same-version challenger was timed.

| Challenger | Wall | User CPU | System CPU | Conversation scan |
|---|---:|---:|---:|---|
| Before | 7.791 s | 7.875 s | 0.738 s | 2,366 conversations |
| After | 0.041 s | 0.020 s | 0.003 s | none |

That is a 99.5% wall-time reduction for losing autostart contenders. Actual successful daemon startup semantics are unchanged: the snapshot still completes synchronously before the index is consumed or listeners serve requests.

For completeness, with #458 underneath this PR, the same 30-render profile falls from 52.95 to 0.72 CPU-seconds (-98.6%) and logical reads during the run fall to 0.80 MB.

## UI / wire observation

A 20-second read-only live SSE capture contained one initial 826,490-byte sessions snapshot, one 14,728-byte world snapshot, and 43 tiny activity events; it did not show periodic full-world broadcasts. Evidence points to daemon CPU/GC starvation from transcript reparsing, with the large initial frame adding reconnect/browser work. Wire redesign remains out of scope.

## Validation

- `pnpm moon run gmuxd:test gmuxd:lint`
- takeover integration guard verifies a healthy challenger exits without opening a blocking conversation FIFO

Depends on #458.

### DIFF
--- a/services/gmuxd/cmd/gmuxd/serve_central.go
+++ b/services/gmuxd/cmd/gmuxd/serve_central.go
@@ -234,10 +234,7 @@ func serveCentral(stderr io.Writer, replace bool) int {
 		return h
 	}
 
-	converter := &wire.Converter{Titlers: make(map[string]func([]string) string), SemanticAgents: make(map[string]bool), ResumeCommand: func(adapterName, ref string) []string {
-		legacy := &compatSession{Adapter: adapterName, ConversationRef: ref}
-		return discovery.ResolveResumeCommandFor(legacy.Adapter, legacy.ConversationRef)
-	}, IsLocalPeer: func(name string) bool { return peerManager != nil && peerManager.IsLocalPeer(name) }}
+	converter := &wire.Converter{Titlers: make(map[string]func([]string) string), SemanticAgents: make(map[string]bool), ResumeCommand: convIndex.LookupResumeCommand, IsLocalPeer: func(name string) bool { return peerManager != nil && peerManager.IsLocalPeer(name) }}
 	for _, a := range adapters.All {
 		if titler, ok := a.(adapter.CommandTitler); ok {
 			converter.Titlers[a.Name()] = titler.CommandTitle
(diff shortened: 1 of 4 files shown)


mgabor3141 (commented):

The investigation report precedes and supports the implementation: profiles isolate repeated transcript parsing as the dominant recurring load (owned and fixed by stacked PR #458) and full conversation indexing before ownership check as the independent contender cost fixed here. The track's own delta only moves `convIndex.Snapshot` after successful ownership; it does not alter wire contracts or independently patch the list/converter mechanism. Successful daemon startup still snapshots synchronously before converter use/listener service. The FIFO regression fails closed if a same-version contender reaches indexing, and repeated focused takeover tests pass. Reported contender time falls 7.791s → 0.041s on isolated copied state; all required CI checks pass. No remaining findings.

PR #462, opened 2026-08-04
perf(web): index family projections once

mgabor3141:
## Summary

- build and cache one task-family index per web session snapshot
- memoize cycle-safe membership/root classification and reuse the children index across drawer, routing, sidebar, and header callers
- replace gmuxd's per-tree-node row scan with ID and parent indexes
- add 1,000-session / 500-child regression coverage and benchmarks for both paths

## Performance

Corpus: 1,000 sessions, one root with 500 direct children, 499 unrelated sessions, on an i7-9700K.

### Web (`vitest bench`)

The legacy drawer-projection benchmark did not complete within a 120s runner timeout. After:

- changed-snapshot classification: **1,078 ops/s** (0.31ms p75)
- selected-child drawer projection from the shared index: **16,587 ops/s** (0.060ms mean)

The regression test proxies the 1,000-row snapshot and asserts all projection callers together perform only one indexed pass.

### gmuxd (`go test -bench ... -benchmem`, 3 runs)

- before linear row lookup: **1.04–1.10ms/op**
- after ID index: **0.145–0.163ms/op**
- approximately **7x faster**

## Validation

- `moon run gmux-web:test gmux-web:lint gmuxd:test gmuxd:lint`
- web: 641 tests passed
- gmuxd: all packages passed
- `pnpm exec vitest bench src/family.bench.ts --run`
- `go test ./services/gmuxd/cmd/gmuxd -run '^$' -bench BenchmarkProjectSessionTreeRows -benchmem -count=3`

### DIFF
--- a/apps/gmux-web/src/family.bench.ts
+++ b/apps/gmux-web/src/family.bench.ts
@@ -1,0 +1,32 @@
+import { bench, describe } from 'vitest'
+import { createFamilyIndex, projectFamily } from './family'
+import { makeSession } from './test-helpers'
+
+const agent = (id: string, parent?: string) => makeSession({
+  id,
+  cwd: '/perf',
+  title: id,
+  parent_session_id: parent,
+  semantic_agent: true,
+})
+
+// Mirrors the user-blocking corpus: 1,000 sessions, with 500 children of one
+// task-family root and 499 unrelated sessions.
+const root = agent('root')
+const children = Array.from({ length: 500 }, (_, i) => agent(`child-${i}`, root.id))
+const sessions = [
+  root,
+  ...children,
+  ...Array.from({ length: 499 }, (_, i) => agent(`other-${i}`)),
+]
+
+describe('family projection (1,000 sessions / 500 children)', () => {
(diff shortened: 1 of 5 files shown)


mgabor3141 (commented):

I read every changed line and traced all family callers in store selectors, routing, header, and drawer code. One WeakMap-cached index is reused for an exact computed session-array identity; snapshot/pending changes produce a new array identity. Membership retains the reviewed direct-edge rules (resolved parent, both semantic agents, unpromoted child) and marks complete cycle components unsafe. Child ordering remains `byRecency`; drawer/root/navigation semantics and all pre-existing `family.test.ts` assertions are unchanged and green (the only test-file change adds the bounded-work regression). The daemon twin preserves root existence errors, breadth-first traversal, and child-ID sorting while replacing repeated scans with indexes.

The required ~1,000-session/500-child evidence is present: legacy web projection exceeded 120s while indexed projection averages 0.060ms; gmuxd improves ~1.04–1.10ms to 0.145–0.163ms. Focused repeated Go tests, web family tests, and all required CI checks pass. No sessions were deleted and no remaining findings.

PR #463, opened 2026-08-04
fix(web): ignore collapsed terminal resize measurements

mgabor3141:
## Summary
- reject terminal fit measurements that cannot contain the minimum grid
- prevent a transient zero-width main panel from resizing a live PTY to two columns
- cover collapsed, boundary, and overlay-rounded dimensions

## Corrected investigation scope
The reported garbled screenshot is a **dead-session ReplayView** (the Resume button proves this), so this PR does not claim to fix that screenshot. I reproduced replay of a captured Claude trust prompt at its recorded geometry and it rendered correctly; without the affected raw scrollback plus recorded dimensions, changing replay/reflow would be speculative and risky for cursor-addressed TUI output.

Pressing `2` worked: Claude’s trust prompt maps option 2 to “No, exit”, explaining why the session became dead. Fresh-profile focus and first-command delivery pass end-to-end in Chromium against an isolated daemon. Linux Playwright WebKit could not launch on this host due an ICU symbol mismatch, so the user’s Safari/app-shell path remains unverified.

The fix retained here is independently valid: the fixed desktop sidebar can transiently consume all available width, and the old fit calculation converted a non-positive width into an announced 2-column live PTY resize. It now waits for a usable measurement.

## Tests
- `pnpm exec vitest run src/terminal-fit.test.ts src/terminal-resize.test.ts`
- `pnpm exec tsc --noEmit`
- targeted Biome lint
- isolated Chromium E2E: fresh navigation focused xterm and the first typed command reached shell scrollback

### DIFF
--- a/apps/gmux-web/src/terminal-fit.test.ts
+++ b/apps/gmux-web/src/terminal-fit.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, it } from 'vitest'
+import { terminalGridSize } from './terminal-resize'
+
+describe('terminalGridSize', () => {
+  it('rejects a viewport that cannot fit the minimum grid', () => {
+    expect(terminalGridSize(0, 200, 10, 20)).toBeNull()
+    expect(terminalGridSize(19, 200, 10, 20)).toBeNull()
+    expect(terminalGridSize(200, 19, 10, 20)).toBeNull()
+  })
+
+  it('calculates a usable grid at the minimum boundary', () => {
+    expect(terminalGridSize(20, 20, 10, 20)).toEqual({ cols: 2, rows: 1 })
+  })
+
+  it('supports overlay row rounding', () => {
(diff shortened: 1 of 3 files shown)


mgabor3141 (commented):

The investigation report precedes the code and separates the three reported symptoms. I traced ordinary digit input through xterm `onData` rather than configured key sequences, verified no bare-digit navigation contract exists, and agree the available disconnect log cannot justify speculative focus/proxy/tmux changes. The concrete screenshot shape is explained by the fixed desktop sidebar collapsing available terminal width while the old clamp converted a non-positive measurement into a real 2-column PTY resize.

The fix rejects widths below two cells and heights below one cell before xterm/PTY resize decisions, while preserving the existing 2×1 boundary and overlay rounding. All `measureTerminalFit` callers already handle null as “wait for a usable measurement.” Focused fit/resize tests and all required CI checks pass. Docker-socket waiting and sparse project-update logs remain correctly out of scope. No remaining findings.

PR #464, opened 2026-08-04
feat(family): processes join the family tree

mgabor3141:
## Summary

- treat any unpromoted session with a resolved direct agent parent as a family child
- hide process children from root-level Projects, Activity, and Home projections while keeping them in the existing family drawer/tree
- apply the existing committed-transition notification suppression and cancellation semantics to process children
- preserve process-child family facts across peer projection and update the family capability documentation

## Before / after

**Before:** A shell or other process launched inside an agent retained `parent_session_id`, but appeared as a separate root in Projects/Activity/Home and could emit its own completion notification while the parent agent was active.

**After:** The process appears under its direct agent parent using the existing family UI and is omitted from root lists. Its completion/unread attention is suppressed when the direct agent parent is active at the child's committed inactive transition. Missing parents, non-agent parents, cycles, and promoted sessions continue to fail open as visible roots.

## Provenance

The generic process launch path already calls `inheritLaunchParent` before adapter resolution and stamps `GMUX_SESSION_ID` into `runDirectives.ParentSessionID`; that value is written to runner session metadata. This covers both `gmux -- <cmd>` and `gmux run` fresh nested launches without changing resume/restart provenance.

## Verification

- `pnpm moon run :test :lint`
- `go test -race ./services/gmuxd/cmd/gmuxd -run 'TestCentralNotify' -count=1`

### DIFF
--- a/apps/gmux-web/src/family.test.ts
+++ b/apps/gmux-web/src/family.test.ts
@@ -7,23 +7,25 @@ const agent = (id: string, parent?: string, extra = {}) => makeSession({
 })
 
 describe('task-family projection', () => {
-  it('uses semantic capability and promotion to distinguish presentation roots', () => {
+  it('groups agent and process children of semantic parents unless promoted', () => {
     const root = agent('root')
     const child = agent('child', 'root')
     const shell = makeSession({ id: 'shell', cwd: '/p', parent_session_id: 'root', adapter: 'shell' })
     const promoted = agent('promoted', 'root', { promoted_to_root: true })
     const sessions = [root, child, shell, promoted]
     expect(isFamilyChild(child, sessions)).toBe(true)
-    expect(isFamilyChild(shell, sessions)).toBe(false)
+    expect(isFamilyChild(shell, sessions)).toBe(true)
     expect(isFamilyChild(promoted, sessions)).toBe(false)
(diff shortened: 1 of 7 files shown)


mgabor3141 (commented):

I traced the predicate on both sides. Web family membership now relaxes only the child capability check: the direct parent must still resolve and be semantic, promotion still breaks the edge, self/cyclic ancestry still fails open, and orphan/non-agent-parent sessions remain roots. Root-list filtering reuses that same family predicate. The daemon notification path makes the matching one-condition relaxation while retaining committed inactive-transition timing, direct active semantic-parent requirement, promotion escape, and existing delivery cancellation/locking. Raw process-child family facts are preserved through peer namespacing; there is no separate membership predicate in `peering/projection.go`.

Process provenance is already stamped before adapter resolution for fresh nested `gmux --`/`gmux run` launches via `GMUX_SESSION_ID` → `runDirectives.ParentSessionID` → runner session config, while resume remains excluded. No styling changed. Focused web family/store tests pass, `go test -race ./services/gmuxd/cmd/gmuxd -run TestCentralNotify -count=10` passes, and all required CI checks pass. This is ready to merge before the separate UX implementation. No remaining findings.

PR #466, opened 2026-08-04
fix(web): prevent terminal pane collapse instead of guarding measurement

mgabor3141:
## Summary
- revert the measurement-level two-column guard from #463
- keep the desktop main/terminal pane at a usable 320px minimum beside the fixed sidebar
- preserve full-width mobile behavior by removing the minimum when the sidebar becomes an overlay

When a desktop app window is narrower than the 272px sidebar plus terminal minimum, the layout now clips at the viewport boundary instead of collapsing the terminal to zero width and propagating a degenerate PTY resize.

## Verification
- isolated Chromium E2E repro shrank an attached terminal to a 240px viewport: `.main-panel` remained 320px and xterm stayed above 20 columns; restoring 1200px recovered above 80 columns
- `pnpm exec vitest run src/terminal-resize.test.ts`
- `pnpm exec tsc --noEmit`
- targeted Biome lint
- full `./scripts/build.sh`


mgabor3141 (commented):

The #463 measurement-level mechanism is fully reverted: `terminalGridSize` is removed, its dedicated test is deleted, and `measureTerminalFit` returns to the pre-guard calculation with no dead helper/import left. The replacement acts at the cause: desktop `.main-panel` now has a 320px flex minimum beside the fixed sidebar, so fine-pointer windows narrower than sidebar+pane clip honestly rather than collapsing the terminal and announcing a 2-column PTY.

The responsive interaction is correct: the existing coarse-pointer breakpoint turns the sidebar into an overlay and explicitly resets `.main-panel { width: 100%; min-width: 0; }`, preserving phone/tablet width and mobile toolbar fitting. Normal desktop widths are above the new minimum, so their measured PTY geometry is unchanged. The isolated shrink/recovery evidence reports >20 columns at 240px and >80 after restoring 1200px; all required CI checks pass. No remaining findings.

PR #453, opened 2026-08-02
feat: add root-led agent families

mgabor3141:
## Summary

Add root-led task-family navigation for semantic agent sessions and suppress redundant completion attention for children still managed by an active parent.

### Web UI

- Treat a session as a family child only when its resolved direct parent and the child both carry the server-derived `semantic_agent` capability, the child is unpromoted, and the IDs differ.
- Keep shells, shell-parent/child edges, missing-parent orphans, and promoted sessions visible as roots.
- Show one root-titled row per family across Projects, Activity, and Home; selecting a child highlights the root row, and a relevant child keeps a dead root locatable.
- Route children through the presentation root's project while retaining the selected child's host/adapter/slug.
- Add an Agents / Family header pill, compact Parent/Root controls, and an overlay drawer with ancestor/sibling/descendant context. The drawer remains open across family navigation and restores focus to the selected row.
- Preserve immutable launch provenance; `promoted_to_root` only breaks the presentation edge.

### Daemon notifications

- On a semantic child Active → Inactive transition, suppress user-facing completion/unread attention only when its direct semantic-agent parent is Active in the committed outcome stream.
- Cancel existing pending attention for the child when suppression applies.
- Do not recurse through missing/inactive parents; promoted children and shell endpoints retain root notification behavior.
- A later parent transition does not retroactively revise the child's committed suppression decision.

## Integration notes

- **#436 routing:** kept centralized `navigate()` / `tabHref()` tab-param carrying and sidebar-mode repair. Family serialization is implemented inside the current `viewToPath` / `resolveViewFromPath` model. Slug-collision detection now uses the same root-project namespace as family routing.
- **#448 / #450 adapter capability changes:** `semantic_agent` is derived on the server from `adapter.ConversationSource`, not adapter-name inference in the frontend. Tests pin Pi, Claude, and Codex as semantic and shell as non-semantic, including the current Codex rollout adapter.
- Family membership is resolved against the complete snapshot rather than a tab-filtered subset, so filtering cannot manufacture an orphan or expose a child row.
- Snapshot equivalence excludes the new post-cutover family fields from the retired legacy composer comparison; direct converter tests pin their wire behavior.

### ACP seam

[...]

PR #467, opened 2026-08-04
feat(web): family drawer noise reduction and status roll-up

mgabor3141:
Implements the P0/P1 findings and the noise-reduction design from the family-UX assessment (follow-up to #453/#462, builds on #463 and #464's process children). Amended per adversarial review round 1 (see thread).

## Noise reduction (the big one)

Every children list in the drawer — the top-level sibling list included — is partitioned into **attention → working → idle → finished** groups:

- **Bucketing follows `sessionDotState` precedence exactly** (error > working > unread; error/working alive-gated, unread not), so a row's dot and its group never disagree. A dead member whose final output was never viewed classifies as *attention* — unseen output pings until viewed.
- **Subtree-max inheritance**: a node sorts into the highest-urgency bucket found in its entire subtree, so a collapsed *finished* summary can never hide an urgent descendant.
- **Caps ∞ / 20 / 10 / 3** (attention/working/idle/finished) behind ghost summary rows (`+66 finished`), two-state expansion keyed per (parent, bucket), reset when the drawer closes.
- **Within a bucket**: agents before processes, recency desc, then natural title sort (`swarm-425/426/…` reads ordered, not shuffled by last-output timestamps).
- **Frozen against ordinary live updates; explicit actions re-project.** The drawer's freeze/expansion behavior lives in a pure model (`family-drawer-model.ts`): live status changes repaint dots in place but never re-sort rows under the cursor; selecting another session or toggling a group re-projects from current session facts.
- **Sticky heading** with status-truth counts (`500 need attention · 4 working · 10 idle · 69 finished · 1 process`); only the tree scrolls.

## Status roll-up outside the drawer (P0)

A root row stands in for its whole family, so it must show the family's state:

- New `familyDotById` computed: aggregates the highest-precedence member dot (`sessionDotState` vocabulary, dead members included — a never-viewed dead child surfaces unread) onto the presentation root; used by both the sidebar row and the dashboard session row. Selection muting is applied per member *before* aggregation — viewing one child no longer mutes a sibling's attention.
- `unreadCount` (logo blink, mobile hamburger badge) now adds alive unread descendants to their folder-visible root (keeps the count's existing alive gate). The tab-title count already scanned raw sessions and needed no change.

## Five-state vocabulary in the drawer (P0)

Drawer rows use the shared `.session-dot-indicator` states via `sessionDotState` (error/working/unread/active/fading), verbatim.

## Process children (#464 presentation)

`$` glyph instead of a dot (teal while running — shape survives color-blindness), muted title, sorted after agents within their bucket, own summary noun (`+12 processes done`), excluded from the pill count.

## P1 polish

- Opaque drawer background (terminal text no longer ghosts through).
- Pill → `Agents · N` (alive agents in the family); drawer heading → "Agents"; header root glyph `⌂` → `⇈`.

## Perf

Projection still builds on the shared O(n) `familyIndex`; the #462 proxy guard also covers `bucketedFamily` (still ≤ n+1 indexed reads). New bench: bucketing the 500-child drawer ~0.9 ms per explicit re-projection.

## Tests

Unit + behavioral: bucket classification per dot precedence, subtree-max hoisting, in-bucket ordering, pill count, roll-up aggregation/muting, `unreadCount` child roll-up, and a drawer-model suite proving (a) live status change does **not** auto-reorder while a group toggle **does** refresh the projection, (b) the rendered 20/10/3 caps are applied by the actual split path with attention uncapped, and (c) mixed hidden agent/process summary wording (`+4 finished · 2 processes done`).

## Screenshots

Live specimen: 578-child family (562 dead `swarm-*`), plus a small normal family (3 agents + 2 processes). Before = pre-change build on the same data. After-shots reflect round-1 spec semantics.

| | Before | After |
|---|---|---|
| Monster family, drawer opened | before top | after top |
| Monster family, scrolled to the end | before wall (~38 screens, unordered) | after summary (grouped; ends at `+66 finished`) |
| Normal family (3 agents, 2 processes) | before normal | after normal |

<details><summary>More after-shots: expanded finished group, child-selected view, mobile (taken pre-round-1; caps/expansion mechanics unchanged, attention counts smaller than current semantics produce)</summary>

Expanded finished group → ends with `show fewer` (sticky heading holds):
expanded

Child selected (spine + Parent, groups at sibling level):
child

Mobile (390px):
mobile

</details>

Screenshots live on the throwaway `mgabor/family-drawer-shots` branch (not for merge; delete after this PR closes).

**Owner decision point (product, not blocking):** with the approved semantics, attention is uncapped and unread is not alive-gated, so a corpus of hundreds of dead never-viewed children renders ~500 attention rows (see monster after-shot). If that's undesirable, the amendment would be either bulk mark-read tooling or a spec change (dead+unread → finished, as this PR briefly did before round 1 reverted it).

### DIFF
--- a/apps/gmux-web/src/family-drawer-model.test.ts
+++ b/apps/gmux-web/src/family-drawer-model.test.ts
@@ -1,0 +1,128 @@
+import { describe, expect, it } from 'vitest'
+import { splitGroup, syncDrawer, toggleDrawerGroup, groupKey } from './family-drawer-model'
+import type { Session } from './types'
+import { makeSession } from './test-helpers'
+
+const at = (minute: number) => `2026-08-04T10:${String(minute).padStart(2, '0')}:00Z`
+const root = makeSession({
+  id: 'root', cwd: '/p', title: 'root', semantic_agent: true,
+  created_at: at(0), last_output_at: at(1),
+})
+const child = (id: string, extra: Partial<Session> = {}) => makeSession({
+  id, cwd: '/p', title: id, parent_session_id: 'root', semantic_agent: true,
+  created_at: at(0), last_output_at: at(1), ...extra,
+})
+const procChild = (id: string, extra: Partial<Session> = {}) => makeSession({
+  id, cwd: '/p', title: id, parent_session_id: 'root', adapter: 'shell',
+  created_at: at(0), last_output_at: at(1), ...extra,
+})
+
+/** The root-selected drawer has one top-level group containing the root
+ * node; its children groups are what the drawer body renders. */
+function childGroups(model: ReturnType<typeof syncDrawer>) {
+  return model.projection.groups[0].nodes[0].groups
+}
+
+describe('drawer model: frozen ordering with explicit refresh', () => {
+  it('ignores ordinary live updates but re-projects on group toggle', () => {
+    const mover = child('mover')
+    const before = [root, mover, child('idle-friend')]
+    const model = syncDrawer(null, root, before)
+    expect(childGroups(model).map(g => g.bucket)).toEqual(['idle'])
+
+    // Live update: `mover` starts working. An ordinary re-render syncs
+    // against the new snapshot and MUST return the same frozen model —
+    // no auto-reorder under the cursor.
+    const after = [root, { ...mover, status: { active: true } }, child('idle-friend')]
+    const synced = syncDrawer(model, root, after)
+    expect(synced).toBe(model)
+    expect(childGroups(synced).map(g => g.bucket)).toEqual(['idle'])
+
+    // Explicit user action: toggling any group re-projects from current
+    // facts — `mover` now leads a working group.
+    const toggled = toggleDrawerGroup(synced, root, after, groupKey('root', 'idle'))
+    expect(toggled).not.toBe(synced)
+    expect(childGroups(toggled).map(g => g.bucket)).toEqual(['working', 'idle'])
+    expect(childGroups(toggled)[0].nodes.map(n => n.session.id)).toEqual(['mover'])
+  })
+
+  it('tracks expansion per (parent, bucket) across toggles and selection changes', () => {
+    const snapshot = [root, ...Array.from({ length: 5 }, (_, i) => child(`dead-${i}`, { alive: false }))]
+    let model = syncDrawer(null, root, snapshot)
+    const key = groupKey('root', 'finished')
+    model = toggleDrawerGroup(model, root, snapshot, key)
+    expect(model.expanded.has(key)).toBe(true)
+    // Selecting another family member re-projects but keeps expansion.
+    const other = snapshot[1]
+    const reselected = syncDrawer(model, other, snapshot)
+    expect(reselected).not.toBe(model)
+    expect(reselected.expanded.has(key)).toBe(true)
+    // Toggling again collapses back (two-state, no increments).
+    const collapsed = toggleDrawerGroup(reselected, other, snapshot, key)
+    expect(collapsed.expanded.has(key)).toBe(false)
+  })
+})
+
+describe('drawer model: rendered caps and summary rows', () => {
+  it('applies the 20/10/3 caps to what is actually shown', () => {
+    const snapshot = [
+      root,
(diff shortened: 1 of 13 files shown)


mgabor3141 (commented):

The screenshots, client-only roll-up, five-state agent dots, process presentation, recursive grouping/caps structure, and #462 bounded-snapshot guard are present, but two implementation choices diverge from the approved design that this track was explicitly asked to implement exactly:

1. `familyBucket` forces every dead session to `finished` before checking error/unread. The design says Attention = error or unread using `sessionDotState` precedence, with no dead exception; the PR body itself acknowledges this deviation. Restore the specified classification (and roll-up behavior) rather than changing product semantics based on the current corpus, unless the owner separately approves a spec amendment. Update the tests that currently cement dead+unread/error as finished.

2. The design freezes ordering until an explicit user action, then says expansion/collapse re-projects from current facts. `FamilyDrawer` stores one projection per selected ID and `toggle` only changes a Set, so expanding/collapsing never refreshes stale bucket membership/order. Rebuild the frozen projection from `sessions.peek()` on each bucket toggle (while preserving expansion state), then reveal/cap the refreshed group. Add a regression that changes status while open, verifies no automatic reorder, toggles a bucket, and verifies the projection refreshes.

Also add a behavioral guard for the rendered caps/summary split (including mixed hidden agent/process wording), since the current test only asserts the cap constants rather than proving 20/10/3 are applied by the drawer.

I otherwise verified family membership helpers are untouched, aggregation is client-side, process children remain excluded from `Agents · N`, recursive subtree-max prevents a lower bucket hiding urgent descendants, and all current CI checks pass.

mgabor3141:
Round 1 addressed in e63400b:

1. **Spec bucket semantics restored.** `familyBucket` now follows `sessionDotState` precedence with no dead early-return: unread is not alive-gated, so dead never-viewed members classify as attention; `familyDotById` dropped its dead-descendant gate for the same reason. Tests updated to cement the spec (`dead+unread → attention`; dead+error with no unread stays finished because error is alive-gated in the dot vocabulary). `unreadCount` keeps its pre-existing alive gate, matching how roots are counted today.

2. **Toggle re-projects.** Freeze/expansion state moved into a pure model (`family-drawer-model.ts`): `syncDrawer` returns the identical model for ordinary live updates (regression asserts `===` and unchanged grouping after a status flip), `toggleDrawerGroup` flips the (parent, bucket) key and rebuilds the projection from `sessions.peek()` (regression asserts the flipped member now leads a new working group). Expansion survives selection changes, resets on close via unmount, and stays two-state.

3. **Behavioral cap/summary coverage.** `splitGroup` is the exact code path `GroupRows` renders from; new tests drive real projections through it: 25/14/10 children → 20/10/3 shown with `+5 more working` / `+4 idle` / `+7 finished`; 30 unread children → uncapped, no summary row; expanded → all rows + `show fewer`; mixed remainder → `+4 finished · 2 processes done`; process-only remainders → `+2 processes done` / `+2 processes`.

Monster/normal screenshots retaken under the restored semantics (PR body updated). One consequence surfaced for the owner: with attention uncapped and unread not alive-gated, this corpus (562 dead swarm children, mostly never viewed) renders ~500 attention rows — flagged in the body as a product decision (bulk mark-read tooling vs. a future spec amendment), not blocking this PR.

mgabor3141 (commented):

The delta now matches the approved specification. `familyBucket` mirrors `sessionDotState` exactly, including dead-unread attention and alive-gated error/working; the root roll-up uses the same vocabulary. Ordinary session updates retain the identical frozen drawer model, while selection and bucket-toggle actions rebuild from the current snapshot and preserve per-parent/bucket expansion state. The extracted `splitGroup` is the actual render path and its tests prove uncapped attention, 20/10/3 visible limits, expanded/show-fewer behavior, and mixed agent/process summary wording.

I re-ran family, drawer-model, and store tests plus TypeScript checking; all pass. #462’s bounded snapshot guard remains green, family membership code is unchanged, aggregation remains client-side, CI is fully green, and the PR body contains updated before/after monster and normal-family screenshots. The documented uncapped dead-unread corpus consequence is a product decision inherent in the approved spec, not an implementation defect. No remaining findings.

PR #468, opened 2026-08-04
feat: add session drive-mode axis and ADR 0033 capability refusals

mgabor3141:
Slice 0 of the ACP host program (`docs/acp-host-program.md`, #455/#465): the drive-mode session axis and the honest ADR 0033 capability refusals. No ACP code — this is the boundary work everything else keys on.

**Mode axis**
- `drive_mode` (`terminal`|`acp`) as a durable session fact: schema migration v2 (v1 stays frozen; backfill default `terminal` is exact — every pre-migration session was a terminal session), `RunnerRegistration`/`NewSession` validation (unknown modes refused, never defaulted), immutable across re-registration (`ErrDriveModeMismatch` — conversion is a future explicit operation).
- PTY runner `/meta` splices `drive_mode: "terminal"` beside the incarnation; gmuxd registration persists it.
- Wire: `drive_mode` emitted only when not terminal (pre-mode shape stays byte-identical); protocol zod schema defaults absence to `terminal`; peering projection carries it; statetool export includes it.
- Display: `gmux ls` marks the adapter cell `claude (acp)` and round-trips `drive_mode` in `--json`; web session menu shows `adapter (acp)`.

**Refusals (final copy, per the program's settled decision 3)**
- New daemon-side gate for `agent prompt/cancel`, checked **before residency/readiness/activity** — so prompting a dead terminal claude session no longer transparently resumes a real Claude TUI just to be refused by its runner one round-trip later.
- Terminal claude/codex: `unsupported_mode` (422) naming the mode boundary — ADR 0033 §3 wording; `--steer` gets the permanent variant. Shell/editors keep the runner's `unsupported_adapter` wording, now pre-delivery. Unknown adapters pass through: the daemon never invents a capability verdict. The runner's own refusal stays as defense in depth.
- ACP-mode sessions (representable in tests; producible from slice 1): PTY surfaces answer `no_terminal` naming the working surface — raw input → `gmux agent prompt`, scrollback/tail → `gmux agent logs`, attach/terminal-WS → the web conversation view. CLI `attach` refuses client-side with the same copy; `tail`/`send` surface the daemon message.
- New `adapter.ACPDrivable` capability marker (claude, codex) — slice-0 consumer is the refusal wording, per the "every advertised capability has a consumer in a shipped slice" rule.

Tests: verdict table over the production adapter lookup, refusal-precedes-resume, cancel at the boundary, PTY-surface refusals on an ACP row, drive-mode persistence/mismatch/validation, wire emission rules, `/meta` splice, migration checksums (v2 frozen alongside v1).

### DIFF
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -376,7 +376,10 @@ function SessionMenu({ session, onRestart, onResume, resuming }: {
           <div class="session-menu-section-title">Session info</div>
           <div class="session-menu-row">
             <span class="session-menu-label">Adapter</span>
-            <span class="session-menu-value">{session.adapter}</span>
+            <span class="session-menu-value">
+              {session.adapter}
+              {session.drive_mode === 'acp' ? ' (acp)' : ''}
+            </span>
           </div>
           <div class="session-menu-row">
             <span class="session-menu-label">Version</span>
--- a/apps/gmux-web/src/types.ts
+++ b/apps/gmux-web/src/types.ts
@@ -24,6 +24,12 @@ export interface Session {
   remotes?: Record<string, string>
   adapter: string
   /**
+   * Drive mode (ADR 0033): how gmux hosts this harness. Absent means
+   * terminal; 'acp' sessions have no PTY. Server-derived; never inferred
+   * from adapter names.
+   */
+  drive_mode?: 'terminal' | 'acp'
+  /**
    * Session this one was spawned from (e.g. `gmux edit` invoked as
    * $EDITOR inside an existing session). Drives adjacent placement in
    * the sidebar: the child renders directly under its parent.
--- a/cli/gmux/cmd/gmux/actions.go
+++ b/cli/gmux/cmd/gmux/actions.go
@@ -20,10 +20,14 @@ import (
 // session is the subset of gmuxd's Session model that the CLI cares
 // about. Defined locally to avoid pulling in the gmuxd store package.
 type cliSession struct {
-	ID            string `json:"id"`
-	Peer          string `json:"peer,omitempty"`
-	Cwd           string `json:"cwd,omitempty"`
-	Adapter       string `json:"adapter"`
+	ID      string `json:"id"`
+	Peer    string `json:"peer,omitempty"`
+	Cwd     string `json:"cwd,omitempty"`
+	Adapter string `json:"adapter"`
+	// DriveMode is how gmux hosts this harness (ADR 0033). Absent on the
+	// wire means terminal; "acp" sessions have no PTY. Round-trips through
+	// `gmux ls --json` so scripts can see the mode axis.
+	DriveMode     string `json:"drive_mode,omitempty"`
 	Alive         bool   `json:"alive"`
 	Pid           int    `json:"pid,omitempty"`
 	Title         string `json:"title,omitempty"`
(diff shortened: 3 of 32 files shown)


mgabor3141 (commented):

The durable axis, migration/backfill, immutable re-registration, PTY `/meta`, wire/defaulting, and terminal Claude/Codex refusals are well covered, but three honesty gaps remain:

1. `agent_mode_gate.go:68` allows **every** ACP-mode row before consulting the harness capability. That makes known unsupported pairs such as `(pi, acp)` and `(shell, acp)` pass semantic delivery, contrary to ADR 0033’s rule that capabilities attach to `(harness, mode)` and to the no-speculative-capabilities guard. For a known adapter in ACP mode, pass only `ACPDrivable` harnesses; refuse other known pairs explicitly with an honest `unsupported_mode`/capability error. Unknown adapters may still defer to their runner. Add table cases for pi+ACP and shell+ACP.

2. The raw-input mode check in `serve_central.go:1168` silently ignores store errors (`storeErr == nil && ...`) and proceeds to the runner. On a store failure it therefore cannot establish the mode but may still write to a live ACP endpoint. Return an internal error on store failure, then apply the mode refusal; do not bypass a safety boundary because persistence is unavailable. Add a scheduled store-error test if the handler seam permits it.

3. The claimed terminal-WebSocket refusal is snapshot-dependent: the wsproxy resolver checks only `fanout.Current().Sessions`. Immediately after ACP registration, or whenever composition is stale/unavailable, it falls through to the live registry endpoint and attempts a terminal proxy. Use the authoritative store (and handle its errors conservatively) before resolving a backend. Add a real `/ws/{id}` regression with an ACP store row and a stale/empty fanout; the existing `TestPTYSurfacesRefusedOnACPSession` exercises the REST attach action, not the WebSocket route.

These are all boundary fixes, not new capabilities. Current CI is green and the remaining refusal paths are explicit rather than no-ops.

mgabor3141:
Round 1 findings addressed in `0f6fbbb` (amended in place; all boundary-only):

1. **(harness, mode) enforced in the gate.** ACP mode now passes only for harnesses that have an ACP drive mode (`adapter.SupportsACPDrive`); known contradictory pairs — `(pi, acp)`, `(shell, acp)` — are refused explicitly with `unsupported_mode` ("has no ACP drive mode … refused rather than guessed at"). Unknown adapters defer to their runner in either mode. Table cases added: pi+ACP (prompt and steer), shell+ACP, unknown+ACP.

2. **Raw input fails closed on store errors.** The `/input` mode check now returns `500 internal` when the store read fails, and only then applies the ACP refusal — bytes never reach a live endpoint whose mode could not be established. Regression: `TestInputRefusesOnStoreFailure` (row inserted, store closed, POST input → 500/internal).

3. **Terminal-WS gate reads the authoritative store.** The `/ws/{id}` resolver is extracted as `terminalWSEndpoint` and answers the mode boundary from `boot.Store` before any registry fallthrough; store failures refuse conservatively ("drive mode could not be verified"). Regressions: `TestTerminalWSEndpointRefusesACPSessionFromStore` reproduces the exact reported state — a real coordinator registration of an ACP runner (live registry endpoint) with an **empty fanout** — and asserts refusal, plus a terminal session still resolving; `TestTerminalWSEndpointRefusesOnStoreFailure` covers the conservative path.

Full gmuxd/cli/adapter suites, vet, gofmt green locally.

mgabor3141 (commented):

The delta closes all three findings. Known ACP pairs now pass only when the harness implements `ACPDrivable`; pi/shell ACP rows refuse explicitly while unknown adapters still defer without an invented verdict. Raw input fails closed on store errors before endpoint lookup. Terminal WebSocket resolution now reads the authoritative store before registry/fanout fallback, with regressions for an ACP registration against an empty snapshot and for store failure.

I re-ran race suites for gmuxd, centralstore, and adapters; all pass, as do all required CI checks. The complete slice now carries a durable immutable mode axis end-to-end, preserves terminal wire shape, keys semantic capability decisions on `(harness, mode)`, emits explicit actionable refusals, and advertises no unconsumed ACP/fs/terminal capability. No remaining findings.

mgabor3141:
Rebased onto latest main (`be4b8f0`). One real conflict: `runnerMetaWire`'s status field index in the JSON-shape pin test, where #470's additive `exit_code`/`exited_at` and this PR's `drive_mode` both grew the struct — both kept, index resolved to 23 and the exact-table test asserts the merged shape. #470's exit metadata rides `session.State`'s own document, so the `/meta` splice (incarnation + drive_mode) merged without interaction. Full local suites (gmuxd/cli/adapter Go, protocol + web, biome, sqlc-regen check, gofmt) and all 11 CI checks are green.

PR #470, opened 2026-08-04
fix: wave-2 sweep — exit metadata, index slug retention, mock timestamps

mgabor3141:
## Summary
- preserve runner `exit_code` and `exited_at` through `/meta`, live exit events, and SSE reconnect replay
- retain the last known conversation slug/title during transient empty-slug refreshes
- correct impossible timestamp ordering in mock sessions

## Verification
- `pnpm test`
- `pnpm lint`

### DIFF
--- a/apps/gmux-web/src/mock-data/sessions/codex-refactor-adapters.ts
+++ b/apps/gmux-web/src/mock-data/sessions/codex-refactor-adapters.ts
@@ -33,7 +33,7 @@ export default {
   status: { active: true },
   unread: false,
   project_slug: 'my-project',
-  last_output_at: ago(3),
+  last_output_at: ago(0),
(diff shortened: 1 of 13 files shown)


mgabor3141 (commented):

I traced all three fixes. Runner exit metadata is additive: `SetExited` now emits the original RFC3339 timestamp, `/meta` already serializes the state fields and now consumes them daemon-side, and reconnect replay emits the same exit event; older events without `exited_at` retain the observation-time fallback. The state snapshot is lock-safe, and focused race suites for session, ptyserver, gmuxd, and conversations pass.

The empty-slug update keeps the established key/slug and only retains the prior title when the transient update has no title; a later real slug/title update still applies. Mock timestamps are adjusted or explicitly seeded so creation/start precede output/exit. All required CI checks pass. No remaining findings.

PR #471, opened 2026-08-04
chore: delete dead tests/e2e, add container smoke scenario

mgabor3141:
## Summary
- delete the dead standalone `tests/e2e` module and all repository references
- build the production `gmux` binary in the isolated, networkless production-container harness
- add a real CLI-to-daemon registration, durable crash-death, restart repair, and socket/process cleanup smoke scenario

## Validation
- `pnpm test`
- `pnpm lint`
- `services/gmuxd/tools/production-e2e.sh` (new scenario observed passing)


mgabor3141 (commented):

Deletion/reference cleanup and CI execution are verified, but the new smoke has one test-harness race:

- `production_container_e2e_test.go` assigns the child process stdout/stderr to a `bytes.Buffer`, then polls `output.String()` while `os/exec` may concurrently copy the marker into that buffer. `bytes.Buffer` is not concurrency-safe; this is an uninstrumented data race in the container test and can make the new required smoke flaky. Use a synchronization-safe writer/read path (for example a mutex-protected buffer) or have the isolated child write the marker to a fixture file and poll that file. Preserve the assertion that the launched payload actually executed.

Everything else matches the approved scope: `tests/e2e` is removed from `go.work` and Docker context, repository-wide grep finds no lingering reference, existing container scenarios are unchanged, and PR run 30926192664 explicitly logs `TestProductionContainerE2E/real_cli_registration_and_crash_cleanup` RUN/PASS in 5.13s.

mgabor3141:
Addressed round 1 advisory in `84209bb`: replaced concurrent `bytes.Buffer` polling with a child-written marker fixture file. The harness now polls and verifies the exact marker file contents, preserving proof that the payload executed without shared-memory access.

Validation: `pnpm lint`; package compile check; full `services/gmuxd/tools/production-e2e.sh` (real CLI smoke PASS in 5.16s, full container suite PASS).

mgabor3141 (commented):

The delta removes the unsynchronized output capture and has the isolated payload atomically demonstrate execution through a marker file whose exact contents are polled. Registration, durable crash death, process exit, daemon-restart socket cleanup, and non-resurrection assertions remain intact.

Repository-wide grep finds no remaining `tests/e2e` references; only the dead module, its `go.work` entry, and obsolete Docker copy are removed, with all existing container scenarios untouched. On the amended head, CI run 30927067108 explicitly logs `TestProductionContainerE2E/real_cli_registration_and_crash_cleanup` RUN and PASS (5.10s), and every required check is green. No remaining findings.

PR #472, opened 2026-08-04
feat: child session attention semantics

mgabor3141:
## Summary

- make unread universal for completed agent/process outcomes, including failed agent turns
- suppress completion delivery only from the direct parent's committed activity at completion (agent semantic activity or OSC 133 process activity), independent of family presentation/promotion
- consume results from successful `gmux wait`, `tail`, `agent logs`, prompt/steer/send, explicit `gmux read`, and web interaction
- add `gmux read <id>...` and `gmux read --family <root-id>` for one-time retained child-pile remediation
- preserve focused-view unread until the next click/keypress while suppressing notification delivery for that focused+visible session
- roll live and retained-dead unread children into the family root's attention count

## Observed-result acknowledgement protocol

Every completed result advances an opaque `unread_token`, even when the previous result is still unread. The token is projected additively through runner events/meta, durable central state, session snapshots, wait responses, scrollback/conversation response headers, the CLI session model, and the web session model.

Consumers acknowledge the exact observation they rendered:

- wait returns the token in its resolved outcome (including predicate matches)
- tail and agent logs capture `X-Gmux-Unread-Token` with the rendered response
- web selection/interaction captures `{id, token}` from the viewed snapshot; its optimistic overlay is token-scoped
- explicit `gmux read` uses the token from the session snapshot it resolved

`POST /v1/sessions/{id}/read?token=...` remains incarnation-conditional for live runners and is now token-conditional for both live and retained-dead owners. If completion N+1 wins after a consumer observed N, the owner returns `409 result_changed`; the consumer treats N as successfully consumed while N+1 remains unread. Dead-session acknowledgement additionally retains the existing row-version/lifecycle fencing.

Result-producing status/exit edges fuse `unread + unread_token` with the completion edge. This both gives waiters the exact generation and preserves the settled notification ordering: direct-parent suppression is committed before unread can schedule delivery. The committed-transition latch, late-unread suppression, and `deliveryMu` pending→active→send cancellation atomicity remain intact; a rejected old acknowledgement does not cancel newer attention.

Existing retained rows migrate with an empty token, which is still a valid observed generation (`?token=`). New completions immediately receive a fresh token.

### Runner replacement fence

Acknowledgement owner resolution now takes the coordinator lifecycle mutex. A request arriving while registration is between durable commit and live-generation install waits for that window to settle, then targets the installed runner or retained-dead row. Token-bound dead acknowledgement returns `ErrAckOwnerChanged` instead of success whenever a live/fenced generation owns the result. Runner replies are followed by a second lifecycle-serialized owner check; if that generation was displaced in flight, acknowledgement retries against the surviving owner. Notification cancellation happens only after this proof succeeds.

## Settled semantic changes from #453 / #464

The earlier regression tests are consciously updated where the owner-settled launcher contract supersedes family-presentation semantics:

- promotion no longer exempts a child from direct-parent delivery suppression
- an active process parent now suppresses its direct child; suppression is not semantic-agent-only
- a dead parent is inactive even if its retained status fact was active

## Activity seam

Process activity already exists on main through runner OSC 133 prompt-cycle tracking (command active → prompt idle), so this change uses the first-class committed `Active` fact. No new OSC tracking is introduced. The focused-session inactivity timer remains an explicit future seam.

## Deterministic race coverage

The tests force N to be observed, advance state to N+1, and only then release/send N's acknowledgement; they do not depend on sleeps:

- `TestWaitDelayedAcknowledgementCannotClearNewerCompletion`
- `TestTailDelayedAcknowledgementCannotClearNewerCompletion`
- `TestAgentLogsDelayedAcknowledgementCannotClearNewerCompletion`
- focused-view delayed acknowledgement in `store.test.ts`
- `TestReadEndpointConsumesUnreadForExactRunner`
- `TestCentralReadRouteRejectsDelayedAcknowledgementForNewerCompletion`
- `TestAcknowledgeDeadSessionTokenRejectsDelayedResult`
- fused completion/suppression projection and router tests
- `TestAcknowledgeDeadTokenNeverSucceedsForLiveOrFencedOwner`
- `TestAcknowledgementRuntimeWaitsForReplacementInstall`
- `TestCentralReadRouteWaitsForCommitToInstallFenceBeforeCancellation` parks registration after durable commit, proves delivery remains pending, installs the live owner, blocks its valid `/read` response after clearing, and proves cancellation occurs only after acknowledgement completes
- `TestAcknowledgeSessionRetriesSameEndpointIncarnationReplacementGeneration` forces generation 1 → 2 while endpoint and incarnation remain identical and requires a second runner `/read`

## Verification at `3685f3b335e5be3917cd892031b41051a4bb588b`

- rebased onto latest `main` (`614cc88f0f39`)
- `go test ./cli/gmux/... ./services/gmuxd/...`
- `go test -race -count=1 ./cli/gmux/cmd/gmux ./cli/gmux/internal/ptyserver ./services/gmuxd/cmd/gmuxd ./services/gmuxd/internal/presence ./services/gmuxd/internal/sessioncoord ./services/gmuxd/internal/centralstore`
- deterministic delayed-result race tests repeated 25× under `-race`; replacement-fence owner/route tests repeated 20× under `-race`; same-endpoint/incarnation generation replacement repeated 100× under `-race` after the final rebase
- `go test ./services/gmuxd/...`
- `go test -race -count=1 ./services/gmuxd/cmd/gmuxd ./services/gmuxd/internal/sessioncoord ./services/gmuxd/internal/centralstore`
- `pnpm moon run gmux-web:test gmux-web:lint protocol:lint gmuxd:check-centralstore-generated gmuxd:check-centralstore-cross-build` — web: 34 files / 666 tests


mgabor3141 (commented):

The direct-parent latch, committed event ordering, late-unread suppression, and `deliveryMu` cancellation machinery remain intact, and OSC 133 activity is real on main (`PromptMarkTracker` C→active, D/A→idle). Four correctness issues remain:

1. **A terminal crash during an OSC-tracked command can still produce neither unread nor notification.** `finalizeSessionState` sets unread only when the lifetime turn remains open. Once OSC prompt tracking has taken over, a shell that dies while `Active=true` follows the documented “stays active and resolves as died” path, calls only `SetExited`, and never sets unread. In the router the dead snapshot becomes inactive, but `finished` is gated on `cur.Alive`, so an inactive/dead direct parent receives no unread notification either. This violates universal completion unread and the required process-crash notification case. Add a real state/runner regression for OSC C→runner/child exit before D, with inactive/dead parent, and make the terminal completion persist unread without breaking intentional interruption semantics.

2. **Cancel consumes unread even though it is not a consumption boundary in the settled spec.** `handleCancel` shares `deliver`, whose unconditional post-write `SetUnread(false)` includes `ActionInterrupt` (the comment even says “or interrupt”). Canceling the current turn does not prove the caller consumed a previous result. Clear only for prompt/follow-up/steer delivery, not interrupt, and add a cancel-preserves-unread test.

3. **Tail and agent logs acknowledge before output is successfully delivered to the caller.** Both call `consumeSession` before `os.Stdout.Write`. A broken pipe or write failure therefore clears unread even though the command fails and the result was not consumed. Move acknowledgement after successful output write; add a failing-writer seam/test for both paths so failures preserve unread. Wait already has the correct render-then-ack order.

4. **Hidden/unfocused selection still suppresses delivery.** Although `AnyViewing` now correctly requires focused+visible, `presence.OnSessionSelected` unconditionally calls `notifier.CancelForSession`; any hidden tab retaining/selecting that ID can cancel pending/active delivery. Remove or gate that callback by the same focused+visible predicate. Actual web interaction already calls `/read`, which performs cancellation under `deliveryMu`. Add presence/router coverage proving hidden/unfocused selection does not cancel while focused+visible selection/interaction does.

Also remove the PR/docs claim that terminal attach consumes unread unless you intentionally implement and test it; attach is not among the strict consumption boundaries in this review brief. Current CI is green, but these schedules are not covered.

mgabor3141:
Addressed all four round-1 blockers in the existing two-commit stack:

1. **OSC crash completion:** terminal finalization now sets unread on process exit even after prompt marks replaced the lifetime turn, while preserving an explicitly interrupted turn's unread state. Coverage includes a real runner/child (`OSC 133 C` → exit 7 before D), event ordering (unread meta before exit), retained OSC-active status, an interruption regression, and a dead-parent crash notification outcome.
2. **Cancel:** `ActionInterrupt` is excluded from the shared post-delivery unread clear. A successful cancel with pre-existing unread now proves unread is preserved; prompt delivery still consumes.
3. **Output-before-ack:** tail and agent logs now write successfully before POSTing `/read`. Both have injected failing-writer regressions proving command failure sends no read acknowledgement.
4. **Presence:** production no longer installs the unconditional `OnSessionSelected` cancellation callback. Hidden and unfocused selected-client router cases now prove notification delivery remains pending; focused+visible viewing still suppresses delivery without clearing unread, and the `/read` interaction route has direct coverage proving it cancels pending attention.

I also removed terminal attach from the docs and PR consumption claims.

Fresh evidence:

- full `go test ./cli/gmux/... ./services/gmuxd/...`
- web: 34 files / 664 tests
- web + protocol lint/typecheck tasks
- notification/read races 10× under `-race`
- OSC/cancel/output-failure regressions 5× under `-race`

mgabor3141 (commented):
Round 1 — AMEND

**High: consumption acknowledgements are not tied to the result that was actually consumed, so completion-vs-consumption can erase a newer result.** `waitSession` renders a successful wait response and only afterwards calls `consumeSession`; `tail`/`agent logs` likewise fetch/write one snapshot and then POST `/read` (`cli/gmux/cmd/gmux/wait.go`, `actions.go`, `agent.go`). The daemon forwards an unconditional `/read`, and the runner only checks runner incarnation before `SetUnread(false)` (`services/gmuxd/cmd/gmuxd/consume_central.go`, `cli/gmux/internal/ptyserver/ptyserver.go`). Reproducible schedule: wait observes completion N; before its `/read` arrives another actor starts and completes turn N+1 in the same runner incarnation; the delayed acknowledgement clears N+1 unread. The equivalent race exists for tail/logs and web interaction against a completion concurrent with the request. Incarnation protects runner replacement, not outcome replacement. This violates universal unread and the required completion-vs-wait/view race semantics. Please make acknowledgement conditional on the observed completion/version (or otherwise linearize consumption so it cannot clear a later unread transition), and add deterministic interleaving tests for wait and view/read.

Validation: full Go suite passed. `go test -race -count=5` exposed one existing-looking flaky `TestBootstrapConvergenceClassifiesCandidatesAndSeedsBus` failure (other repetitions/packages passed; no race report). Web ran 33 suites/526 tests but `store.test.ts` could not resolve the unbuilt `@gmux/protocol` package in this workspace, so the claimed web gate needs rerun via the repository build/moon prerequisite. I independently verified OSC 133 tracking exists on main (`packages/adapter/promptmarks.go`, wired by `cli/gmux/internal/ptyserver/ptyserver.go`), so the process-active premise is real.

mgabor3141 (commented):
Round 2 — FUNDAMENTAL-FLAW verdict (current PR)

No new PR commit was pushed after Round 1, so the unconditional acknowledgement race remains unchanged at head `07e7753257b6b16f6959446637cabbf0471f40cd`. The implementation session exhausted its context without returning a response or update, and a follow-up prompt fails immediately with a context-window error. I therefore cannot truthfully advance this PR to MERGE.

The blocker is fundamental to the current consumption protocol, not a cosmetic omission: `/read` identifies only session/runner incarnation, while correctness requires identifying the observed outcome (or an equivalent atomic/versioned boundary). As shipped, an acknowledgement for result N can clear result N+1. This needs protocol/state design work and deterministic completion-vs-wait/view tests before a fresh review; additive wire compatibility must be preserved.

mgabor3141:
Delta review of `9d073aa4`: **AMEND**.

**High — a valid token acknowledgement is silently dropped during the runner-replacement fence.** `acknowledgeSession` sees no current runtime and delegates to `AcknowledgeDeadToken` (`services/gmuxd/cmd/gmuxd/consume_central.go:16-20`). But while replacement is in its commit-to-install window, the registry deliberately makes the superseded live entry absent; `acknowledgeDead` detects `fenced(id)` and returns `nil` without acknowledging either owner (`services/gmuxd/internal/sessioncoord/acknowledge.go:43-47`). The route then cancels notification and returns success (`services/gmuxd/cmd/gmuxd/serve_central.go:1138-1150`), while the consumed result remains unread.

I reproduced this deterministically with a throwaway probe: install generation 1, supersede it to establish the fence, call `AcknowledgeDeadToken(id, token)`, and assert it must not report success without a write. It returned nil and performed no durable acknowledgement. The probe was deleted.

Please serialize/retry the acknowledgement through commit-to-install, then target the installed live generation or retained dead row with both incarnation and token conditions. Add a route-level deterministic fence test proving a valid acknowledgement clears its observed result before cancellation.

Dominance result: token checks correctly dominate unconditional `/read` for N-observed/N+1-completes in steady live and retained-dead ownership, including wait, tail, agent logs, web view, explicit read, and concurrent duplicate consumers. They do **not** dominate the old liveness expectation across ownership handoff: an accepted read boundary must not become a successful no-op. Focused Go schedules passed 10x; web `store.test.ts` passed 140 tests after building protocol locally; generated files were removed.

mgabor3141:
# PR 472 final adversarial delta review — 2026-08-07 — AMEND

Head reviewed: `6f630912a5d14b2b4f0d649dd7859696d41f1c85`.

## Blocking finding

**High — post-I/O ownership validation can accept the wrong generation and return false success.** `acknowledgeSession` captures a `Runtime` containing the coordinator generation, but after runner `/read` it defines `sameOwner` using only incarnation and endpoint (`services/gmuxd/cmd/gmuxd/consume_central.go:37-41`). A replacement is a new coordinator owner even if it reuses the same endpoint/incarnation values. Under the deterministic schedule old generation N receives and completes `/read`, generation N+1 installs with the same endpoint/incarnation while that response is in flight, and then validation observes N+1, the function misclassifies N+1 as N, returns success, and the route cancels delivery without acknowledging N+1.

Deterministic throwaway race probe replaced `acknowledgementRuntime` with the exact owner sequence `{generation:1, endpoint:E, incarnation:I}` then `{generation:2, endpoint:E, incarnation:I}`, served successful `/read` replies on E, and required a retry against generation 2. It failed deterministically under `-race`: only one runner request occurred before nil success (`requests=1 resolutions=2`). Probe deleted.

Required amendment: include `current.Generation == runtime.Generation` in post-I/O owner identity and add this same-endpoint/same-incarnation replacement schedule as a permanent test. Generation is the coordinator's actual ownership token; endpoint/incarnation alone do not prove the installed registry entry survived the I/O.

## Re-run and path audit

- Delayed acknowledgement N versus completion N+1: retained route and store token rejection repeated 10x under `-race`; stale token preserved N+1 unread/token and pending delivery.
- Commit-to-install fence: route, coordinator fence, and live/fenced token-owner tests repeated 10–20x under `-race`; owner resolution waited for install, valid owner acknowledgement completed before notification cancellation.
- Retained/dead and restart durability tests repeated 10x under `-race`; passed.
- The lifecycle-mutex wait has no new lock-order cycle in the exercised fence schedule; bounded ownership retries return an error rather than false success after churn, and route cancellation occurs only on nil/not-found. Token mismatch remains a conflict and does not cancel.
- The generation omission above defeats those guarantees in the endpoint/incarnation reuse case and remains merge-blocking.

Working copy is clean; no production edits retained.

mgabor3141:
# PR 472 final adversarial delta re-review — 2026-08-07 — MERGE

Head reviewed: `3685f3b335e5be3917cd892031b41051a4bb588b`.
Delta reviewed: generation-identity amendment after `6f630912` only.

## Resolution

The prior blocking finding is fixed. `acknowledgeSession` now requires equality of coordinator `Generation` as well as endpoint and incarnation before accepting the post-I/O owner as unchanged (`services/gmuxd/cmd/gmuxd/consume_central.go:38-42`). Therefore a successful `/read` against displaced generation N cannot be reported as consumption by installed generation N+1 merely because N+1 reused the same endpoint and incarnation.

The added permanent test reproduces the exact identity collision and proves two `/read` attempts plus all four owner resolutions (`services/gmuxd/cmd/gmuxd/consume_central_test.go:14-69`).

## Deterministic verification

- Re-ran the permanent same-endpoint/same-incarnation generation-replacement schedule 20x under `-race`: pass; generation mismatch forced retry to N+1.
- Ran a throwaway route-level probe 20x under `-race`: N's response completed, owner resolution changed from generation 1 to 2 with identical endpoint/incarnation, and N+1's `/read` was held. While held, the route had not returned and pending notification remained present. After releasing N+1, the route returned success, canceled notification, and exactly two runner acknowledgements had occurred.
- Re-ran the commit-to-install fence route schedule alongside these probes 20x under `-race`: pass.
- No deadlock or false success occurred; the retry reached the replacement owner, no acknowledgement was attributed to the wrong generation, and cancellation happened only after the valid replacement acknowledgement completed.

Throwaway probe deleted. Working copy is clean; no production edits retained.

PR #477, opened 2026-08-09
feat(peering): stream bounded atomic session bootstrap

mgabor3141:
## Summary

- replace unbounded protocol-3 session snapshots with 48 KiB begin/batch/ready transactions of complete rows
- quarantine only an oversized row, emit a bounded safe diagnostic, and complete ready for the remaining projection
- show persistent omitted-session warnings in the sidebar until a later clean bootstrap
- stage browser/peer rows atomically, enforce strictly increasing epochs, and lock each transport to its first accepted session protocol
- explicitly negotiate protocol 3 for current browsers and peers while retaining unversioned protocol 2 for old tabs/custom consumers for one release
- implement standard multiline SSE parsing with bounded line and aggregate memory

This removes the oversized initial/full-list **session** frame. It does not implement archive/live-set selection or change `snapshot.world` framing.

## Protocol and consistency

A protocol-3 replacement is:

1. `snapshot.sessions.begin {version:3,epoch}`
2. zero or more `snapshot.sessions.batch {epoch,sessions:[...]}` events
3. zero or more non-fatal `snapshot.sessions.error` row diagnostics
4. `snapshot.sessions.ready {epoch}`

Batches contain complete rows and are at most 48 KiB JSON. Receivers publish only at matching ready. `fanout.Subscribe` captures/install-subscribes under the publication mutex, so a racing mutation is either in baseline N or queued after N's ready.

Epochs increase strictly per transport. The first accepted session event locks that transport to legacy or protocol 3: protocol-3 -> injected legacy -> stale protocol-3 cannot roll back, and protocol 3 cannot take over a legacy-only connection. Disconnect releases staging and resets mode/epoch history. One 10-second deadline covers the complete transaction, with cancellation checks between writes.

## Availability, diagnostics, and bounds

A >48 KiB row is omitted rather than split or allowed to brick every reconnect. Its diagnostic uses a fixed reason and safe ID (long IDs become SHA-256 identities); ready commits every other row. This includes the old-spoke-large-row -> new-hub -> browser/downstream schedule.

The browser displays a persistent sidebar warning with an exact omitted total and at most 256 safe ID/reason details. A counted summary cannot be dropped by the detail cap (300 omitted reports 300 while rendering 256 identifiers). A later clean bootstrap clears both. Peers retain bounded diagnostics for the transaction and log bounded details/count. Diagnostics never become fake session rows and never prevent ready.

- session event: 48 KiB maximum
- sender/receiver transaction: 100,000 rows / 64 MiB, with 1 MiB sender envelope reserve
- diagnostics: 256 individual plus one counted summary
- Go SSE line and accumulated event: 1 MiB transitional ceiling

Malformed/overflow transactions retain prior visible state and recover at the next newer begin. Session rows contain no transcripts/scrollback.

## Transitional compatibility

- current browser: `/v1/events?session_stream=3`
- current peer: `/v1/events?as=peer&session_stream=3`
- old tab/custom non-peer consumer without a marker: legacy `snapshot.sessions`
- old peer without marker: legacy `snapshot.sessions`
- new hub -> old spoke: legacy response accepted
- unknown marker: legacy fallback, never guessed

The new browser also retains a defensive legacy listener. Protocol-2 old-browser behavior remains unchanged: an unversioned connection receives `snapshot.sessions` followed by the ordinary `snapshot.world` event. The new browser's mode lock prevents mixed-mode replay.

## SSE and world scope

The Go SSE parser accumulates all `data:` fields, joins with newline, and dispatches once at the blank boundary; field ordering, comments, default `message` events, line limits, and aggregate limits are tested.

`snapshot.world` is deliberately unchanged. This PR makes no endpoint-wide boundedness claim: 48 KiB applies only to protocol-3 session events. A realistic 1,000-membership/50-project/20-peer fixture remains as a **26,177 B characterization only**, not a maximum. World semantic framing is a separate future design if production evidence demonstrates it.

## Measurements (1,000 realistic session rows)

| metric | old | bounded stream |
|---|---:|---:|
| max session JSON event | 687,678 B | 48,889 B |
| total session SSE bytes | 687,711 B | 688,721 B |
| events | 1 | 17 |
| median serialization (3 runs) | ~1.30 ms | ~1.77 ms |

Wire overhead is 1,010 B (~0.15%); local serialization adds ~0.47 ms. The pre-change reproduction generated an 860,541-byte `data:` line and `bufio.Scanner: token too long`.

Full design/evidence: `.memory/report-semantic-session-stream.md`.

## Validation

Passed locally:

- full `go test ./...` in `services/gmuxd`
- `go test -race` across sessionstream, sseclient, apiclient, peering, sessioncoord, centralstore, store, snapshot, and `cmd/gmuxd`
- actual `services/gmuxd/tools/production-e2e.sh` container gate
- gmux-web full suite: 667 tests; typecheck and production build
- protocol build/tests
- website generated-reference + Astro build
- Biome lint, `go vet ./...`, centralstore generated check and CGO-free cross-build
- focused mixed-mode rollback schedules in peer and browser
- protocol-2 legacy-only sessions + unchanged world sequence
- persistent exact diagnostic total above the 256-detail cap and clean-bootstrap clearance
- oversized-row amplification, stale epochs, disconnect release, multiline SSE, aggregate parser bound, and transaction deadline schedules

All CI checks passed on the prior remediation head; this head reruns the same required workflow set.

## Open tradeoffs

- mutations still re-enumerate the full selected set; no archive/live-set, delta, or replay work is claimed
- quarantined rows remain available through CLI/REST but are absent from SSE until they shrink; the UI now makes that degradation explicit
- transitional protocol-2 sessions remain single frames for one release, bounded at 1 MiB on new Go clients
- world framing and limits are intentionally unchanged and require a separate design if real production evidence demonstrates need

### DIFF
--- a/.memory/report-semantic-session-stream.md
+++ b/.memory/report-semantic-session-stream.md
@@ -19,8 +19,9 @@ framing. This does not implement archive/live-set selection.
 A row larger than 48 KiB no longer aborts or closes the stream. The sender omits
 only that row, emits a bounded diagnostic with a fixed reason and safe identity
 (IDs over 128 bytes become `sha256:<digest>`), and completes `ready` for every
-other row. The browser publishes a persistent omitted-session count with safe
-ID/reason details; a later clean bootstrap clears it. Peers retain each bounded
+other row. The browser publishes a persistent exact omitted-session total with
+at most 256 safe ID/reason details; the sender's counted summary cannot consume
+or be dropped by the detail cap, and a later clean bootstrap clears both. Peers retain each bounded
 diagnostic for the transaction and log it. This handles the confirmed old-spoke
 amplification schedule: a new hub accepts a legacy row up to its 1 MiB
 compatibility ceiling, then quarantines that row rather than bricking its
--- a/.memory/report-semantic-session-stream.md
+++ b/.memory/report-semantic-session-stream.md
@@ -19,9 +19,12 @@ framing. This does not implement archive/live-set selection.
 A row larger than 48 KiB no longer aborts or closes the stream. The sender omits
 only that row, emits a bounded diagnostic with a fixed reason and safe identity
 (IDs over 128 bytes become `sha256:<digest>`), and completes `ready` for every
-other row. This handles the confirmed old-spoke amplification schedule: a new
-hub accepts a legacy row up to its 1 MiB compatibility ceiling, then quarantines
-that row rather than bricking its browser or downstream peer projection.
+other row. The browser publishes a persistent omitted-session count with safe
+ID/reason details; a later clean bootstrap clears it. Peers retain each bounded
+diagnostic for the transaction and log it. This handles the confirmed old-spoke
+amplification schedule: a new hub accepts a legacy row up to its 1 MiB
+compatibility ceiling, then quarantines that row rather than bricking its
+browser or downstream peer projection.
 
 Likely row-size causes are `command`, `cwd`, `remotes`, `title`, `subtitle`,
 `socket_path`, and `conversation_file`. Rows contain no scrollback/transcript.
@@ -47,9 +50,12 @@ queued as a later full replacement. The handler serializes all of N through
 `ready` before reading N+1.
 
 Epochs must increase strictly within one transport in both browser and peer
-receivers. Duplicate/older begin events are ignored without destroying a newer
-in-flight transaction; stale batch/ready cannot publish. Disconnect releases
-staging and resets epoch history, allowing the next transport to restart at 1.
+receivers. The first accepted session event also locks the transport to legacy
+or protocol 3, so a legacy injection cannot reset v3 replay protection and v3
+cannot take over a legacy-only connection. Duplicate/older begin events are
+ignored without destroying a newer in-flight transaction; stale batch/ready
+cannot publish. Disconnect releases staging and resets mode/epoch history,
+allowing the next transport to restart at 1.
 Browser tests prove release by sending batch+ready without a new begin after an
 error; the partial rows do not publish.
 
@@ -83,24 +89,14 @@ the accumulated event remain bounded.
 
 ## `snapshot.world` scope
 
-World remains a single semantic object, separate from session rows. It now has
-an explicit **512 KiB JSON sender maximum**, below the Go transport's 1 MiB
-ceiling. Oversize/encode failure emits a small `snapshot.world.error`; no
-unbounded world line is written. Browser updates retain the previous world; an
-initial error exposes safe empty defaults so session availability is preserved.
-
-The realistic transport-seam fixture includes 1,000 memberships across 50
-projects, path/remote match rules, 20 peers, health, launchers, peer projects,
-and peer discovery. It measures **26,177 bytes**, about 5% of the explicit
-maximum. The earlier synthetic ID-only fixture was rejected as insufficient.
-
-I did not semantically fragment world: at gmux's documented supported scale
-(single user, dozens of sessions/peers; ADR 0001 explicitly says the snapshot
-model is unsuitable for thousands), the composed stress fixture is 20× below
-the sender maximum. Arbitrarily large operator metadata is now rejected before
-the transport rather than producing an unbounded SSE line. Claims are narrowed:
-48 KiB applies to protocol-3 session events; world has its separate 512 KiB
-maximum; transitional protocol-2 sessions retain the 1 MiB client ceiling.
+World remains the pre-existing single event, separate from session rows. This PR
+adds no world cap/error behavior and makes no endpoint-wide boundedness claim:
+48 KiB applies only to protocol-3 session events. This preserves protocol-2 old
+tabs and deep-link hydration exactly. The realistic shape fixture (1,000
+memberships across 50 projects, match rules, 20 peers, health, launchers, peer
+projects, and discovery) remains only a size characterization (**26,177
+bytes**), not a supported maximum. World semantic framing is a separate future
+design if production evidence demonstrates it.
 
 ## Reproduction and measurements
 
@@ -131,14 +127,14 @@ diagnostic cap) and does not affect healthy fixtures.
   `internal/sessionstream/sessionstream_test.go`
 - standard multiline SSE parsing and line/aggregate limits:
   `internal/sseclient/client_test.go`
-- peer reconnect, diagnostics-through-ready, overflow recovery, monotonic epoch,
-  old-spoke-large-row amplification:
+- peer reconnect, retained diagnostics-through-ready, overflow recovery,
+  monotonic epoch, mixed-mode rollback, old-spoke-large-row amplification:
   `internal/peering/session_stream_test.go`
-- browser explicit negotiation, legacy fallback, real disconnect release,
-  monotonic epoch, degraded ready:
+- browser explicit negotiation, legacy-only fallback, protocol-mode lock, real
+  disconnect release, monotonic epoch, persistent degraded-ready warning:
   `apps/gmux-web/src/sse-reconnect.test.ts`
-- subscription boundary, compatibility matrix, realistic world transport bound,
-  oversized-world diagnostic:
+- subscription boundary, compatibility matrix, realistic world-size
+  characterization:
   `cmd/gmuxd/session_stream_boundary_test.go`
 - transaction-wide deadline/cancellation:
   `cmd/gmuxd/sse_transaction_test.go`
@@ -153,5 +149,5 @@ diagnostic cap) and does not affect healthy fixtures.
   remain inspectable/actionable through one-shot CLI/REST paths.
 - Transitional protocol-2 session frames are still single events for one
   release; the new client bounds them at 1 MiB.
-- World is bounded rejection rather than fragmentation. If supported deployment
-  scale grows near 512 KiB, it should gain its own semantic row transaction.
+- World framing and limits are deliberately unchanged and require a separate
+  design if real-world evidence demonstrates an oversized world path.
--- a/.memory/report-semantic-session-stream.md
+++ b/.memory/report-semantic-session-stream.md
@@ -1,141 +1,157 @@
-# Bounded semantic session stream
+# Bounded semantic session stream — amended after review
 
-## Summary
+## Protocol
 
-Protocol 3 replaces each unbounded `snapshot.sessions` SSE event with a
-transaction of complete semantic rows:
+A protocol-3 session replacement is one transaction:
 
 1. `snapshot.sessions.begin {version:3,epoch}`
-2. zero or more `snapshot.sessions.batch {epoch,sessions:[...]}`
-3. `snapshot.sessions.ready {epoch}`
+2. zero or more `snapshot.sessions.batch {epoch,sessions:[...]}` events
+3. zero or more non-fatal `snapshot.sessions.error` row diagnostics
+4. `snapshot.sessions.ready {epoch}`
 
-Receivers stage batches privately and replace the visible projection only at
-`ready`. This applies to initial hydration and the existing coalesced full-list
-replacements after mutations. It deliberately does not select an archive or a
-bounded live set; a future implementation can change which rows are enumerated
-without changing framing.
+Batches contain complete semantic rows and have a 48 KiB JSON payload maximum.
+Receivers stage privately and replace the visible set only at the matching
+`ready`. Initial hydration and later coalesced full replacements use the same
+framing. This does not implement archive/live-set selection.
 
-## Bounds and failure behavior
+## Availability and bounds
 
-- Maximum JSON payload per session-stream event: **48 KiB (49,152 bytes)**.
-  This leaves 16 KiB below the default 64 KiB Scanner token for SSE syntax,
-  envelope growth, and intermediary metadata.
-- Rows are JSON-encoded independently and never byte-split.
-- One row which cannot fit causes encoding to fail before `begin` is sent. The
-  server emits a bounded `snapshot.sessions.error`, logs the row ID, and closes
-  that response. The receiver discards staging and retains the previous ready
-  set. Likely unbounded fields are `command`, `cwd`, `remotes`, `title`,
-  `subtitle`, `socket_path`, and `conversation_file`. Session rows contain no
-  scrollback or transcript content.
-- Receiver staging is capped at 100,000 rows and 64 MiB.
-- The Go SSE Scanner line ceiling is bounded at 1 MiB, up from 256 KiB. This
-  admits the measured 1,000-row protocol-2 frame during mixed-version operation.
-  Protocol 3 does not rely on that increase. Lines above it return a protocol
-  error before unbounded growth.
+A row larger than 48 KiB no longer aborts or closes the stream. The sender omits
+only that row, emits a bounded diagnostic with a fixed reason and safe identity
+(IDs over 128 bytes become `sha256:<digest>`), and completes `ready` for every
+other row. This handles the confirmed old-spoke amplification schedule: a new
+hub accepts a legacy row up to its 1 MiB compatibility ceiling, then quarantines
+that row rather than bricking its browser or downstream peer projection.
 
-## Consistency boundary
+Likely row-size causes are `command`, `cwd`, `remotes`, `title`, `subtitle`,
+`socket_path`, and `conversation_file`. Rows contain no scrollback/transcript.
+They are never truncated or arbitrary-byte-split.
 
-`sseFanout.Subscribe` takes the same mutex as `BroadcastFrames`, installs the
-subscriber, and copies the current full projection before releasing it. A
-mutation therefore falls on exactly one side of the boundary:
+Sender and receiver use symmetric transaction limits:
 
-- before capture: reflected by baseline epoch N; or
-- after capture: queued as a later full replacement epoch N+1.
+- 100,000 staged rows
+- 64 MiB receiver encoded staging
+- sender accepts at most 63 MiB of row JSON, reserving 1 MiB for batch envelopes
+- at most 256 individual diagnostics, followed by one bounded summary
+- Go SSE line/event accumulation ceiling: 1 MiB for transitional protocol 2
 
-The HTTP handler serially writes all events for one epoch, including `ready`,
-before reading the next fanout message. A concurrent mutation cannot overtake
-baseline readiness. The bounded fanout may drop intermediate full replacements
-for a slow subscriber, but retains a later authoritative replacement, matching
-ADR 0001's existing latest-only semantics. Activity remains intentionally
-lossy.
+A malformed/overflow transaction is abandoned without changing visible state.
+A strictly newer begin starts cleanly, so rejection does not create permanent
+staleness. Correct senders remain inside receiver bounds.
 
-Staging belongs to one transport connection in the Go peer. Browser staging is
-cleared on EventSource error and reset by every begin marker. Disconnect before
-ready therefore cannot expose or carry partial rows into reconnect.
+## Epoch and consistency schedules
+
+`fanout.Subscribe` installs the subscriber and captures its baseline under the
+same mutex as publication. A mutation is either included in baseline epoch N or
+queued as a later full replacement. The handler serializes all of N through
+`ready` before reading N+1.
+
+Epochs must increase strictly within one transport in both browser and peer
+receivers. Duplicate/older begin events are ignored without destroying a newer
+in-flight transaction; stale batch/ready cannot publish. Disconnect releases
+staging and resets epoch history, allowing the next transport to restart at 1.
+Browser tests prove release by sending batch+ready without a new begin after an
+error; the partial rows do not publish.
+
+The complete session transaction has one 10-second write deadline. Cancellation
+is checked between fragment writes. The deadline is cleared after completion,
+so timeout no longer multiplies by event count.
 
 ## Compatibility
 
-Browser assets are daemon-served/version-locked and always receive protocol 3.
-New peer clients request `?as=peer&session_stream=3`.
+Protocol 3 is explicit for all consumers:
 
-- New hub -> old spoke: the old daemon ignores the query parameter and sends
-  legacy `snapshot.sessions`; the new hub still handles it authoritatively.
-- Old hub -> new spoke: absence of `session_stream=3` selects the legacy event.
-- New hub -> new spoke: bounded begin/batch/ready events.
-- Unknown requested versions select legacy fallback rather than guessing.
+- current browser: `/v1/events?session_stream=3`
+- current peer: `/v1/events?as=peer&session_stream=3`
+- unversioned browser tab/custom consumer: transitional protocol-2
+  `snapshot.sessions`
+- old peer omitting the marker: protocol 2
+- new hub to old spoke: old spoke ignores the marker; new hub accepts protocol 2
+- unknown requested version: legacy fallback rather than guessing
 
-Thus mixed peers do not silently ignore/misparse new event names. The legacy
-fallback remains intrinsically unbounded and can still exceed an old peer's
-Scanner limit for very large sets; retaining wire compatibility and fixing an
-old binary's parser cannot both be guaranteed by the new spoke.
+The new browser also retains a legacy listener defensively. An old tab opened
+across a daemon upgrade requests no marker and therefore continues receiving the
+event it understands rather than silently freezing.
 
-## Oversized-path inspection
+## SSE parser
 
-`snapshot.sessions` and `snapshot.world` are distinct frames. The actual large
-row projection is `snapshot.sessions`; peer subscriptions suppress world
-entirely. `snapshot.world` carries project membership IDs but no session rows.
-With 1,000 eight-character IDs in one project its encoded payload measured
-**11,110 bytes**, below the session event budget. World framing was therefore
-left unchanged.
+The Go SSE client now follows event boundaries: all `data:` fields are
+accumulated, joined with `\n`, and dispatched once at the blank line. Field order
+is unconstrained, no-event blocks use type `message`, comments do not terminate
+an event, and empty/no-data blocks do not dispatch. Both individual lines and
+the accumulated event remain bounded.
+
+## `snapshot.world` scope
+
+World remains a single semantic object, separate from session rows. It now has
+an explicit **512 KiB JSON sender maximum**, below the Go transport's 1 MiB
+ceiling. Oversize/encode failure emits a small `snapshot.world.error`; no
+unbounded world line is written. Browser updates retain the previous world; an
+initial error exposes safe empty defaults so session availability is preserved.
+
+The realistic transport-seam fixture includes 1,000 memberships across 50
+projects, path/remote match rules, 20 peers, health, launchers, peer projects,
+and peer discovery. It measures **26,177 bytes**, about 5% of the explicit
+maximum. The earlier synthetic ID-only fixture was rejected as insufficient.
+
+I did not semantically fragment world: at gmux's documented supported scale
+(single user, dozens of sessions/peers; ADR 0001 explicitly says the snapshot
+model is unsuitable for thousands), the composed stress fixture is 20× below
+the sender maximum. Arbitrarily large operator metadata is now rejected before
+the transport rather than producing an unbounded SSE line. Claims are narrowed:
+48 KiB applies to protocol-3 session events; world has its separate 512 KiB
+maximum; transitional protocol-2 sessions retain the 1 MiB client ceiling.
 
 ## Reproduction and measurements
 
-Before code changes, a realistic 1,000-row projection (commands, cwd/workspace,
-remotes, title/subtitle, socket path, conversation reference, runner/project
-metadata; no transcripts) produced:
+Pre-change realistic fixture:
 
 ```
 legacy payload=860535 frame=860568 max_line=860541 events=1
 lines=1 err=bufio.Scanner: token too long
 ```
 
-The checked-in deterministic Go fixture (slightly narrower but still realistic)
-measures old versus protocol 3:
+Checked-in 1,000-row fixture:
 
-| metric | old full event | protocol 3 |
+| metric | old | protocol 3 |
 |---|---:|---:|
-| rows | 1,000 | 1,000 |
-| maximum JSON event payload | 687,678 B | 48,889 B |
-| total SSE bytes | 687,711 B | 688,721 B |
-| event count | 1 | 17 |
-| serialization latency (median of 5 benchmark runs) | ~1.47 ms | ~2.71 ms |
+| max session JSON event | 687,678 B | 48,889 B |
+| total session SSE bytes | 687,711 B | 688,721 B |
+| session event count | 1 | 17 |
+| median serialization, 3 runs | ~1.30 ms | ~1.77 ms |
 
-Total wire overhead is 1,010 bytes (~0.15%). Serialization adds ~1.24 ms on the
-fixture. On a local/in-memory initial synchronization this is the measured
-latency delta; real links are dominated by transferring the same ~689 KiB, while
-the receiver intentionally exposes the set at the final ready marker rather
-than incrementally. Benchmark command:
-
-```
-go test -run '^$' -bench BenchmarkInitialSync1000 -benchmem -count=5 ./internal/sessionstream
-```
+Wire overhead is 1,010 bytes (~0.15%). The measured local serialization delta is
+~0.47 ms. Quarantine adds one bounded diagnostic per bad row (subject to the
+diagnostic cap) and does not affect healthy fixtures.
 
 ## Evidence map
 
-- Legacy >64 KiB reproduction, empty/one/many rows, event bounds,
-  oversized-single-row failure, byte/event measurements:
+- bounds, empty/one/many, legacy Scanner reproduction, quarantine, safe IDs,
+  sender/receiver limit symmetry, measurements:
   `internal/sessionstream/sessionstream_test.go`
-- Scanner compatibility acceptance and hard ceiling:
+- standard multiline SSE parsing and line/aggregate limits:
   `internal/sseclient/client_test.go`
-- Browser disconnect/reconnect and mutation staging:
-  `apps/gmux-web/src/sse-reconnect.test.ts`
-- Peer disconnect, reconnect, atomic readiness, mutation epoch, legacy spoke:
+- peer reconnect, diagnostics-through-ready, overflow recovery, monotonic epoch,
+  old-spoke-large-row amplification:
   `internal/peering/session_stream_test.go`
-- Subscribe/publication boundary, browser/peer version selection, world size:
+- browser explicit negotiation, legacy fallback, real disconnect release,
+  monotonic epoch, degraded ready:
+  `apps/gmux-web/src/sse-reconnect.test.ts`
+- subscription boundary, compatibility matrix, realistic world transport bound,
+  oversized-world diagnostic:
   `cmd/gmuxd/session_stream_boundary_test.go`
-- Real server event order and matched world:
-  `cmd/gmuxd/serve_switch_integration_test.go`
+- transaction-wide deadline/cancellation:
+  `cmd/gmuxd/sse_transaction_test.go`
+- production initial event contract:
+  `cmd/gmuxd/production_container_e2e_test.go`
 
 ## Open tradeoffs
 
-- Full session sets are still re-enumerated after mutations. This PR bounds
-  transport frames and atomic application; it does not implement deltas,
-  archive/live-set selection, resumable epochs, or server-side replay.
-- A single semantic row over 48 KiB is rejected rather than split. A future
-  protocol could model specific large fields semantically, but arbitrary-byte
-  checkpointing is intentionally avoided.
-- `snapshot.world` remains one event. The measured many-session path is small;
-  independently unbounded project/peer metadata may warrant its own semantic
-  framing if real deployments approach the Scanner ceiling.
-- Browser EventSource provides no custom negotiation headers; this is safe only
-  because the browser bundle and daemon are served/versioned together.
+- Every mutation still re-enumerates the full selected set; no delta/replay or
+  archive/live-set policy is introduced.
+- Quarantined rows are absent from the SSE projection until they shrink. They
+  remain inspectable/actionable through one-shot CLI/REST paths.
+- Transitional protocol-2 session frames are still single events for one
+  release; the new client bounds them at 1 MiB.
+- World is bounded rejection rather than fragmentation. If supported deployment
+  scale grows near 512 KiB, it should gain its own semantic row transaction.
--- a/.memory/report-semantic-session-stream.md
+++ b/.memory/report-semantic-session-stream.md
@@ -1,0 +1,141 @@
+# Bounded semantic session stream
+
+## Summary
+
+Protocol 3 replaces each unbounded `snapshot.sessions` SSE event with a
+transaction of complete semantic rows:
+
+1. `snapshot.sessions.begin {version:3,epoch}`
+2. zero or more `snapshot.sessions.batch {epoch,sessions:[...]}`
+3. `snapshot.sessions.ready {epoch}`
+
+Receivers stage batches privately and replace the visible projection only at
+`ready`. This applies to initial hydration and the existing coalesced full-list
+replacements after mutations. It deliberately does not select an archive or a
+bounded live set; a future implementation can change which rows are enumerated
+without changing framing.
+
+## Bounds and failure behavior
+
+- Maximum JSON payload per session-stream event: **48 KiB (49,152 bytes)**.
+  This leaves 16 KiB below the default 64 KiB Scanner token for SSE syntax,
+  envelope growth, and intermediary metadata.
+- Rows are JSON-encoded independently and never byte-split.
+- One row which cannot fit causes encoding to fail before `begin` is sent. The
+  server emits a bounded `snapshot.sessions.error`, logs the row ID, and closes
+  that response. The receiver discards staging and retains the previous ready
+  set. Likely unbounded fields are `command`, `cwd`, `remotes`, `title`,
+  `subtitle`, `socket_path`, and `conversation_file`. Session rows contain no
+  scrollback or transcript content.
+- Receiver staging is capped at 100,000 rows and 64 MiB.
+- The Go SSE Scanner line ceiling is bounded at 1 MiB, up from 256 KiB. This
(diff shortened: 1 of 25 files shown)


mgabor3141:
Advisory review of exact head `b58e1590eff465d056bbb6dcb9506fd2e7440b5b`: **AMEND**.

**P1 — `snapshot.world` remains an unbounded SSE line.** `services/gmuxd/cmd/gmuxd/serve_central.go:921-925` sends it via `sendSSEFrame`; `services/gmuxd/cmd/gmuxd/main.go:747-753` has no payload/line bound. The checked-in `session_stream_boundary_test.go:32-48` only measures a synthetic world with 1,000 short IDs, not actual composed projects/match rules/peers/health/launchers. A sufficiently large world can still hit intermediary/client line limits (and the same unbounded-line failure mode for programmatic consumers), even though session batches are bounded.

**P2 — Go SSE parsing is not standards-compliant for multiline data.** `services/gmuxd/internal/sseclient/client.go:282-317` invokes the handler per `data:` line instead of accumulating data and dispatching once at the blank line with `\\n` joining, despite its API comment at lines 54-56 claiming concatenation. `event: x\\ndata: one\\ndata: two\\n\\n` is delivered as two callbacks, not one `one\\ntwo` event. Current senders use one data line, but the compatibility claim is false; tests cover only single-line data.

The core session-fragment mechanism otherwise matches the requested model: complete rows, bounded batches, atomic ready, empty snapshots, and connection-local staging. Peer negotiation has explicit `session_stream=3` with legacy fallback; browser protocol 3 is version-lock-by-deployment, not negotiated. The checked-in 1,000-row benchmark is only a JSON microbenchmark (discarded output; no actual write/flush/parse/allocation measurement or recorded median), so the `.memory` benchmark claims remain unverified. Please fix the SSE state machine and bound/frame world (or narrow the guarantee and test the real composed world) before merge.

mgabor3141 (commented):
# Independent review: gmuxapp/gmux PR #477

Revision: `b58e1590eff465d056bbb6dcb9506fd2e7440b5b`  
Focus: concurrency, state machine, failure paths. No subagents.

## Finding

### MEDIUM — same-connection duplicate/older epoch can roll back the peer projection

`services/gmuxd/internal/peering/peer.go:483-517` accepts every valid `begin`, including an already committed or older epoch, and retains no last-committed epoch. Therefore `begin(1), batch(old), ready(1), begin(2), batch(new), ready(2), begin(1), batch(old), ready(1)` commits `old` after `new`. A malformed/replayed authenticated peer stream can silently restore dismissed sessions or stale status until a later snapshot.

Deterministic probe under `-race`:

```text
go test -race ./services/gmuxd/internal/peering \
  -run TestReview477DuplicateOldEpochRollsBackProjection -count=1
--- FAIL: TestReview477DuplicateOldEpochRollsBackProjection
projection rolled back: [{ID:old@spoke ...}]
```

The temporary test directly drove `handleStreamEvent` with epochs 1, 2, then 1 and asserted `new@spoke` remained; actual was `old@spoke`. Probe deleted.

Suggested mechanism: track accepted/committed epoch for the life of a transport connection and reject non-increasing begin epochs (or fail/reconnect the stream on protocol violations). Reset monotonic state only with the connection-local bootstrap on reconnect. Browser state should enforce the same ordering.

## State-transition table

| Input/schedule | Sender | Receiver staging | Visible state |
|---|---|---|---|
| subscribe vs snapshot | `fanout.Subscribe` registers and copies initial under one mutex | none → begin(E) | initial cannot miss a concurrent broadcast; queued full replacements follow |
| begin(E) | emits after complete `Encode` succeeds | replaces any staging with active E | unchanged |
| batch(E) | ordered synchronous writes | matching active E appends; mismatch clears | unchanged |
| ready(E) | follows all E batches | matching E detaches rows and clears stage | atomic replacement |
| mutation during send | full snapshot queued in subscriber channel | next epoch stages after prior ready | newest replacement eventually visible; queue overflow retains a later full snapshot |
| disconnect/cancel | write/subscribe exits | Go connection-local stage dies; browser error clears global stage | manager disconnect prunes peer projection; browser retains prior visible projection |
| reconnect | sender epoch counter restarts at 1 | fresh connection-local stage | fresh ready replaces/pruned projection |
| peer replacement | manager cancels and waits for old generation before cleanup/new generation | old stage becomes unreachable | old generation cannot commit after wait returns |
| malformed begin/batch/ready | sender normally cannot emit | clears stage, stream continues | unchanged |
| stale ready alone | n/a | mismatch clears the current stage | unchanged, but valid in-flight epoch is abandoned |
| duplicate/older full epoch | not emitted by correct sender | **accepted as fresh** | **stale rows commit over newer rows** |
| legacy snapshot | new sender uses it only without protocol-3 opt-in | clears partial v3 stage, parses authoritative snapshot | immediate replacement |
| oversized row | Encode returns error before any frame is sent; handler sends bounded error then closes | error clears stage | disconnect path prunes peer projection |

## Bounds and failure-path assessment

- Scanner line allocation is bounded by `DefaultMaxLineBytes` (1 MiB), and initial scanner capacity does not exceed configured max.
- Protocol event payloads produced by `Encode` are <=48 KiB, but a row is fully JSON-marshaled before its size is checked, and `Encode` materializes the entire event list before sending; there is no sender aggregate/count bound.
- Go receiver staging rows/bytes are checked before appending to persistent staging, but `json.Unmarshal` allocates the current batch before those aggregate checks. Scanner still caps that transient batch input to 1 MiB.
- Browser `EventSource`/`JSON.parse` allocate event and batch before row/aggregate checks; there is no client-side event-size enforcement. Correct same-origin sender output is bounded, but the constants are not pre-allocation defenses against a faulty endpoint.
- Partial staging did not leak to the sink. Clearing or replacing the bootstrap releases references; no abandoned-bootstrap retention was found in the reviewed ownership paths.

## Test vacuity

`TestPeerMutationEpochAppliesAfterPriorReadyExactlyOnce` does not exercise mutation capture, fanout queueing, or concurrent send: its `emit` closure directly calls receiver handlers sequentially. It supports receiver atomicity only, not the subscribe-vs-snapshot/queued-mutation claim in its comment. Existing fanout locking was traced separately.

## Commands

- `go test -race ./services/gmuxd/internal/sessionstream ./services/gmuxd/internal/sseclient ./services/gmuxd/internal/peering ./services/gmuxd/cmd/gmuxd`
- deterministic duplicate-epoch probe above
- `jj status`, `jj diff --stat`

## Verdict

**AMEND** — reject non-increasing epochs on each live transport (and add a non-vacuous regression covering rollback). The normal sender/receiver schedule is atomic and race-clean, but the receiver state machine currently permits stale authoritative rollback.

mgabor3141 (commented):
Adversarial anchor review at `b58e1590eff465d056bbb6dcb9506fd2e7440b5b` (advisory; not a formal approval/request-changes). Full report: `.memory/review-477-opus.md`.

The framing, epoch discipline and atomicity core are sound — I could not construct a partial-visibility, duplicate/out-of-order, cross-reconnect-staging, or mixed-version misparse failure. Two findings should block, though.

## H1 (High, reproduced) — one session row >48 KiB permanently disables the entire session stream for every protocol-3 subscriber

- `internal/sessionstream/sessionstream.go:96-99` — an unfittable row aborts `Encode` for the **whole** replacement.
- `cmd/gmuxd/serve_central.go:854-862` — the handler emits `snapshot.sessions.error` and then **returns the error**; `:870` / `:910` `return`, closing the SSE response.
- `apps/gmux-web/src/store.ts:1876-1880` — browser discards staging, `connState` → `error`/`reconnecting`. EventSource reconnects, hits the same row, forever.

Reproduction (throwaway probe in `cmd/gmuxd`, modeled on `serve_switch_integration_test.go`, since deleted): fake runner reporting `"command": ["python","-c", strings.Repeat("x", 60*1024)]`:

```
session stream: refusing epoch 1: session stream row too large: session "1dehpbm1" cannot fit the 49152-byte event payload limit
browser subscriber      events=[snapshot.sessions.error]   (stream closed)
peer(v3) subscriber     events=[snapshot.sessions.error]   (stream closed)
legacy peer subscriber  events=[snapshot.sessions]         (works)
/v1/sessions bytes=123266
```

Expected: a pathological row degrades that row. Actual: the browser sees **zero** sessions on every connect, with no in-UI recovery — the offending session can only be found/killed via CLI. Note the inversion: the protocol-2 fallback is the only path that still works, so this is a strict regression against what the old unbounded frame tolerated (EventSource has no line limit).

Amplification (traced): a hub accepts such a row from an **old spoke** over the legacy fallback (sseclient now tolerates 1 MiB), mirrors it, then fails to encode its own browser stream — one legacy spoke can brick a new hub's UI. And for peer subscribers the report/ADR claim "the receiver discards staging and retains the previous ready set" is false end-to-end: `Encode` error → stream close → `StatusDisconnected` → `Manager.removePeerSessions` (`internal/peering/manager.go:72-73,453`), so the previous set is pruned and the peer reconnect-loops with an empty projection.

Fix direction: skip/truncate the offending row and publish the rest, and/or do not close the connection after `snapshot.sessions.error` so the next mutation can recover.

## H2 (High) — required check is red; the removed contract is still asserted

`cmd/gmuxd/production_container_e2e_test.go:397-410` (`assertInitialSSE`) still requires the first browser SSE event to be `snapshot.sessions`; it wasn't updated with the other call sites.

```
production_container_e2e_test.go:409: first SSE="snapshot.sessions.begin"
--- FAIL: TestProductionContainerE2E/unread_restart_sse_sqlite
```
(`Central store E2E (container, fast)` → failure)

## M1 (Medium) — non-peer consumers are force-upgraded with no opt-out; protocol docs not updated

`cmd/gmuxd/main.go:663-665`: `!asPeer || requested == "3"`. A peer that omits the marker is served legacy, but a non-peer consumer is served protocol 3 even if it explicitly sends `session_stream=2`. `apps/website/src/content/docs/migrating-to-2.md:125` explicitly tells "custom consumers of the daemon SSE stream" to consume `snapshot.sessions` (also `architecture.md:108`, `develop/state-management.md:117`, `develop/session-schema.md:41`, `multi-machine.md:115` — none updated). Those consumers now silently see no session events, which is the failure mode the ADR amendment claims to prevent.

## M2 (Medium) — browser tab open across a daemon upgrade silently freezes

`store.ts` drops the `snapshot.sessions` listener entirely. "Version-locked" only holds for a fresh page load; a tab/PWA open before `gmuxd` restarted reconnects to a protocol-3 daemon, receives only `snapshot.sessions.*`, and never updates again while showing a stale list at `connState='reconnecting'`. There is no build/protocol handshake or forced reload. Cheapest mitigation: keep the legacy listener for one release.

## M3 (Medium) — the headline reconnect test is vacuous for the mechanism it names

`apps/gmux-web/src/sse-reconnect.test.ts:82-95` is the cited evidence for `discardSessionsBootstrap()` on EventSource `error` (`store.ts:1826-1828`). Deleting that call keeps the suite green:

```
# store.ts error handler: discardSessionsBootstrap() removed
npx vitest run src/sse-reconnect.test.ts   ->  ✓ 5 tests passed
```

The recovery step uses the `ready()` helper, which emits `begin` first, and `beginSessionsBootstrap` resets staging unconditionally — so the test can't distinguish "discarded on error" from "overwritten by the next begin". A real assertion would send `batch`/`ready` for an epoch *without* a preceding begin after the error.

## Low

- **L1** `sendSSEBytesFrame` (`main.go:726-733`) sets a fresh 10 s `sseWriteTimeout` **per event**, and `serve_central.go:864-868` loops over all of them. A stalled receiver used to block a handler goroutine ~10 s per replacement; at 1 000 rows it is now up to 17×10 s and scales with session count. No `r.Context().Done()` check inside the loop.
- **L2** `useSemanticSessionStream` was inserted between `shouldForwardActivity` and its 14-line doc comment (`main.go:649-667`): the comment now documents the wrong function and the new helper has none.
- **L3** Receivers cap staging at 100 000 rows (`sessionstream.go:63-65`) but `Encode` has no row cap; above that every transaction is silently discarded and, since the sender only re-sends on mutation, the receiver stays stale forever. No receiver-initiated resync exists for any discarded transaction.

## Verified sound (tried to break, failed)

Partial visibility (transaction fully written within one fanout message; activity/world are separate messages); out-of-order/duplicate/missing-begin/epoch-mismatch/malformed handling on both receivers; epoch ownership (per-connection `sessionEpoch`, per-`subscribe` staging — `Peer.run` calls `subscribe` once per connection and `sseclient.Subscribe` does not reconnect internally); `Subscribe`/`BroadcastFrames` mutex boundary and `fanoutEnqueue` drop-oldest; `Encode` byte accounting and non-aliasing batch buffers; the full mixed-version peer matrix.

Commands: `go test -race ./internal/{sessionstream,peering,sseclient,apiclient}/...` (pass), `go test ./cmd/gmuxd/ -count=1` (pass, container E2E skipped locally), `npx vitest run` (521 pass; `store.test.ts` fails locally on an unbuilt `@gmux/protocol`, unrelated). Not exercised: real two-daemon protocol-3 handshake over a network transport, Playwright locally, buffering reverse proxies, >100k rows.

**Verdict: AMEND.** H1 and H2 before merge; M1–M3 are cheap and worth folding in.

mgabor3141 (commented):
# Delta re-review — c828905dce055c29ab8cf6b72273bf85af568f61

Compared with `b58e1590eff465d056bbb6dcb9506fd2e7440b5b`. The original pure protocol-3 rollback is fixed for both Go peer and browser: each transport now rejects non-increasing epochs, a stale begin does not destroy a newer browser/peer transaction, and reconnect resets monotonic state. Exact peer probe passed under `-race`; exact browser test passed.

## Finding

### MEDIUM — legacy fallback resets protocol-3 replay protection and permits same-transport rollback

Evidence: `services/gmuxd/internal/peering/peer.go:552-555` zeroes the entire bootstrap, including `lastEpoch`, whenever a legacy `snapshot.sessions` appears. `apps/gmux-web/src/store.ts:1892-1896` likewise calls `resetSessionsTransport()`, which zeros `lastSessionEpoch`, before applying a legacy frame. Neither receiver locks the negotiated mode for the transport.

Deterministic peer and browser schedules:

1. commit protocol-3 epoch 1 (`old`)
2. commit protocol-3 epoch 2 (`new`)
3. receive a legacy authoritative snapshot (`legacy-current`)
4. replay protocol-3 epoch 1 (`replayed-old`)

Expected after step 4: the replay is rejected and the legacy projection remains. Actual in both receivers: epoch 1 is accepted as fresh and `replayed-old` replaces the projection. The Go probe failed under `-race` with `mixed-mode rollback: [{ID:replayed-old@spoke ...}]`; the Vitest probe failed with received `['replayed-old']`. Both probes were deleted.

The fallback should be accepted only before protocol-3 begins (or mode should be locked by the first session event for that transport). Only a real disconnect/reconnect should reset epoch ordering. This retains compatibility with an old endpoint while preventing a malformed, replayed, or proxy-corrupted mixed stream from disabling the new fix.

## Additional state/failure-path checks

- **Reconnect reset:** peer staging is connection-local; browser `error` calls `resetSessionsTransport`; epoch 1 is accepted on the new transport. Passed.
- **Duplicate/older epoch:** fixed in pure v3 for peer and browser. Passed exact rollback probes.
- **Malformed begin:** clears active staging but preserves peer `lastEpoch`; browser clears rows while retaining monotonic epoch. No false ready found.
- **Malformed/mismatched batch/ready:** cannot commit mismatched rows. Go abandons current staging; browser ignores mismatches. Recovery requires a newer begin. No partial visibility found.
- **Quarantine diagnostics:** valid diagnostic leaves staging alive and ready commits accepted rows. Malformed diagnostics do not publish. No retained row references found after `abandon`/reset.
- **Limits:** sender checks each accepted row/count against reachable bounded state; additions cannot approach native integer overflow because prior state is capped at 64 MiB/100k and each row is <48 KiB. Go receiver has syntactically unchecked `int` additions, but overflow is unreachable from a valid zero state for the same reason. Browser requires safe positive epochs; production `encodedBytes` is `e.data.length`, hence finite and nonnegative. No reachable integer bypass reproduced.
- **Epoch maximum:** browser safely rejects values above `Number.MAX_SAFE_INTEGER`; Go uses `uint64`. Sender wrap to zero would be rejected, but requires 2^64 snapshots on one connection and is not a practical failure path.
- **Legacy fallback alone:** accepted correctly and clears partial v3 staging. The regression is specifically mixed-mode fallback resetting replay protection.
- **False ready:** stale ready does not commit the wrong rows. Peer abandons the newer transaction; browser leaves it intact. This availability difference is conservative but should ideally be specified.

## Commands

- `go test -race ./services/gmuxd/internal/peering -run 'TestPeerRejectsNonIncreasingEpochWithoutRollback|TestPeerSessionBootstrapIsAtomicAndDisconnectLocal|TestPeerRejectsBootstrapBeyondAggregateMemoryBound|TestPeerDiagnosticDoesNotInvalidateReady|TestPeerAcceptsLegacySnapshotFromOldSpoke' -count=1`
- `pnpm --dir apps/gmux-web test -- src/sse-reconnect.test.ts` (target file passed; broader invocation also collected 33 other files, with one unrelated unresolved `@gmux/protocol` package-entry failure)
- focused `pnpm exec vitest run src/sse-reconnect.test.ts` with temporary mixed-mode probe
- focused Go `-race` mixed-mode probe
- `jj status`; `jj diff --stat`

## Verdict

**AMEND** — the original same-mode finding is fixed, but fallback must not reset epoch replay protection within an already protocol-3 transport.

mgabor3141:
Delta review at exact head `c828905dce055c29ab8cf6b72273bf85af568f61`: **AMEND**.

The prior transport findings are fixed:

- `sseclient` now accumulates `data:` fields and dispatches once at the WHATWG blank-line boundary. An exact-head probe and the checked-in multiline test passed; mutation-testing the aggregate guard made the bound test fail.
- `sendWorldSSEFrame` checks marshaled JSON before writing, with a 512 KiB cap and fixed bounded `snapshot.world.error` diagnostics. The realistic composed fixture (1,000 memberships, 50 projects/rules, 20 peers, health, launchers, peer projects/discovery) is 26,177 bytes and passed the transport-seam test. Accepted world lines remain below the 1 MiB Go scanner ceiling; supported fields cannot bypass the byte check.
- Protocol negotiation and replay guards are coherent: browser and peers explicitly request `session_stream=3`, unversioned/unknown requests fall back to legacy, and old-spoke responses remain accepted. Row quarantine preserves `ready` and complete rows remain unsplit.

**P1 — initial `snapshot.world.error` destroys deep links.** `apps/gmux-web/src/store.ts:1930-1936` sets `worldLoaded=true` on the first rejected world. The view gate at `832-839` then releases URL normalization although `projects` is still the empty safe default. `apps/gmux-web/src/routing.ts:128-133` requires a local URL's project to exist, so `/project/adapter/slug` resolves to home; the normalization effect at `store.ts:1947-1959` can rewrite it to `/`. An oversized supported field can keep emitting `world.error` indefinitely, so this is not necessarily transient. Retain `worldLoaded=false` until a valid world arrives, or suppress normalization in the initial-error state. Add a deep-link/error test.

Performance recheck of the exact fixture: legacy ~1.44–1.50 ms, 0.78 MB/3,002 allocs; semantic ~2.04–2.10 ms, 1.86 MB/5,034 allocs. Wire overhead is ~0.15%, but framing costs ~1.4x CPU, 2.4x allocated bytes, and 1.7x allocations. The benchmark remains serialization-only, so these are tradeoff evidence rather than an end-to-end claim.

mgabor3141 (commented):
Delta re-review at `c828905dce055c29ab8cf6b72273bf85af568f61` (advisory). Re-ran every reproduction from my review of `b58e159`; full report appended to `.memory/review-477-opus.md`.

## Original findings: all fixed, verified

| # | Finding | Status |
|---|---|---|
| H1 | oversized row disables the whole stream | **fixed** |
| H2 | container E2E red | **fixed** (11/11 checks green) |
| M1 | forced protocol 3 for non-peers, stale docs | **fixed** |
| M2 | old tab freezes across daemon upgrade | **fixed** |
| M3 | vacuous reconnect test | **fixed** |
| L1 | per-event write deadline | **fixed** |
| L2 | stolen doc comment | **fixed** |
| L3 | sender/receiver bound asymmetry | **fixed** |

**H1 re-run** (real daemon, two fake runners, one with a 60 KiB argv):

```
browser(v3): [begin, error {"code":"row_too_large","id":"bbbbbbbb","message":"session row
              encoded to 123185 bytes and exceeds the 49152-byte event limit; row omitted"},
              batch ids=aaaaaaaa, ready, snapshot.world bytes=1425]
peer(v3):    [begin, error {...}, batch ids=aaaaaaaa, ready]
old browser: [snapshot.sessions bytes=123493, snapshot.world bytes=1425]
old peer:    [snapshot.sessions bytes=123493]
```

Stream stays open, every other row commits at `ready`. Provenance is irrelevant to the encode path, so old-spoke→new-hub→browser collapses to the same mechanism: the hub accepts the legacy row (1 MiB ceiling) and quarantines it on its own protocol-3 stream instead of dying.

**L3 re-run** — real `Encode` fed through the real peer receiver:

```
rows=100000            -> published=100000 events=190 diagnostics=0
rows=100003            -> published=100000 diagnostics=3 (transaction_limit)
dense 47KiB rows=1369  -> senderAcceptedJSON=66026870 receiverCounted=66061095
                          cap=67108864 headroom=1047769 published=1369
tiny rows=100000       -> envelope overhead 1.148% of accepted bytes
```

Worst-case envelope overhead 34 225 B against a 1 MiB reserve; row cap exact on both sides. Browser UTF-16 accounting undercounts bytes, so it is strictly more permissive and still bounded.

**M3 re-run** — three mutations, each now caught:

```
remove resetSessionsTransport() in error handler -> "discards an interrupted bootstrap" FAILS
remove the epoch <= lastSessionEpoch guard       -> "ignores duplicate and stale epochs" FAILS
make snapshot.sessions.error discard staging     -> "keeps a degraded epoch valid" FAILS
```

## Adjudicating the 512 KiB world cap — F1 (Medium)

Scaling the PR's own composed fixture (20 memberships/project, 20 peers, launchers, health, peer projects, discovery):

```
sessions   plain ids   namespaced (id@peer)
   1 000      26 177       34 177
   5 000     102 827      142 827
  10 000     198 827      278 827
  20 000     390 827      550 827   <-- over
  25 000     487 577      687 577   <-- over
 100 000   1 938 827    2 738 827   <-- over
crossing ~26 900 sessions (plain) / ~19 000 (namespaced); limit 524 288
```

The session transaction cap in the same commit is **100 000 rows**, so world dies at ~19–27 % of the envelope this PR chose to support. Past the crossing the browser gets `snapshot.world.error` on *every* update forever: `projects`/`peers`/`launchers` empty, project URLs mis-resolve to home (`store.ts:832-834`), signal = one `console.warn`.

Capping instead of framing is a legitimate interim call — far cheaper than a second protocol, and it does keep the stream up. But two things undercut the stated rationale:

- The ADR justifies 512 KiB as "below the Go transport's 1 MiB line ceiling", yet world never crosses that transport. `serve_central.go:869` gates the initial world on `!asPeer`, and `:903-907` `continue`s for peers before the world branch — the only consumers are browsers, whose `EventSource` has no line limit. On the real consumer path the cap protects nothing and only adds a loss mode.
- `TestRealisticComposedWorldBoundAtTransportSeam` measures 26 177 B, 5 % of the cap: it shows headroom at 1 000 sessions but says nothing about where the cap bites.

Resolution is a constant: raise it to cover the same 100 000-row envelope (~4 MiB), or defer capping until world is framed.

## F2 (Medium) — the world cap breaks the transitional old-tab guarantee added in the same delta

`sendWorldSSEFrame` is protocol-independent (both call sites use it regardless of `semanticSessions`). A protocol-2 subscriber — exactly the "old tab, peer, or custom consumer" this delta added the legacy fallback for — also receives `snapshot.world.error`, which the old bundle has no listener for. `worldLoaded` never flips, `view` (`store.ts:832-833`) stays `null`, and the old tab hangs on the connecting screen indefinitely. Before this delta it received the large world fine. Suggested: apply the cap only when `semanticSessions`, or send the oversized world unchanged to protocol-2 subscribers.

## F3 (Medium) — quarantined rows are indistinguishable from deletion

`sessionstream.go:104-118`: `ready` now means "every row that fit". A quarantined session vanishes from the sidebar and from a hub's mirror exactly as if killed; the only receiver-side signal is a `console.warn` / log line, and `/v1/sessions` (CLI) is the sole place it survives — so it can't be inspected or killed from the UI that dropped it. If every row is oversized, `ready` publishes the empty set.

Verified **not** durable: peer projections are runtime (`manager.go:380-425`), `centralPeerAdapter.assign` only upserts (`bootstrap_peering_adapters.go:111-141`), pruning is disconnect-driven via `PruneLocalPeer`. Suggested: carry omitted IDs (or a count) in `ready` so receivers retain the previous row and badge it instead of treating it as removed.

## F4 / F5 (Low)

- **F4** `snapshot.world.error` has no browser test. `store.ts:1930-1936` (`if (!worldLoaded.value) worldLoaded.value = true`) is the only thing preventing a permanent "Connecting" hang on first hydration. A throwaway probe (deleted) confirms the intent — `view` resolves, sessions render, `projects === []`, no user-visible indicator — but nothing checked in would catch mutating that line. Go side is covered.
- **F5** sseclient parser rewrite probed and correct: multi-line data joined with `\n`, CRLF, with/without space after colon, last `event:` field wins, aggregate cap enforced (`20x100KiB -> sse event data exceeds 1048576-byte limit`). Two behaviour changes, both benign or better: `data:` without `event:` now dispatches as `message` (peer's default case ignores it), and an unterminated final block at EOF is now dropped — a **dominance improvement**, since `bufio.Scanner` returns the final partial line at EOF and the old per-line parser could dispatch truncated JSON from a mid-write disconnect.

## Dominance argument for the replaced mechanisms

1. **All-or-nothing epoch** protected against exposing a partial set. Quarantine still exposes nothing before `ready` (mutation-verified: batch/ready without begin are inert), so commit atomicity is retained; it trades completeness for availability — dominant on availability, weaker on semantics (F3).
2. **Fatal error frame** protected receivers from acting on a broken transaction. Both receivers now treat it as advisory; m3 proves the browser no longer discards staging, and `peer.go:534-549` bounds the logged id/message.
3. **Per-event write deadline** → one transaction deadline plus `ctx.Err()` between events (`main.go:764-782`): restores protocol-2 stall parity *and* adds cancellation the old code lacked. Strictly better.
4. **Unconditional epoch reset** → strictly-increasing epoch guard with `abandon()` preserving `lastEpoch` (`peer.go:490-500`, `store.ts:1224-1233`), so a replayed begin can't destroy a newer in-flight transaction (m2).
5. **`!asPeer` implies protocol 3** → explicit opt-in for everyone plus a re-added legacy browser listener, covering both skew directions.

Commands: `go build ./...`, `go test -race ./internal/...`, `go test ./cmd/gmuxd/` (all pass), four throwaway probe suites (deleted), `npx vitest run` (525 pass; `store.test.ts` unresolved `@gmux/protocol` locally, green in CI), 3 mutation runs, `gh pr checks 477` (11/11 green). Not exercised: container E2E locally (docker-gated, green in CI), real two-daemon network peering, >27 k sessions in a live UI.

**Verdict: AMEND (narrow).** Every earlier finding is genuinely fixed and holds under mutation and end-to-end reproduction — the availability regression, the red check, negotiation, old-tab compatibility, deadline/cancellation and aggregate symmetry are all resolved. What remains is confined to the newly-added world cap: F2 creates a permanent hang for precisely the compatibility path this delta introduced, and F1 shows 512 KiB is inconsistent with the 100 000-row envelope the same commit chose. Both are constant-sized fixes. If you'd rather ship, F1/F2 are defensible as an immediate follow-up and the rest is merge-ready; F3/F4 deserve a tracking note.

mgabor3141 (commented):
# Final delta re-review — ca147c74fb6abf396332fd89267f6c52152971ae

Compared with `c828905dce055c29ab8cf6b72273bf85af568f61` and reviewed at the exact requested head. No subagents.

## Result

No remaining actionable finding in the requested state-machine scope. The mixed-mode rollback is fixed in both receivers by locking the first valid session protocol for the life of the transport:

- Peer: `services/gmuxd/internal/peering/peer.go:497-508` refuses v3 after legacy selection; `peer.go:559-573` refuses legacy after v3 selection. `abandon()` releases rows/diagnostics without clearing mode or `lastEpoch`.
- Browser: `apps/gmux-web/src/store.ts:1224-1236` refuses v3 after legacy; `store.ts:1908-1921` refuses legacy after v3. Only `resetSessionsTransport()` clears mode and epoch, and production invokes it at EventSource creation or its transport `error` boundary.

## Reproduced schedules

| Schedule | Peer | Browser | Result |
|---|---|---|---|
| v3 epoch 1, v3 epoch 2, injected legacy, replay v3 epoch 1 | `TestPeerLocksProtocol3AgainstLegacyInjectionAndStaleReplay` under `-race` | focused Vitest mode-lock test | remains on epoch-2 `new`; no rollback |
| legacy snapshot, then v3 begin/batch/ready | `TestPeerAcceptsLegacySnapshotFromOldSpoke` under `-race` | focused legacy-only test | legacy remains authoritative; malformed mode switch ignored |
| v3 epochs only, duplicate/older begin, stale begin during newer staging | `TestPeerRejectsNonIncreasingEpochWithoutRollback` under `-race` | focused stale-epoch test | newer projection/staging preserved |
| interrupted v3, disconnect/reconnect, epoch restarts at 1 | connection-local peer staging test under `-race` | focused reconnect test | staging released; epoch 1 accepted on new transport |
| malformed/stale batch or ready without matching active begin | mechanism trace plus focused reconnect/stale tests | focused tests | no false ready and no partial commit |

## Compatibility assessment

No valid compatibility transition is rejected:

- A one-release-old daemon ignores `session_stream=3` and emits legacy as its first session event; unknown → legacy remains accepted.
- A v3 daemon emits begin first; unknown → v3 remains accepted.
- Browser EventSource reconnection raises the `error` event before a replacement response, resetting mode. Peer reconnect creates a fresh connection-local bootstrap. Thus an old-to-new or new-to-old daemon/proxy change across a real transport boundary renegotiates normally.
- Switching formats within one HTTP/SSE response is not a valid negotiated transition; rejecting it is the required replay defense.
- Invalid begin while mode is unknown clears staging but does not select v3, so a subsequently valid legacy-first stream is still accepted. Invalid batch/ready cannot publish rows.

Rows, warning diagnostics, and aggregate buffers are nilled on abandonment/reset. No new leak, false-ready path, or stale-epoch bypass was found.

## Commands

- `go test -race ./services/gmuxd/internal/peering -run 'TestPeerLocksProtocol3AgainstLegacyInjectionAndStaleReplay|TestPeerAcceptsLegacySnapshotFromOldSpoke|TestPeerRejectsNonIncreasingEpochWithoutRollback|TestPeerSessionBootstrapIsAtomicAndDisconnectLocal' -count=1`
- `cd apps/gmux-web && pnpm exec vitest run src/sse-reconnect.test.ts`
- `jj status`; delta inspection from `c828905d...` to `ca147c74...`

Both focused executions passed. No temporary probe or production file remains; working-copy diff is clean.

## Verdict

**MERGE** — the original stale-epoch finding and the mixed-mode follow-up are both closed, reconnect renegotiation remains valid, and no requested-scope regression was reproduced.

mgabor3141:
Final delta review at exact head `ca147c74fb6abf396332fd89267f6c52152971ae`: **AMEND**.

The previous deep-link failure is gone: `snapshot.world.error` and the 512 KiB cap were removed, so an initial world rejection can no longer set `worldLoaded=true` and trigger URL normalization against empty projects. Valid world loading remains gated correctly; legacy session fallback is mode-locked and cannot mix with v3. Quarantine warnings are staged by epoch, published only at `ready`, cleared by the next valid ready, and do not interfere with session slug rewrite/navigation. Exact-head probes passed multiline WHATWG SSE parsing, aggregate/1 MiB scanner bounds, session bounds, quarantine/ready, and sender-limit tests.

**P1 — removing the world cap restores the unbounded SSE-line failure.** `services/gmuxd/cmd/gmuxd/main.go:720-725` now only has generic `sendSSEFrame`; the previous `sendWorldSSEFrame`/`maxWorldEventPayload` is absent, and `serve_central.go` sends world through the generic path. Supported projects, peers, health, launchers, peer-project, or discovery fields can therefore make one JSON `snapshot.world` data line arbitrarily large. The Go peer does not receive world, but programmatic/browser-proxy consumers and the Go client's 1 MiB ceiling remain exposed.

The ADR now accurately documents this at `docs/adr/0001-snapshot-push-protocol.md:286-291`: 48 KiB applies only to protocol-3 session events and no world/endpoint-wide bound is claimed. That accurately scopes the guarantee but does not mitigate the safety gap. Either make this explicit as an accepted design tradeoff, restore bounded world handling without marking an initial error as loaded, or semantically fragment world.

Performance remains the prior tradeoff: semantic framing costs ~1.4x CPU and ~2.4x allocated bytes on the 1,000-row serialization fixture for ~0.15% wire overhead; no additional blocker found.

mgabor3141 (commented):
Final delta re-review at `ca147c74fb6abf396332fd89267f6c52152971ae` (advisory). Verified only `c828905..ca147c7` plus adjacent failure paths; report appended to `.memory/review-477-opus.md`.

## Round-2 findings verified

**F1 + F2 (world cap) — fully resolved, no stale claims.** `sendWorldSSEFrame`, `maxWorldEventPayload` and `worldStreamError` are gone; both call sites revert to `sendSSEFrame(rc, w, "snapshot.world", ...)` (`serve_central.go:872,911`) and the browser's `snapshot.world.error` listener is removed. Exhaustive grep across `*.go/*.ts/*.tsx/*.md` for `world.error|maxWorldEventPayload|world_too_large|sendWorldSSEFrame|512 KiB|worldStreamError` returns **zero** hits. The ADR now explicitly disclaims the over-reach ("makes no world-size or endpoint-wide boundedness claim... retained as a size characterization, not a supported maximum"), the boundary test is demoted to `TestRealisticComposedWorldSizeCharacterization`, and `state-management.md` matches.

**Legacy old-tab sequence unchanged — differential probe.** Same two-runner fixture (one 60 KiB argv), same probe, run against the pre-PR merge base and this head:

```
merge base 0980f999   LEGACY-BROWSER [snapshot.sessions[bytes=123505] snapshot.world[bytes=1425]]
                      LEGACY-PEER    [snapshot.sessions[bytes=123505]]
head       ca147c7    LEGACY-BROWSER [snapshot.sessions[bytes=123505] snapshot.world[bytes=1425]]
                      LEGACY-PEER    [snapshot.sessions[bytes=123505]]
```

Byte-identical, including the oversized row and the world frame — an unversioned tab or custom consumer sees exactly the pre-PR protocol-2 stream.

**Quarantine surfaced and cleared; no availability regression.**

```
V3-BROWSER [begin[23], error{"epoch":1,"code":"row_too_large","id":"bbbbbbbb","message":"session row
            encoded to 123191 bytes and exceeds the 49152-byte event limit; row omitted","count":1},
            batch[ids=aaaaaaaa], ready[11], snapshot.world[bytes=1425]]
V3-PEER    [begin[23], error{...}, batch[ids=aaaaaaaa], ready[11]]
```

Diagnostics stage per-epoch (`store.ts:1254-1257`) and publish atomically at `ready` into `sessionStreamWarnings`, rendered as a persistent sidebar banner with a per-row title (`sidebar.tsx:645-646,685-693`); cleared by the next clean `ready`, by legacy takeover, and at `initStore` (probe-confirmed, including that warnings deliberately survive a transport drop alongside the retained list). Mutations each kill exactly one test:

```
remove `if (sessionStreamMode === 'legacy') return`  -> "accepts the transitional legacy replacement" FAILS
remove `if (sessionStreamMode === 'v3') return`      -> "locks a negotiated transport to protocol 3"  FAILS
remove `sessionStreamWarnings.value = warnings`      -> "keeps a degraded epoch valid"                FAILS
```

**Mode locking (new mechanism) — adjacent failure paths probed.** Mode lives in the per-connection `peerSessionBootstrap` (`peer.go:493`) and a browser module variable reset by `resetSessionsTransport()`; `abandon()` deliberately preserves `mode`/`lastEpoch`.

```
malformed legacy frame       -> mode stays unknown; protocol 3 still usable afterwards
version-2 begin              -> mode stays unknown, staging inactive; legacy still usable
diagnostic for a stale epoch -> not attached to the live epoch
```

Only a well-formed frame locks the transport, so a corrupt event cannot pin a connection to the wrong protocol. With strict epoch monotonicity this closes the mixed-mode rollback vector.

## Residual findings (Low, none blocking)

**G1 — omitted-session count undercounts beyond 256.** `sessionstream.go:100-105,146-151` emits at most 256 individual diagnostics plus one `diagnostics_suppressed` carrying the remainder in `count`. The browser caps staging at `warnings.length < 256` (`store.ts:1256`) and the summary arrives **last**, so the one event whose `count` exists to carry the remainder is exactly the one dropped:

```
300 oversized rows -> events=259 batches=0 individualDiagnostics=256 suppressedCount=44 sum(count)=300
store probe        -> stored=256 displayed=256 (true omitted=300)
```

Banner reads "256 sessions are omitted" when 300 are. Fix: exempt `diagnostics_suppressed` from the cap, or cap individual warnings at 255.

**G2 — `peerSessionBootstrap.diagnostics` is write-only.** Appended at `peer.go:568-569`, cleared at `:499`, never read (`EventReady` at `:541-551` discards it). The only observable effect is the log line, yet `session_stream_test.go:112-114` asserts the field — an assertion over inert state — and the ADR's "peers retain the bounded diagnostic for the transaction" overstates what exists.

**G3 — duplicated legacy path, one copy unreachable.** `handleStreamEvent`'s `case "snapshot.sessions"` (`peer.go:573-587`) returns on every branch and `handleEvent` has exactly one caller (`:588`), so `handleEvent`'s own `case "snapshot.sessions"` (`:604`) is dead. Two copies of the same unmarshal-and-apply logic invite divergence.

**G4 (carried)** — with every row oversized, `batches=0` and `ready` commits the empty set; now at least surfaced by the banner rather than silent, and `/v1/sessions` still lists the rows.

Commands: `go build ./...`, `go test -race ./internal/...`, `go test ./cmd/gmuxd/` (pass), differential probe at both revisions, peering mode-lock probes, store warning probe, 3 mutation runs, `npx vitest run` (526 pass; `store.test.ts` unresolved `@gmux/protocol` locally, green in CI), `gh pr checks 477` → 11/11 pass. All probes deleted, working copy clean. Not exercised: container E2E locally (docker-gated, green in CI), real two-daemon peering, visual rendering of the banner.

**Verdict: MERGE.** All four verification targets hold — world cap fully removed with no stale claim anywhere, legacy sequence byte-identical to the pre-PR daemon, quarantine correctly surfaced and cleared with mutation-backed tests, and no availability regression on any of the four consumer shapes. The new mode lock is sound and fails open to the correct protocol on malformed input. G1 is a cosmetic undercount in a new indicator; G2/G3 are dead code and an inert assertion — worth a follow-up, not another round.

mgabor3141 (commented):
G1 count-fix verification at `4bc28a2f41b807f12ec8fd10ac8b74d90836a86e` (advisory). Scope limited to `ca147c7..4bc28a2f`; report appended to `.memory/review-477-opus.md`.

## Mechanism

The single capped list is split into an unbounded-safe scalar plus bounded detail: `sessionStreamOmittedTotal` (`store.ts:249`), staged per epoch (`:1221,1242`). `appendSessionStreamWarning` (`:1257-1263`) sanitizes `count`, always adds it to the total (clamped at `MAX_SAFE_INTEGER`), and appends to the detail list only when `warning.id` is non-empty **and** the list is under 256 — so the sender's ID-less `diagnostics_suppressed` event can never be starved out of the total by the detail cap. Both publish atomically at `ready` (`:1269-1273`) and reset at `initStore` (`:1855`) and legacy takeover (`:1931`). The sidebar (`sidebar.tsx:646-648,687-696`) shows the total and derives `suppressed = max(0, total − Σ detail.count)`.

## Reproduction — real daemon payloads through the real listeners

A throwaway Go probe dumped the exact wire output of `sessionstream.Encode` for 300 oversized rows; a throwaway vitest probe replayed those bytes through the production `snapshot.sessions.*` listeners.

```
events by type: {begin:1, error:257, ready:1}
first={"epoch":1,"code":"row_too_large","id":"oversized-000","message":"session row encoded to
       49189 bytes and exceeds the 49152-byte event limit; row omitted","count":1}
last ={"epoch":1,"code":"diagnostics_suppressed","message":"44 additional omitted session rows
       were not individually reported","count":44}
```

| property | expected | observed |
|---|---|---|
| exact total | 300 | **300** |
| detail entries | ≤ 256 | **256**, every entry has a non-empty ID |
| summary consumes a detail slot | no | no (`diagnostics_suppressed` absent from details) |
| sidebar suppressed arithmetic | 44 | `total − Σ detail.count = 44` |
| visible before `ready` | nothing | total 0, details 0 while all 257 staged |
| clean `ready` clears both | yes | epoch 2 clean bootstrap → total 0, details `[]`, rows `['ok']` |

Adversarial probes:

```
50 000 diagnostics, 200-char messages -> stored=256  total=50000   (O(1) counter + O(256) detail)
missing / negative / non-integer count -> total=3 stored=3         (each sanitized to 1)
malformed JSON diagnostic              -> ignored, staging intact
transport drop, then fresh epoch        -> total 7 retained, then cleared by the next ready
```

Diagnostic memory is therefore bounded regardless of sender behaviour: the detail list is hard-capped at 256 objects and everything above collapses into one integer. Totals/details deliberately survive a transport drop alongside the retained session list and are replaced atomically at the next `ready`.

## Test quality

The checked-in browser test (`sse-reconnect.test.ts:170-186`) calls `appendSessionStreamWarning` directly rather than emitting through the EventSource, so the listener's `id: diagnostic.id ?? ''` mapping isn't covered by it — my probe closed that gap end-to-end and the mapping is correct. Mutation check, gating `omittedTotal` on the detail cap (pre-fix behaviour):

```
-> "keeps the exact omitted total above the bounded detail cap" FAILS
restored -> 11 passed
```

`TestEncodeDiagnosticSummaryPreservesTotalAboveDetailCap` pins the sender contract the browser now relies on (`details == 256`, `Σ count == 300`, `summary.Count == 44`), and the ADR text matches.

Commands: `go build ./...`, `go test -race ./internal/sessionstream/... ./internal/peering/...` (pass), two throwaway probes (deleted), 1 mutation run, `npx vitest run` (527 pass; `store.test.ts` unresolved `@gmux/protocol` locally, green in CI), `gh pr checks 477` → 11/11 pass. Probes deleted, working copy clean.

**Verdict: MERGE.** G1 is fixed exactly as specified — exact total of 300, detail bounded at 256 and never displaced by the summary, clean `ready` clears both, and diagnostic memory is O(1) + O(256) even under a 50 000-event flood. No new issue in this delta. G2 (write-only `peerSessionBootstrap.diagnostics`) and G3 (unreachable duplicate legacy branch in `handleEvent`) remain cosmetic follow-ups.

PR #480, opened 2026-08-10
feat(web): add mobile Shift+Tab terminal key

mgabor3141:
## Summary
- Adds an accessible inline-SVG BackTab button to the mobile terminal keyboard.
- Places it in the former Esc slot and moves Esc above the hamburger/menu control in portrait; responsive 600–819 and >=820 layouts retain their intended ordering.
- Uses the canonical key sequence path: `keyComboToSequence('shift+tab')` emits one xterm BackTab sequence, `ESC [ Z` (`\x1b[Z`).
- Preserves existing focus/modifier semantics and single-fire behavior.
- Keeps the hamburger status dot visible and scoped: only the mobile keyboard/menu instance uses the inset badge; home/floating/footer placements retain their established overhang.
- Reserves horizontal safe-area insets in landscape.

## Focused browser regression
`e2e/tests/mobile-shift-tab.spec.ts` uses a real live terminal session and Playwright mobile context. It:
- holds and activates the accessible BackTab control;
- observes terminal-input binary WebSocket frames separately from JSON resize/control frames and asserts exactly one `[0x1b, 0x5b, 0x5a]` payload (`\x1b[Z`), while preventing fixture-shell mutation;
- proves no repeat-on-hold, menu activation, focus change, or Ctrl/Alt latch change;
- compares the unread badge against the actual `.mk-esc .mkey-face` painted rect;
- asserts portrait, 600–819, >=820, and small-phone ordering/visibility;
- asserts the production `env(safe-area-inset-left/right)` → custom-property declarations, plus synthetic 24px/28px side-inset geometry.

Mutation checks confirmed the suite fails when:
- the mobile dot is reverted to `top/right: -3px`;
- either production safe-area `env()` declaration is replaced with a literal.

## Tests
- `pnpm --filter @gmux/protocol build` ✅
- `pnpm --filter @gmux/web test` ✅ — 34 files, 711 tests
- `pnpm --filter @gmux/web lint` ✅
- `pnpm --filter @gmux/web build` ✅
- `pnpm test:e2e` ✅ — 49 tests
- Focused Playwright regression ✅ — 2 tests

## Visual evidence
- `.memory/screenshots/mobile-shift-tab-portrait.png` — 390×844.
- `.memory/screenshots/mobile-shift-tab-landscape.png` — 844×390.
- `.memory/screenshots/mobile-shift-tab-medium.png` — 700×390.
- `.memory/screenshots/mobile-shift-tab-small-phone.png` — 320×568.
- `.memory/screenshots/mobile-shift-tab-landscape-safe-area-synthetic.png` — 844×390 with synthetic 24px left/28px right insets.
- `.memory/report-mobile-shift-tab.md` documents observations and validation.

### DIFF
--- a/apps/gmux-web/src/keybinds.test.ts
+++ b/apps/gmux-web/src/keybinds.test.ts
@@ -64,8 +64,9 @@ describe('keyComboToSequence', () => {
     expect(keyComboToSequence('esc')).toBe('\x1b')
   })
 
-  it('converts tab', () => {
+  it('converts tab and physical Shift+Tab to terminal sequences', () => {
     expect(keyComboToSequence('tab')).toBe('\t')
+    expect(keyComboToSequence('shift+tab')).toBe('\x1b[Z')
   })
 
   it('converts backspace', () => {
--- a/apps/gmux-web/src/keybinds.ts
+++ b/apps/gmux-web/src/keybinds.ts
@@ -218,7 +218,9 @@ export function keyComboToSequence(combo: string): string {
   } else if (baseKey === 'escape' || baseKey === 'esc') {
     seq = '\x1b'
   } else if (baseKey === 'tab') {
-    seq = '\t'
+    // xterm's canonical BackTab/Shift+Tab sequence. Do not synthesize a
+    // browser Tab event: terminal applications expect ESC [ Z as one key.
+    seq = shift && !ctrl && !alt ? '\x1b[Z' : '\t'
   } else if (baseKey === 'backspace') {
     // Ctrl+Backspace: ESC DEL (readline backward-kill-word) so it deletes a
     // word rather than a char. Plain backspace: \x7f (DEL).
--- a/apps/gmux-web/src/keyboard.test.ts
+++ b/apps/gmux-web/src/keyboard.test.ts
@@ -171,6 +171,12 @@ describe('applyArmedModifiers', () => {
     expect(applyArmedModifiers('\x1b[A', true, true).seq).toBe('\x1b[1;7A')   // ctrl+alt+up
   })
 
+  it('encodes armed modifiers on BackTab without splitting the key', () => {
+    expect(applyArmedModifiers('\x1b[Z', false, false).seq).toBe('\x1b[Z')
+    expect(applyArmedModifiers('\x1b[Z', true, false).seq).toBe('\x1b[1;5Z')
+    expect(applyArmedModifiers('\x1b[Z', false, true).seq).toBe('\x1b[1;3Z')
+  })
+
   it('encodes alt+enter as ESC CR', () => {
     expect(applyArmedModifiers('\r', false, true))
       .toEqual({ seq: '\x1b\r', ctrlApplied: false, altApplied: true })
--- a/apps/gmux-web/src/keyboard.ts
+++ b/apps/gmux-web/src/keyboard.ts
@@ -298,7 +298,7 @@ export interface ArmedModifierResult {
  *    CSI-u for uppercase), alt via ESC prefix, both combined when armed
  *    together (legacy ESC + control code; CSI-u modifier bits for
  *    uppercase).
- *  - CSI cursor keys (arrows, Home, End): modifier parameter injection,
+ *  - CSI cursor keys (arrows, Home, End, BackTab): modifier parameter injection,
  *    e.g. \x1b[D + ctrl → \x1b[1;5D (the standard word-left encoding).
  *  - Enter / Tab / Esc: alt uses the universal ESC prefix. Ctrl has no
  *    legacy encoding for these (ctrl+enter is byte-identical to enter),
@@ -319,7 +319,7 @@ export function applyArmedModifiers(
 
   // CSI cursor-key sequences: inject the modifier parameter.
   // biome-ignore lint/suspicious/noControlCharactersInRegex: matching real ESC (\x1b) in terminal control sequences
-  const csi = /^\x1b\[([A-DHF])$/.exec(data)
+  const csi = /^\x1b\[([A-DFHZ])$/.exec(data)
   if (csi) return { seq: `\x1b[1;${mod}${csi[1]}`, ctrlApplied: ctrl, altApplied: alt }
 
   // Enter / Tab / Esc with ctrl involved: CSI-u (no legacy encoding exists).
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -456,7 +456,7 @@ const IconDown  = () => <svg viewBox="0 0 14 14" width="16" height="16" {...S}><
 const IconLeft  = () => <svg viewBox="0 0 14 14" width="16" height="16" {...S}><path d="M10 7H4m0 0 3-3M4 7l3 3"/></svg>
 const IconRight = () => <svg viewBox="0 0 14 14" width="16" height="16" {...S}><path d="M4 7h6m0 0-3-3m3 3-3 3"/></svg>
 
-const IconBackTab  = () => <svg viewBox="0 0 18 14" width="20" height="16" {...S}><line x1="14.5" y1="3" x2="14.5" y2="11"/><path d="M12 7H5m0 0 3-3M5 7l3 3"/></svg>
+const IconBackTab  = () => <svg viewBox="0 0 24 18" width="22" height="16" {...S}><path d="M5 6V2m0 0L3 4m2-2 2 2"/><line x1="10" y1="7" x2="10" y2="15"/><path d="M19 11h-6m0 0 3-3m-3 3 3 3"/></svg>
 const IconWordLeft  = () => <svg viewBox="0 0 18 14" width="20" height="16" {...S}><line x1="3.5" y1="3" x2="3.5" y2="11"/><path d="M13 7H6m0 0 3-3M6 7l3 3"/></svg>
 const IconWordRight = () => <svg viewBox="0 0 18 14" width="20" height="16" {...S}><line x1="14.5" y1="3" x2="14.5" y2="11"/><path d="M5 7h7m0 0-3-3m3 3-3 3"/></svg>
 const IconSend = () => <svg viewBox="0 0 14 14" width="16" height="16" fill="currentColor" stroke="none"><path d="M3 2.5l8 4.5-8 4.5V8.5L7.5 7 3 5.5z"/></svg>
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -6,6 +6,7 @@ import '@xterm/xterm/css/xterm.css'
 import './styles.css'
 
 import { applyArmedModifiers } from './keyboard'
+import { keyComboToSequence } from './config'
 import { isTouchDevice } from './touch'
 import { ReplayView } from './replay-view'
 import { TerminalView } from './terminal'
@@ -548,7 +549,7 @@ function MobileTerminalBar({
 
   // The bar is a CSS grid laid out via named areas (.mk-* → grid-area), so the
   // DOM order below is only tab/reading order — the visual arrangement lives
-  // in styles.css. Narrow phones get a 7×2 grid (empty top-left corner;
+  // in styles.css. Narrow phones get a 7×2 grid (Esc sits above the hamburger;
   // scroll-end or empty top-right). Wider viewports (landscape / tablets)
   // collapse to a single row, and the widest step folds the word-jumps back
   // in. Keys never relabel; ctrl/alt only arm + highlight.
@@ -557,8 +558,9 @@ function MobileTerminalBar({
   return (
     <div class="mobile-bottom-bar" role="toolbar" aria-label="Terminal keys" onMouseDown={keepFocus}>
       <MenuButton variant="bar" onMenu={onMenu} />
-      <button class="mobile-bottom-action mk-esc" disabled={!canSend} onClick={() => sendKey('\x1b')} title="Escape"><span class="mkey-face">esc</span></button>
-      <button class="mobile-bottom-action mk-tab" disabled={!canSend} onClick={() => sendKey('\t')} title="Tab"><span class="mkey-face">tab</span></button>
+      <button class="mobile-bottom-action mk-esc" disabled={!canSend} aria-label="Escape" onClick={() => sendKey('\x1b')} title="Escape"><span class="mkey-face">esc</span></button>
+      <button class="mobile-bottom-action mk-shift-tab" disabled={!canSend} aria-label="Shift+Tab (BackTab)" onClick={() => sendKey(keyComboToSequence('shift+tab'))} title="Shift+Tab (BackTab)"><span class="mkey-face">⇧tab</span></button>
+      <button class="mobile-bottom-action mk-tab" disabled={!canSend} aria-label="Tab" onClick={() => sendKey('\t')} title="Tab"><span class="mkey-face">tab</span></button>
       <button class={`${armedClass(ctrlArmed)} mk-ctrl`} disabled={!canSend} aria-pressed={ctrlArmed} onClick={onToggleCtrl} title={ctrlArmed ? 'Ctrl armed for next key' : 'Arm Ctrl'}><span class="mkey-face">ctrl</span></button>
       <button class={`${armedClass(altArmed)} mk-alt`} disabled={!canSend} aria-pressed={altArmed} onClick={onToggleAlt} title={altArmed ? 'Alt armed for next key' : 'Arm Alt'}><span class="mkey-face">alt</span></button>
       <button class="mobile-bottom-action mk-wl" disabled={!canSend} {...repeat(() => sendWord('\x1b[D'), WORD_REPEAT_MS)} title="Word left"><span class="mkey-face"><IconWordLeft /></span></button>
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -456,6 +456,7 @@ const IconDown  = () => <svg viewBox="0 0 14 14" width="16" height="16" {...S}><
 const IconLeft  = () => <svg viewBox="0 0 14 14" width="16" height="16" {...S}><path d="M10 7H4m0 0 3-3M4 7l3 3"/></svg>
 const IconRight = () => <svg viewBox="0 0 14 14" width="16" height="16" {...S}><path d="M4 7h6m0 0-3-3m3 3-3 3"/></svg>
 
+const IconBackTab  = () => <svg viewBox="0 0 18 14" width="20" height="16" {...S}><line x1="14.5" y1="3" x2="14.5" y2="11"/><path d="M12 7H5m0 0 3-3M5 7l3 3"/></svg>
 const IconWordLeft  = () => <svg viewBox="0 0 18 14" width="20" height="16" {...S}><line x1="3.5" y1="3" x2="3.5" y2="11"/><path d="M13 7H6m0 0 3-3M6 7l3 3"/></svg>
 const IconWordRight = () => <svg viewBox="0 0 18 14" width="20" height="16" {...S}><line x1="14.5" y1="3" x2="14.5" y2="11"/><path d="M5 7h7m0 0-3-3m3 3-3 3"/></svg>
 const IconSend = () => <svg viewBox="0 0 14 14" width="16" height="16" fill="currentColor" stroke="none"><path d="M3 2.5l8 4.5-8 4.5V8.5L7.5 7 3 5.5z"/></svg>
@@ -559,7 +560,7 @@ function MobileTerminalBar({
     <div class="mobile-bottom-bar" role="toolbar" aria-label="Terminal keys" onMouseDown={keepFocus}>
       <MenuButton variant="bar" onMenu={onMenu} />
       <button class="mobile-bottom-action mk-esc" disabled={!canSend} aria-label="Escape" onClick={() => sendKey('\x1b')} title="Escape"><span class="mkey-face">esc</span></button>
-      <button class="mobile-bottom-action mk-shift-tab" disabled={!canSend} aria-label="Shift+Tab (BackTab)" onClick={() => sendKey(keyComboToSequence('shift+tab'))} title="Shift+Tab (BackTab)"><span class="mkey-face">⇧tab</span></button>
+      <button class="mobile-bottom-action mk-shift-tab" disabled={!canSend} aria-label="Shift+Tab (BackTab)" onClick={() => sendKey(keyComboToSequence('shift+tab'))} title="Shift+Tab (BackTab)"><span class="mkey-face"><IconBackTab /></span></button>
       <button class="mobile-bottom-action mk-tab" disabled={!canSend} aria-label="Tab" onClick={() => sendKey('\t')} title="Tab"><span class="mkey-face">tab</span></button>
       <button class={`${armedClass(ctrlArmed)} mk-ctrl`} disabled={!canSend} aria-pressed={ctrlArmed} onClick={onToggleCtrl} title={ctrlArmed ? 'Ctrl armed for next key' : 'Arm Ctrl'}><span class="mkey-face">ctrl</span></button>
       <button class={`${armedClass(altArmed)} mk-alt`} disabled={!canSend} aria-pressed={altArmed} onClick={onToggleAlt} title={altArmed ? 'Alt armed for next key' : 'Arm Alt'}><span class="mkey-face">alt</span></button>
--- a/apps/gmux-web/src/styles.css
+++ b/apps/gmux-web/src/styles.css
@@ -3211,7 +3211,7 @@ body {
     grid-template-columns: repeat(7, 1fr);
     grid-template-rows: repeat(2, 40px);
     grid-template-areas:
-      '.    esc  tab  wl  au  wr  end'
+      'esc  stab tab  wl  au  wr  end'
       'menu ctrl alt  al  ad  ar  send';
     /* One knob for spacing: the visual gap between keys. Hitboxes bridge it
        automatically (each .mkey-face extends half of it), so there are no
@@ -3229,6 +3229,7 @@ body {
      only the template (above / below) changes between layouts. */
   .mk-menu { grid-area: menu; }
   .mk-esc  { grid-area: esc; }
+  .mk-shift-tab { grid-area: stab; }
   .mk-tab  { grid-area: tab; }
   .mk-ctrl { grid-area: ctrl; }
   .mk-alt  { grid-area: alt; }
@@ -4033,9 +4034,9 @@ body {
 /* Single row, word-jumps dropped. */
 @media (pointer: coarse) and (min-width: 600px) {
   .mobile-bottom-bar {
-    grid-template-columns: repeat(10, 1fr);
+    grid-template-columns: repeat(11, 1fr);
     grid-template-rows: 40px;
-    grid-template-areas: 'menu esc tab ctrl alt al au ad ar send';
+    grid-template-areas: 'menu esc stab tab ctrl alt al au ad ar send';
   }
   /* Word-jumps aren't part of the single-row layout. */
   .mk-wl,
@@ -4062,8 +4063,8 @@ body {
 /* Wider still: fold the word-jumps back in, flanking the arrow cluster. */
 @media (pointer: coarse) and (min-width: 820px) {
   .mobile-bottom-bar {
-    grid-template-columns: repeat(12, 1fr);
-    grid-template-areas: 'menu esc tab ctrl alt wl al au ad ar wr send';
+    grid-template-columns: repeat(13, 1fr);
+    grid-template-areas: 'menu esc stab tab ctrl alt wl al au ad ar wr send';
   }
   .mk-wl,
   .mk-wr {
--- a/apps/gmux-web/src/styles.css
+++ b/apps/gmux-web/src/styles.css
@@ -3380,15 +3380,22 @@ body {
     width: 8px;
     height: 8px;
     border-radius: 50%;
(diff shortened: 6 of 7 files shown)


mgabor3141 (commented):
## Advisory review — PR #480 @ `3072df216d12f8c288478f338db600d5a6bc9f95`

### Finding (medium): the new tests stop before the behavior this PR promises

`apps/gmux-web/src/keybinds.test.ts:67-70` proves only the pure conversion helper, and `apps/gmux-web/src/keyboard.test.ts:174-177` proves only modifier transformation. Neither test renders/clicks the new button at `apps/gmux-web/src/main.tsx:559-563`, observes the sole `onSend` call at `main.tsx:530-535`, or reaches the canonical raw terminal route `main.tsx:783` → `terminal.tsx:556` → `terminal.tsx:537-546`. Consequently there is no assertion protecting the stated user-visible guarantees: one send (not zero/duplicate), exact bytes, default prevention/no focus traversal, no repeat handlers, unchanged unarmed latches, and no accidental Esc/Tab/menu dispatch. A regression such as wiring the button to Tab, adding `repeat(...)`, or invoking menu as well as send would leave all new tests green.

The responsive-ordering claim is likewise unprotected: the portrait and 600/820px layouts are implemented at `apps/gmux-web/src/styles.css:3211-3215`, `4035-4045`, and `4064-4071`, but there are no responsive geometry/order assertions in the web or E2E suites. Thus there are no ordering assertions to evaluate for vacuity; the PR description's browser checks are manual evidence, not executable regression protection.

Mechanism trace indicates the implementation itself is coherent: the click directly calls `keyComboToSequence('shift+tab')` (`main.tsx:562`), conversion returns the single JS string `\x1b[Z` (`keybinds.ts:220-223`), unarmed modifier handling returns it unchanged (`keyboard.ts:316-317`), `sendKey` calls `onSend` once (`main.tsx:530-535`), and the callback reaches exactly one `WebSocket.send(TextEncoder.encode(data))` through the registered raw-input callback (`main.tsx:783`, `terminal.tsx:537-556`). The BackTab button has only `onClick`, unlike repeatable controls at `main.tsx:566-570`; toolbar `mousedown` is prevented at `main.tsx:524,559`, so the click path synthesizes no browser Tab event and does not traverse focus. Esc, Tab, and menu are separate sibling handlers (`main.tsx:560-563`).

Please add a focused component/browser test that clicks the accessible BackTab button and asserts exactly one raw send of exactly `\x1b[Z`, no menu/focus/latch side effects, and no repeat across a held pointer; add viewport assertions for portrait, 600px, and 820px ordering/visibility so the responsive checks are non-vacuous.

### Mutation evidence

- Mutating `keyComboToSequence` to always return Tab made `keybinds.test.ts:69` fail as expected.
- Removing `Z` from the modifier CSI regex made `keyboard.test.ts:176` fail as expected.
- Those mutations demonstrate the unit assertions are load-bearing for their helpers, but not for toolbar wiring or responsive behavior.

### Commands / coverage

- `pnpm --filter @gmux/protocol build` — pass
- `pnpm --dir apps/gmux-web exec vitest run src/keybinds.test.ts src/keyboard.test.ts` — 67/67 pass
- `pnpm --filter @gmux/web lint` (`tsc --noEmit`) — pass
- `pnpm exec biome lint` on all six touched files — pass
- Initial web-script invocation ran all web tests before protocol build: 33 files / 541 tests passed; `store.test.ts` could not resolve the then-unbuilt `@gmux/protocol`. Focused tests passed after building it.
- Live Chrome probe was blocked because the configured chrome-devtools daemon expects `/opt/google/chrome/chrome`, absent on this host. No production files were changed.

**Verdict: AMEND**

mgabor3141 (commented):
## Adversarial review — advisory · **AMEND**

Reviewed exactly `3072df216d12f8c288478f338db600d5a6bc9f95`. Ran the UI myself (Playwright chromium, `?mock`, `isMobile`+`hasTouch`, DPR 2–3) at **390×844**, **844×390**, **320×568** plus 360/414/599/600/667/819/1024 to catch the breakpoint edges. Full report + all screenshots: `.memory/review-480-opus.md`, `.memory/shots/`.

**The layout claims all hold.** Verified independently:

* **Esc above hamburger** — 390×844: `.mk-esc` at `(4, 755, 50×40)`, `.mk-menu` at `(4, 800, 50×40)`. Same column, adjacent rows. ✅
* **Shift+Tab in the former Esc slot** — `.mk-shift-tab` is row 1 / col 2 at every portrait width (x=59 @390, x=49 @320, x=113 @600). ✅
* **Landscape coherence** — 600–819px → 11 cols `menu esc stab tab ctrl alt al au ad ar send`; ≥820px → 13 cols with the word-jumps folded back in. Reads `☰ esc ⇧tab tab ctrl alt |← ← ↑ ↓ → →| ▷`, matching both the portrait row order and DOM/focus order. ✅
* **No clipping / no overlap / no overflow** at any of the 10 viewports. `.mkey-face` `scrollWidth == clientWidth` everywhere; `⇧tab` ink is 30px inside a 45px face at 320px; label height identical (45px) to `esc`/`tab`/`ctrl` → no wrap. Min target 40.3×40 (portrait unchanged; landscape columns shrink 64.9→59.7 @844 and 54.7→49.3 @600 from the extra column, all still ≥40). ✅
* **Terminal viewport unchanged** — `measureTerminalFit` reserves the bar's measured `offsetHeight` (`terminal.tsx:87-89`); bar height is 89/44 both with this template and with the pre-PR template injected, so row counts are identical. Safe-area padding untouched, bar bottom == viewport bottom. ✅
* **Menu, focus, style parity** — ☰ still opens the sidebar at all viewports; tapping `⇧tab` doesn't steal focus or pop the keyboard; `:focus-visible` ring present; computed `color/background/border/radius/font-weight/touch-action/user-select/tap-highlight` byte-identical to its siblings. Zero console errors. ✅
* **Sequences** — mutation-verified that both new assertions bite: reverting `keybinds.ts:223` → `expected '\t' to be '\u001b[Z'`; narrowing the regex to `[A-DFH]` → `expected '\u001b[Z' to be '\u001b[1;5Z'` (and that shows what the regex fix buys — without it ctrl-armed `⇧tab` sent bare `ESC [ Z` and *left Ctrl armed*, leaking onto the next key). ✅

Two user-visible visual defects in the changed region did survive the PR's own visual check, which is why I land on amend rather than merge.

---

### F1 — Low: the ☰ status dot now collides with the `esc` key border

`styles.css:3213` moves `esc` into row 1 / col 1, i.e. directly above `menu`. `MenuButton`'s status dot is an `::after` at `top:-3px; right:-3px; 8×8` that deliberately overhangs the menu key's top-right corner. Measured at 390×844:

* dot box: x `49.3–57.3`, y `797.0–805.0`
* `.mk-esc .mkey-face`: x `1.5–56.8`, y `752.5–797.5` → **overlap**

Before, the cell above ☰ was empty and the dot read unambiguously as a badge on ☰. Now it sits at a 4-way border junction between `esc`, `⇧tab`, `menu`, `ctrl`, cutting through the `esc` bottom border and attached to nothing in particular. Differential repro: rendered the same page with the pre-PR template injected via `addStyleTag` and `.mk-shift-tab` removed (`.memory/shots/dot-zoom.png` vs `PREPR-dot-zoom.png`). Cosmetic only — a ~0.5×7.5px sliver of the esc hitbox is also shadowed by the dot, which is negligible.

### F2 — Low/Medium: `⇧` (U+21E7) isn't in the bundled webfont, so it always renders from a device-dependent fallback

The key label font is `"Source Sans 3", sans-serif`. Scanning **all 7** bundled `@fontsource/source-sans-3` `400-normal` subsets with fontTools: `U+21E7 found in: NONE`. Confirmed in-browser — the advance width of `⇧` is exactly `11px` under both `13px "Source Sans 3"` and `13px monospace`, i.e. the webfont is never used for this glyph.

Result (see `.memory/shots/stab-key-3x.png`): a hairline outlined arrow beside weight-600 webfont text, over-tall past the cap height, with a wide fallback advance (`⇧` = 11px vs the whole word `tab` = 19px) producing an unintended `⇧ tab` gap. Android's Roboto also lacks U+21E7 (checked Regular/Bold/Black/Italic → `21E7: False`), so Android renders it from Noto Sans Symbols — a third typeface; trimmed font sets render tofu (□).

`☰` is also a fallback glyph (`U+2630` likewise absent from all subsets), but it stands alone on its key so no mismatch is perceivable — here the mismatch is *inside one label*. Every other non-word key in this bar uses an inline SVG (`IconUp`, `IconWordLeft`, …). Suggest an inline SVG matching the existing icon set, or an ASCII label.

### F3 — Low: the two Shift+Tab paths disagree once a modifier is involved

Probe against this revision:

```
keyComboToSequence(shift+tab)       = "\u001b[Z"    ← new, correct
keyComboToSequence(ctrl+shift+tab)  = "\t"          ← Shift silently dropped
keyComboToSequence(alt+shift+tab)   = "\u001b\t"    ← Shift silently dropped
applyArmedModifiers("\x1b[Z", ctrl) -> "\u001b[1;5Z"
applyArmedModifiers("\x1b[Z", alt)  -> "\u001b[1;3Z"
```

`keybinds.ts:223` guards with `shift && !ctrl && !alt`, so a keybound `ctrl+shift+tab` emits a plain Tab while the toolbar path for the same logical combo (arm Ctrl, tap `⇧tab`) emits `CSI 1;5Z`. xterm.js itself (`src/common/input/Keyboard.ts:91-95`) emits `ESC [ Z` for Shift+Tab **and** Ctrl+Shift+Tab, so physical Ctrl+Shift+Tab and the keybind also differ. Pre-existing, but the PR touches this exact branch and the body describes a single canonical path; nothing pins either behavior in tests.

### F4 — Nit: the `CSI … Z` modifier param omits the Shift bit it implies

`keyboard.ts:309` documents `1 + shift(1) + alt(2) + ctrl(4)`, but `keyboard.ts:318` computes `1 + alt*2 + ctrl*4`. BackTab is inherently shifted, so ctrl-armed `⇧tab` emits `CSI 1;5Z` = "Ctrl, no Shift" on a Shift-only key; `1;6Z`/`1;4Z` would be self-consistent, and `keyboard.test.ts:174-178` freezes the current values. **Not a functional bug** — I checked libtermkey/neovim re-adds Shift for final byte `Z` (`src/nvim/tui/termkey/driver-csi.c:625` registers `Z` with `modifier_set = TERMKEY_KEYMOD_SHIFT`, ORed in at `driver-csi.c:25-58`), so nvim reads `1;5Z` as Ctrl+Shift+Tab. Comment vs code just disagree.

### F5 — Nit: stale comment

`styles.css:3207-3209` still says *"The empty top-left corner and the scroll-end / empty top-right simply fall out of the template"*. The top-left corner is now `esc`. `main.tsx:551` was updated for this; the CSS comment wasn't.

### F6 — Nit (accessibility)

`aria-label` was added to `esc`, `⇧tab`, `tab` but not to `ctrl`, `alt`, the arrows, word-jumps, scroll-end, send or ☰, which still rely solely on `title` — functional (title is a name fallback) but now inconsistent; either all or none. WCAG 2.5.3 Label-in-Name is defensible (⇧ ≈ "Shift+"), though voice-control users saying "tap tab" now have two plausible targets. Pre-existing and unchanged: `role="toolbar"` with 13 individually tabbable buttons (APG expects one tab stop + arrow navigation), and the portrait DOM order zig-zags across the two visual rows — landscape DOM order matches visual order exactly, including the new key.

---

### Not exercised

`?mock` renders `MockTerminal`, which never calls `onInputReady`, so `terminalInputRef` is null and toolbar taps are inert; `canAttach` is false so `ctrl`/`alt` arming is a no-op too (`aria-pressed` stayed `false` after tapping `.mk-ctrl`). Byte-level behavior was therefore verified by direct unit probes plus source tracing (`main.tsx:562` → `sendKey` → `applyArmedModifiers` → `onSend` → `terminal.tsx:537 sendRawInput`), **not** end-to-end — I didn't build `gmuxd` for the Playwright harness. Also unobserved: real software-keyboard-open geometry, real `env(safe-area-inset-*)` values, on-device `⇧` fallback rendering, and the scrolled-up state where `end` fills row 1 / col 7 next to the new key (the mock terminal can't scroll, so `terminalScrolledUp` never flips).

**Verdict: AMEND** — F1, F2 and F5 are cheap fixes in the region this PR already touches; F3/F4/F6 are fine as follow-ups. Advisory only; no blocking correctness risk found.

mgabor3141 (commented):
## Delta re-review — PR #480 @ `7f6daeb7299f8f54ef3b2233899ce94d20cad2c1`

No findings.

The added interaction test is an actual browser/daemon test, not a helper-only simulation. It resolves and presses the accessible production BackTab button (`e2e/tests/mobile-shift-tab.spec.ts:32-53`), while the init script intercepts the real `WebSocket.prototype.send` boundary and records the typed-array bytes (`:7-26`). This traverses the production button handler (`apps/gmux-web/src/main.tsx:560-563`), `sendKey` (`:531-535`), App's registered mobile callback (`:783`), and TerminalView's `onInputReady(sendRawInput)` / one `WebSocket.send` (`apps/gmux-web/src/terminal.tsx:537-556`). The exact one-element deep equality at `mobile-shift-tab.spec.ts:55-62` simultaneously rejects wrong bytes, zero sends, duplicates, and hold repetition. The same held pointer gesture exceeds the repeat delay (`:47-53`). Focus, both latch attributes, and actual sidebar-open state are captured from production DOM before/after (`:41-45,55-66`).

Mutation probes confirmed the interaction assertions fail independently when I changed the button to send Tab, duplicated its send, opened the menu, toggled Ctrl, removed toolbar mousedown prevention, or replaced click with repeat handlers. Zero-send rejection follows directly from the same exact-array assertion (an empty array cannot equal the required singleton); no separate production mutation was needed.

Responsive checks execute all CSS branches: 390/320 exercise the base 7×2 branch, 700 exercises `600px <= width < 820px`, and 844 exercises `>=820px` (`mobile-shift-tab.spec.ts:69-125`). They measure actual computed geometry, display state, and computed column count, plus synthetic safe-area insets (`:127-135`). Separate mutations swapping BackTab/Tab in each of the base, medium, and wide `grid-template-areas` all failed at the ordering assertion, proving each branch is non-vacuous. Column-count and word-jump visibility assertions additionally distinguish the three templates.

Validation:
- protocol build: pass
- focused keybind/keyboard unit tests: 67/67 pass
- web TypeScript (`tsc --noEmit`): pass
- Biome lint on delta files: pass
- focused Playwright test: 2/2 pass, including a final clean-head rerun
- six interaction mutation probes and three responsive branch mutation probes: all killed
- final `jj diff --stat`: clean

**Verdict: MERGE**

mgabor3141 (commented):
## Delta re-review — advisory · **AMEND** (mild)

Re-reviewed exactly `7f6daeb7299f8f54ef3b2233899ce94d20cad2c1`. Independent runs at portrait 390×844, landscape 844×390, narrow 320×568 plus 360/414/599/600/667/700/819/1024, synthetic safe areas, and a verbatim mutation harness for the new spec. Report appended to `.memory/review-480-opus.md`; screenshots in `.memory/shots2/`.

### Previous findings — re-verified

| # | Status | Evidence |
|---|---|---|
| **F1** dot/esc collision | **fixed** ✅ | dot now `top:4px right:4px`; at all 11 viewports `dotInsideMenuBtn: true`, `dotOverEscFace/EscBtn/StabFace: false`. Visually clean (`shots2/dot-zoom-new.png`) |
| **F2** `⇧` font fallback | **fixed** ✅ | label is SVG-only (`textContent.length === 0`), `grep -rn '⇧' apps/gmux-web/src` → no hits; icon 20×16, `viewBox 0 0 18 14`, `stroke-width 1.4px`, same `stroke` — byte-identical to `.mk-wl`/`.mk-wr`. No fallback, no tofu |
| **F5** stale CSS comment | **fixed** ✅ | `styles.css:3207` now accurate |
| **F3** keybind/toolbar divergence | unchanged, unadjudicated | re-probed this head: `ctrl+shift+tab → "\t"`, `alt+shift+tab → "\x1b\t"` vs toolbar arms `\x1b[1;5Z` / `\x1b[1;3Z` |
| **F4** missing Shift bit in `1;5Z` | unchanged | `keyboard.ts` untouched |
| **F6** partial `aria-label` | unchanged | only `esc`/`stab`/`tab`; other 10 buttons title-only |

Regression sweep across 11 viewports: grid areas, column counts (7/11/13), bar height (89/44), min target (≥40.3×40), zero overlaps, `overflow −4px`, terminal rows/cols, zero console errors — all identical to the prior head. `getByRole('button', { name: 'Shift+Tab (BackTab)' })` resolves to 1 visible node everywhere; taps don't move `activeElement`; ☰ still opens the sidebar. Synthetic 24/28px insets verified independently at 390/320/844/667 — first/last controls move inside the insets, ordering intact, no clipping.

---

### N2 — Medium (reproduced by mutation): the new spec's status-dot assertions cannot detect the F1 defect they were written for

`e2e/tests/mobile-shift-tab.spec.ts:105-115` measures the dot against **`.mk-esc`'s button rect**, but the visible/hit surface is `.mkey-face`, which by design extends `--key-gap/2` = 2.5px beyond the button on every side.

I ported the spec's `evaluate()` and every assertion verbatim into a mock-app harness (baseline → `ALL SPEC ASSERTIONS PASS`, matching CI) and mutated `styles.css`:

```
M1  dot reverted to top:-3px right:-3px  (the exact F1 defect)  → ALL SPEC ASSERTIONS PASS
M3  'esc stab tab …' → 'esc tab stab …'  (control)              → FAIL: tab.x>stab.x  [portrait, small]
M4  padding-left/right max() removed     (control)              → FAIL: SAFE menu.x>=24, send.right<=width-28
```

Under M1 at 390×844:

```
dot            y 797.0 – 805.0
.mk-esc button y 755.0 – 795.0   →  overlap: false   ← all the spec looks at
.mk-esc face   y 752.5 – 797.5   →  overlap: TRUE    ← the actual defect
```

The spec comment says *"force the existing status state so the badge-vs-Esc regression is exercised too"* and the PR body advertises *"status-dot separation"*. Neither holds — F1 can regress with the suite green. One-selector fix: compare against `.mk-esc .mkey-face` (and `.mk-shift-tab .mkey-face`).

M3/M4 passing/failing correctly means the ordering, column-count and word-jump-visibility assertions **are** load-bearing — that part is good work.

### N3 — Low/Medium (reproduced): the safe-area assertions can't detect loss of the `env()` wiring

`spec:130` overrides `--mobile-safe-area-left/right` with literal px — i.e. it overrides the very properties under test, exercising only `max(4px, var(--x))`:

```
M2  --mobile-safe-area-*: env(safe-area-inset-*, 0px) → 0px literals  → ALL SPEC ASSERTIONS PASS
```

The mechanism that actually ships is unprotected; only the testability seam is. I did confirm the production chain is valid CSS and degrades correctly — with `env()` in place and zero insets, computed `padding-left/right` = `4px` at every viewport, byte-identical to the pre-PR `padding: 0 4px`. (FWIW `env(safe-area-inset-left, 24px)` can't fake an inset: Chromium supports the variable and resolves it to `0px`, so the fallback never fires. Asserting on the computed `padding-left` string is the cheap way to cover this.)

### N1 — Low/Medium: the new BackTab icon is non-standard and collides visually with both word-jump icons

`main.tsx:459-461` now puts three bar+arrow glyphs in one toolbar:

```
IconBackTab   bar RIGHT (x=14.5) + arrow LEFT   →  ←|
IconWordLeft  bar LEFT  (x=3.5)  + arrow LEFT   →  |←
IconWordRight bar RIGHT (x=14.5) + arrow RIGHT  →  →|
```

`←|` vs `→|` differ only in arrow direction (identical right bar); `←|` vs `|←` only in which side the bar is on. In the ≥820px single-row landscape layout **all three are visible at once** (`shots2/glyph-compare.png`). The conventional BackTab glyph `⇤` (U+21E4, *leftwards arrow **to** bar*) is bar-on-the-left — already taken by word-left here — so flipping the bar right avoids the clash but reads backwards and mirrors word-right. It also drops the visual tie to the adjacent `tab` **word** key: `esc`/`tab`/`ctrl`/`alt` are words, `⇧tab` was self-describing, and on touch there's no hover title, so the meaning is now only discoverable via a screen reader. A word-family label the bundled webfont covers (`s-tab`, `sh+tab`) would fix F2 without inventing a glyph. Maintainer call — flagging so it isn't decided silently.

### N4 — Low: the dot fix also moves the badge on the two non-bar MenuButton placements

`.mobile-bottom-action.menu-btn.bg-waiting::after` (`styles.css:3377`) matches all three variants, since `menu-button.tsx:35` puts those classes on `bar`, `footer` and `floating` alike. Measured on the home screen: floating button 48×38 at (10, 796), badge moved from overhanging `(-3,-3)` to inside `(4,4)` (`shots2/home-NEW-menubtn.png` vs `home-OLDDOT-menubtn.png`). So the overhang badge idiom is gone on home / dead-session footer; the adjacent comment *"Ring in the page background so the dot reads cleanly against the button border"* is now inaccurate (on the floating variant the ring is drawn against the button's own `--bg-surface`); and the new rationale comment is bar-specific ("now that Esc sits directly above it") while the selector is not. A `.mk-menu`-scoped override would have contained it. Not obviously worse on home — the unmentioned, untested blast radius is the finding. The `footer` variant isn't reachable under `?mock`, so it's unverified.

### N5 — Nit (probable, not reproduced): flake risk in the byte assertion

`spec:60` asserts `result.sent` `toEqual([[0x1b,0x5b,0x5a]])` after a ~800ms hold. The same socket carries `{"type":"resize",…}` (`terminal.tsx:146`) and the patch records **every** `send` (it only suppresses *forwarding* for exact `ESC [ Z`). A late refit in that window appends a second entry and fails exact equality; `playwright.config.ts:5` sets `retries: 0`, so that's a hard CI failure. Filtering to input-shaped frames would decouple it. Not reproduced — the harness needs a built `gmuxd`, which I didn't build.

Credit where due on that test: the 700ms hold **does** meaningfully cover the no-repeat claim, since `REPEAT_DELAY_MS = 300` (`main.tsx:469`) and `useAutoRepeat` binds `onPointerDown` (`main.tsx:485`), which `mouse.down()` dispatches. Two caveats worth naming: it's a client-side *send* assertion (the patch swallows the bytes, so the PTY never receives BackTab — deliberate per its comment), and the focus assertion would pass vacuously if the `.xterm-helper-textarea` focus call were a no-op.

### Other observations

* Synthetic 320×568 + 24/28px insets shrinks columns to **34px**, below the 40px design target, with no minimum-width assertion. Real devices with side insets are ≥568px wide (≥51px columns at 844 with 59px insets) — a synthetic artefact of the test's own viewport choice, not a shipping defect.
* Only the *bar* is inset horizontally; the terminal still runs under the cutout (`shots2/*-safearea-full.png`). Pre-existing and out of scope, but "reserves horizontal safe-area insets" is toolbar-only.
* `padding` shorthand → longhands is behaviour-preserving: with a synthetic 34px bottom inset, computed `padding-bottom` = `38px`, bar height 89 → 123, bar still flush to the viewport bottom.

---

## Verdict: **AMEND** (mild)

The product code is materially better than at the previous head: **F1 and F2 are genuinely and cleanly fixed** (verified at 11 viewports with pixel-level icon parity), F5 is fixed, side-inset handling is a real improvement, and a real browser regression test now exists where there was none. Nothing here is a correctness or accessibility regression, and the ordering/column/word-jump assertions are mutation-proven load-bearing.

What keeps it off MERGE: two of the new spec's advertised protections are provably vacuous — mutation M1 restores the exact defect I reported last round with the suite green, while the PR body claims "status-dot separation" is asserted — plus the open design question in N1. Must-fix is small: point the badge assertions at `.mkey-face` (N2), and stop the safe-area test overriding the property it tests (N3). N1 needs a maintainer call; N4/N5 and the carried-over F3/F4/F6 are fine as follow-ups.

mgabor3141 (commented):
## Final focused re-review — advisory · **MERGE**

Reviewed exactly `cad197ce2a32166ffb644e03ae82bbc5a39eda79`, scoped to my five outstanding findings. Report appended to `.memory/review-480-opus.md`; visuals in `.memory/shots3/`.

All five are fixed, and the two that were *test vacuity* are now proven load-bearing by mutation rather than by assertion text.

### N4 — mobile-only badge scope ✅

Base rule restored to `top:-3px; right:-3px`; the inset position is now a bar-scoped override (`styles.css:3392-3398`). Measured at 390×844:

| variant | computed `top`/`right` | inside button | overlaps `.mk-esc .mkey-face` |
|---|---|---|---|
| `mk-menu` (bar) | `4px` / `4px` | **true** | **false** |
| `menu-btn-floating` (home) | `-3px` / `-3px` | false — overhang restored | n/a |

The floating badge is back to the established idiom (`shots3/badge-floating.png`). The `footer` variant falls through to the base rule and isn't inside `.mobile-bottom-bar`, so it's a no-op by construction. The bar-specific rationale comment now sits on the bar-specific rule.

### N2 — actual face-overlap mutation ✅ (mutation-proven)

`spec:92,113-114` now runs the clearance disjunction against `get('.mk-esc .mkey-face')`. Re-ran my verbatim assertion port (baseline `ALL SPEC ASSERTIONS PASS`) with the fix removed:

```
M1  delete the bar-scoped ::after override  (restores the exact F1 defect)
    → FAIL: [portrait] DOT-CLEAR-OF-ESC-FACE
            [medium]   DOT-CLEAR-OF-ESC-FACE
            [wide]     DOT-CLEAR-OF-ESC-FACE
            [small]    DOT-CLEAR-OF-ESC-FACE
```

M1 previously passed in all four branches; it now fails in all four — including `medium`/`wide`, where the single-row layout puts `esc` immediately right of `menu` and the overhang crosses its face there too.

### N3 — production safe-area wiring ✅ (mutation-proven, incl. an axis typo)

`spec:129-151` adds a CSSOM contract (walks stylesheets, recurses into `@media`) requiring both custom properties to contain `env(safe-area-inset-left` / `-right`, checked **before** the synthetic override:

```
M2   env(…) → 0px literals            → FAIL: CONTRACT left "0px", CONTRACT right "0px"
M2b  safe-area-inset-left → …-top     → FAIL: CONTRACT left "env(safe-area-inset-top, 0px)"
M2c  drop only the right declaration  → FAIL: CONTRACT right "0px"
```

M2b is the strongest result: an axis typo resolves to `0px` on every device, so **no** geometric assertion could ever catch it — the contract does. Earlier controls still hold (M3 order swap, M4 padding removal both fail). The comment honestly states why the split exists.

### N5 — terminal-frame filtering ✅

The patch now records only binary frames and skips JSON strings. I audited every send site — this is the protocol's own boundary, not a heuristic:

```
terminal.tsx:146   resize            JSON string  → excluded
presence.ts:60,67,98,105  hello/state/perm  JSON string  → excluded
terminal.tsx:545   raw input         binary       → recorded   ← the ONLY binary send in the app
```

Replicating the patch verbatim against those five real shapes: `terminalInputs = [[27,91,90],[9]]`, while all three JSON frames are forwarded untouched and unrecorded, BackTab is recorded and withheld from the fixture shell, Tab is recorded and forwarded. Worth noting the old version also captured `presence.ts` traffic (event-driven `client-state`), so the flake surface was wider than the resize frame I originally named — and is now fully closed.

### N1 — BackTab icon distinctness ✅ (one residual nit)

`main.tsx:459` now composes a shift up-arrow marker + bar at `x=10` + left arrow. Two real improvements: **direction is conventional** (bar left, arrow left into it — matching `⇤`, no longer mirroring `IconWordRight`), and **the shift marker disambiguates it from `IconWordLeft`** — same `|←` core, but a strict superset with an unmistakable shift cue, and both legitimately mean "leftward to a boundary". Legible at the 40px narrow key (`shots3/icon-compare.png`). Optical centering is within the set's existing tolerance (−0.89px/−0.44px vs word-left's −0.84px/0px).

**Residual nit: strokes render ~20% lighter than every neighbour.** The viewBox is now larger than the rendered box, so `preserveAspectRatio` scales it *down* while the others scale *up*:

| icon | viewBox | rendered | uniform scale | stroke on screen | ink |
|---|---|---|---|---|---|
| `.mk-shift-tab` | 24×18 | 22×16 | **0.889** | **1.244px** | 14.22×11.56 |
| `.mk-wl` / `.mk-wr` | 18×14 | 20×16 | 1.111 | 1.556px | 10.56×8.89 |
| `.mk-al` / `.mk-au` | 14×14 | 16×16 | 1.143 | 1.600px | 6.86×6.86 |

Visible in the landscape bar: the BackTab glyph reads finer than the adjacent `|←` while being ~35% wider and ~30% taller. Same category as F2, far milder. One-token fix: `width="24" height="18"`, or `stroke-width` ≈ 1.75 on this icon. Not worth blocking on.

### Regression sweep (11 viewports, this head)

`cols` 7/11/13 at the right breakpoints, `barH` 89/44, `minW` ≥40.3, zero overlaps, `overflow −4px`, `dotInsideMenuBtn: true` and `dotOverEscFace/dotOverStabFace: false` everywhere, icon box constant 22×16, terminal rows/cols unchanged, zero console errors. Final visuals (`shots3/final-*`): portrait has `esc` above `☰` with the badge tucked inside and clear of `esc`; landscape reads `☰ esc ↑|← tab ctrl alt |← ← ↑ ↓ → →| ▷` with no clipping.

### Carried over, not blocking

F3 (`ctrl+shift+tab` → `\t` while the toolbar arm gives `\x1b[1;5Z`), F4 (`1;5Z` omits the implied Shift bit — benign per libtermkey), F6 (`aria-label` on 3 of 13 buttons) — unchanged and still unadjudicated. Fine as follow-ups.

---

## Verdict: **MERGE**

The badge fix is scoped to the mobile bar with the other two MenuButton placements provably untouched; the face-overlap and env-wiring assertions are now mutation-proven (including an axis typo no geometric test could see); the byte capture keys on the protocol's own string/binary boundary, and I verified the app has exactly one binary send site; the icon is both conventional and distinct. The remaining stroke-weight nit is cosmetic — land it or fix it in passing. Nice iteration.

PR #484, opened 2026-08-12
fix: preserve terminal state across reconnect

mgabor3141:
## Summary

Amended after review: the original shared `?1049l/h` checkpoint bytes were removed because `gmux attach` consumes the same raw stream. Browser-only buffer metadata now travels separately (`?client=browser`), and `TerminalView` applies buffer selection only after a complete BSU/ESU checkpoint is staged.

The browser no longer resets xterm before a replacement socket opens. Failed reconnect attempts retain the last committed screen. The narrow demonstrated fix resets only SGR at checkpoint application, preserving existing terminal modes and saved buffers rather than claiming a full structured checkpoint. `TerminalIO` now fences in-flight xterm callbacks across epochs.

The switch E2E launches disposable A/B sessions during the test, leaving the shared harness session untouched. It covers A→B→A isolation, different geometry, B normal→alternate→reconnect→exit, post-reconnect output, and failed-handshake screen preservation.

## Validation

- Full disposable real-daemon Playwright: `pnpm test:e2e` — 50 passed
- `cd cli/gmux && go test ./...` — pass
- `cd services/gmuxd && go test ./...` — pass
- `cd cli/gmux && go test -race ./internal/ptyserver -run TestSnapshotFrame -count=1` — pass
- `pnpm --dir apps/gmux-web lint` — pass; 542 web tests pass, with the existing unrelated `@gmux/protocol` package-entry failure in `src/store.test.ts`

Dead ReplayView truncation remains explicitly out of scope. No daemon was installed/restarted and no live state was touched. Not merged.

## Remediation delta (cc80c593)

- `term.reset()` now runs only on an actual session-ID transition; failed same-session reconnects retain the last committed screen.
- Browser checkpoint metadata is forwarded on local and peer paths using a strict `client=browser` allowlist; duplicate, auth, and routing query parameters are not forwarded.
- Browser replay applies SGR reset before buffer selection/erase and omits `CSI r` so same-session margins are retained. Metadata absence is non-destructive; session switches still isolate xterm modes, buffers, scrollback, and input encoding.
- The real-daemon E2E asserts metadata arrival, A→B isolation, normal→alternate and alternate→normal behavior, mode/input isolation, failed-handshake retention, and `baseY === viewportY` after replay.

Verification: full Playwright **50 passed**; focused switch E2E passed; Go ptyserver/wsproxy/apiclient tests and race tests passed; web lint/build/typecheck passed. Web Vitest had 542 passing tests plus the pre-existing unrelated `@gmux/protocol` package-entry failure in `store.test.ts`. This PR does not claim complete serialization of arbitrary DEC modes or renderer state.

## Final O1 remediation

The browser-only checkpoint metadata now includes authoritative 1-based inclusive DECSTBM margins and rows. Inside the synchronized checkpoint, the browser applies SGR reset, selects the authoritative buffer, resets margins before erase/repaint, emits the rendered screen and cursor state, then restores the exact margin region before ESU/live bytes. The raw CLI/legacy frame remains unchanged and receives no browser metadata or controls. Metadata absence remains a safe fallback; peer forwarding carries only the allowlisted browser capability and the metadata is opaque WS payload.

The real-daemon switch test now launches a disposable C session, sets rows 2..4 after browser geometry settles, fills 40 distinguishable lines, reconnects, asserts all content and exact margins, emits subsequent output, and asserts the region remains active. It does not claim preservation of an intentionally scrolled viewport; existing replay assertions only require convergence to bottom (`viewportY === baseY`).

## Final O1 remediation (8097815c)

Browser checkpoint metadata now includes authoritative 1-based inclusive DECSTBM margins and rows. Within synchronized replay, browser application performs SGR reset, buffer selection, `CSI r` before erase/repaint, cursor state, and exact margin restoration before live bytes resume. Raw CLI/legacy frames remain unchanged. Metadata absence remains safe and non-destructive; peer forwarding remains an allowlisted opaque capability path.

The real-daemon E2E now launches disposable C, sets rows 2..4 after geometry settles, fills 40 distinguishable lines, reconnects, verifies all content and exact margins, emits subsequent output, and verifies the region remains active. Replay assertions claim only bottom convergence (`viewportY === baseY`), not preservation of an intentionally scrolled viewport. Both reset-before-paint and restore-after-paint mutation probes fail deterministically.

## Final O2 remediation (46b51fb3)

The browser-only DECSTBM restore now runs **before** the raw checkpoint's anchored CUP + cursor-visibility tail. Final browser ordering is: SGR reset → authoritative buffer select → margin reset → repaint → authoritative margin restore → authoritative cursor position/visibility → ESU → live PTY bytes. The raw CLI/legacy binary frame is unchanged.

The real-daemon C regression keeps the 40-line O1 content and rows 2..4 subsequent-scrolling checks, parks the PTY cursor at exact xterm coordinates `(19,8)`, reconnects, and asserts exact margins and cursor. The next PTY marker must begin at column 19 on visible row 8 while row 0 remains byte-for-byte unchanged. A's weakened cursor checks are restored to exact `(5,2)`. A mutation moving margin restore behind CUP fails with cursor `(0,0)` instead of `(19,8)`.

Final validation: focused Playwright **1 passed**, full Playwright **50 passed**; web Vitest **712 passed**, lint/typecheck/build pass; root CI build/test entrypoints pass; CLI and gmuxd full Go suites pass; ptyserver race passes; focused raw CLI stream, normal/alternate margin, peer metadata, wsproxy, and apiclient tests pass. After push, all **11/11 GitHub checks passed**. No daemon was installed or restarted; PR remains unmerged.

### DIFF
--- a/apps/gmux-web/src/terminal-io.test.ts
+++ b/apps/gmux-web/src/terminal-io.test.ts
@@ -278,12 +278,31 @@ describe('createTerminalIO', () => {
     h.io.reset(2)
     h.io.enqueue(enc('fresh'), 2)
 
-    expect(h.writes).toEqual(['stale', 'fresh'])
+    // The stale write was already handed to xterm. The new epoch must wait
+    // for its callback instead of overlapping the parser or resize path.
+    expect(h.writes).toEqual(['stale'])
     h.flushOne()
 
     expect(onWritten).not.toHaveBeenCalled()
     expect(h.writes).toEqual(['stale', 'fresh'])
     expect(h.resizes).toEqual([])
+  })
+
+  it('does not let an old completion advance a new epoch past a resize fence', () => {
+    const h = makeHarness()
+    h.io.reset(1)
+    h.io.enqueue(enc('old'), 1)
+    h.io.reset(2)
+    h.io.enqueue(enc('new'), 2)
+    h.io.requestResize({ cols: 80, rows: 24 }, 2)
+
+    expect(h.writes).toEqual(['old'])
+    expect(h.resizes).toEqual([])
+    h.flushOne()
+    expect(h.writes).toEqual(['old', 'new'])
+    expect(h.resizes).toEqual([])
+    h.flushOne()
+    expect(h.resizes).toEqual([{ cols: 80, rows: 24 }])
   })
 
   it('runs completion callback after the final chunk in enqueueMany', () => {
--- a/apps/gmux-web/src/terminal-io.ts
+++ b/apps/gmux-web/src/terminal-io.ts
@@ -271,7 +271,10 @@ export function createTerminalIO(term: TerminalWriter, scroll?: ScrollAccessor):
     reset(epoch: number) {
       currentEpoch = epoch
       queue = []
-      writeInFlight = false
+      // Keep writeInFlight true until the old xterm callback arrives. A
+      // reset can invalidate queued work, but it cannot cancel xterm's
+      // parser callback; clearing this flag here would let a new resize or
+      // write overlap the still-running old parse.
       pendingResize = null
       savedScroll = null
       bufferReset = false
--- a/apps/gmux-web/src/terminal.tsx
+++ b/apps/gmux-web/src/terminal.tsx
@@ -11,7 +11,7 @@ import { applyArmedModifiers, attachKeyboardHandler, attachPasteHandler, default
 import { DEFAULT_THEME_COLORS, type ResolvedKeybind } from './config'
 import { attachMobileInputHandler } from './mobile-input'
 import { isTouchDevice } from './touch'
-import { BSU, createReplayBuffer, startsWith } from './replay'
+import { BSU, createReplayBuffer, ESU, startsWith } from './replay'
 import { createTerminalIO, type TerminalSize } from './terminal-io'
 import { type LinkInfo, linkAtPoint, openLinkAtPoint } from './terminal-link'
 import { createLongPressRecognizer } from './long-press'
@@ -29,12 +29,13 @@ import type { Session } from './types'
 
 const encoder = new TextEncoder()
 const CHECKPOINT_WIRE_RESET = encoder.encode('\x1b[r\x1b[H\x1b[2J\x1b[3J')
-// CSI r is part of the shared raw attach frame for legacy CLI consumers, but
-// it is not needed to repaint a browser checkpoint. Omitting it lets a
-// same-session reconnect retain the PTY's current margins; a session switch
-// has already reset xterm and therefore starts with default margins.
-const CHECKPOINT_BROWSER_RESET = encoder.encode('\x1b[H\x1b[2J\x1b[3J')
+// The browser checkpoint resets margins before repaint and restores the
+// authoritative region before the cursor-position/visibility tail. The raw
+// frame remains unchanged for CLI and legacy consumers.
+const CHECKPOINT_BROWSER_RESET = encoder.encode('\x1b[r\x1b[H\x1b[2J\x1b[3J')
 const CHECKPOINT_SGR_RESET = encoder.encode('\x1b[0m')
+
+type CheckpointMargins = { top: number, bottom: number, rows: number }
 
 /**
  * Add browser-only buffer authority at the complete checkpoint boundary.
@@ -42,11 +43,39 @@ const CHECKPOINT_SGR_RESET = encoder.encode('\x1b[0m')
  * ?1049h/l in it would change an operator's terminal. Reordering the known
  * SGR is emitted before buffer selection and erase, preventing the old
  * background from coloring blank cells. The browser-specific transform also
- * omits the shared frame's CSI r: same-session reconnects retain margins,
- * while session-ID changes perform the full xterm reset. No other terminal
- * state is claimed to be serialized by this rendered checkpoint.
+ * resets margins before repaint and restores the emulator's authoritative
+ * 1-based region before the frame's final cursor-position/visibility tail.
+ * The raw frame's CSI r is never changed, and no other terminal state is
+ * claimed to be serialized.
  */
-function prepareBrowserCheckpoint(chunks: Uint8Array[], activeAlternate: boolean | null): Uint8Array[] {
+function cursorTailOffset(frameBody: Uint8Array): number | null {
+  // The shared frame ends with CUP + DECTCEM + ESU. Find CUP only within that
+  // anchored tail so escape sequences in rendered terminal content cannot be
+  // mistaken for the authoritative cursor position.
+  const visibilityLength = 6 // CSI ? 25 h/l
+  const visibilityOffset = frameBody.length - ESU.length - visibilityLength
+  if (visibilityOffset <= 0 || !startsWith(frameBody.slice(frameBody.length - ESU.length), ESU)) return null
+  const visibility = frameBody.slice(visibilityOffset, visibilityOffset + visibilityLength)
+  if (!(visibility[0] === 0x1b && visibility[1] === 0x5b && visibility[2] === 0x3f
+    && visibility[3] === 0x32 && visibility[4] === 0x35 && (visibility[5] === 0x68 || visibility[5] === 0x6c))) return null
+
+  const cursorEnd = visibilityOffset
+  if (frameBody[cursorEnd - 1] !== 0x48) return null // H
+  for (let i = cursorEnd - 2; i >= Math.max(0, cursorEnd - 24); i--) {
+    if (frameBody[i] !== 0x1b) continue
+    if (frameBody[i + 1] !== 0x5b) return null
+    let semicolons = 0
+    for (let j = i + 2; j < cursorEnd - 1; j++) {
+      const byte = frameBody[j]
+      if (byte === 0x3b) semicolons++
+      else if (byte < 0x30 || byte > 0x39) return null
+    }
+    return semicolons === 1 ? i : null
+  }
+  return null
+}
+
+function prepareBrowserCheckpoint(chunks: Uint8Array[], activeAlternate: boolean | null, margins: CheckpointMargins | null): Uint8Array[] {
   if (chunks.length === 0) return chunks
   const first = chunks[0]
   if (!startsWith(first, BSU)) {
@@ -58,25 +87,43 @@ function prepareBrowserCheckpoint(chunks: Uint8Array[], activeAlternate: boolean
 
   const hasReset = first.length >= BSU.length + CHECKPOINT_WIRE_RESET.length
     && startsWith(first.slice(BSU.length), CHECKPOINT_WIRE_RESET)
-  if (!hasReset) {
+  // Older runners provide neither authoritative buffer nor margins. Keep the
+  // legacy fallback non-destructive rather than guessing a state checkpoint.
+  const validMargins = margins !== null && margins.rows > 0
+    && margins.top >= 1 && margins.bottom <= margins.rows
+    && (margins.top < margins.bottom || (margins.rows === 1 && margins.top === 1 && margins.bottom === 1))
+  if (!hasReset || activeAlternate === null || !validMargins) {
     const prepared = new Uint8Array(CHECKPOINT_SGR_RESET.length + first.length)
     prepared.set(CHECKPOINT_SGR_RESET)
     prepared.set(first, CHECKPOINT_SGR_RESET.length)
     return [prepared, ...chunks.slice(1)]
   }
 
-  const buffer = activeAlternate === null ? null : encoder.encode(activeAlternate ? '\x1b[?1049h' : '\x1b[?1049l')
+  const buffer = encoder.encode(activeAlternate ? '\x1b[?1049h' : '\x1b[?1049l')
+  // DECSTBM rejects equal bounds; on a one-row terminal the only safe
+  // representation is the default full-screen reset itself.
+  const restoreMargins = margins.rows === 1
+    ? encoder.encode('\x1b[r')
+    : encoder.encode(`\x1b[${margins.top};${margins.bottom}r`)
   const body = first.slice(BSU.length + CHECKPOINT_WIRE_RESET.length)
-  const bufferLength = buffer?.length ?? 0
-  const prepared = new Uint8Array(BSU.length + CHECKPOINT_SGR_RESET.length + bufferLength + CHECKPOINT_BROWSER_RESET.length + body.length)
+  const tailOffset = cursorTailOffset(body)
+  if (tailOffset === null) {
+    const prepared = new Uint8Array(CHECKPOINT_SGR_RESET.length + first.length)
+    prepared.set(CHECKPOINT_SGR_RESET)
+    prepared.set(first, CHECKPOINT_SGR_RESET.length)
+    return [prepared, ...chunks.slice(1)]
+  }
+  const repaint = body.slice(0, tailOffset)
+  const cursorTail = body.slice(tailOffset)
+  const prepared = new Uint8Array(BSU.length + CHECKPOINT_SGR_RESET.length + buffer.length + CHECKPOINT_BROWSER_RESET.length + repaint.length + restoreMargins.length + cursorTail.length)
   let offset = 0
   prepared.set(BSU, offset); offset += BSU.length
   prepared.set(CHECKPOINT_SGR_RESET, offset); offset += CHECKPOINT_SGR_RESET.length
-  if (buffer) {
-    prepared.set(buffer, offset); offset += buffer.length
-  }
+  prepared.set(buffer, offset); offset += buffer.length
   prepared.set(CHECKPOINT_BROWSER_RESET, offset); offset += CHECKPOINT_BROWSER_RESET.length
-  prepared.set(body, offset)
+  prepared.set(repaint, offset); offset += repaint.length
+  prepared.set(restoreMargins, offset); offset += restoreMargins.length
+  prepared.set(cursorTail, offset)
   return [prepared, ...chunks.slice(1)]
 }
 
@@ -943,8 +990,9 @@ export function TerminalView({
       // that stream. Apply it only once the complete BSU/ESU checkpoint is
       // staged; the existing screen remains visible during failed attempts.
       let checkpointAlt: boolean | null = null
+      let checkpointMargins: CheckpointMargins | null = null
       const replay = createReplayBuffer((chunks) => {
-        const prepared = prepareBrowserCheckpoint(chunks, checkpointAlt)
+        const prepared = prepareBrowserCheckpoint(chunks, checkpointAlt, checkpointMargins)
         queueMany(prepared, () => {
           termRef.current?.scrollToBottom()
           setTermLoading(false)
@@ -999,6 +1047,9 @@ export function TerminalView({
             // Use it to initialize ptySize if we don't have one yet.
             if (msg.type === 'terminal_checkpoint') {
               checkpointAlt = msg.active_buffer === 'alternate'
+              if (Number.isInteger(msg.scroll_top) && Number.isInteger(msg.scroll_bottom) && Number.isInteger(msg.rows)) {
+                checkpointMargins = { top: msg.scroll_top, bottom: msg.scroll_bottom, rows: msg.rows }
+              }
               return
             }
 
--- a/apps/gmux-web/src/terminal.tsx
+++ b/apps/gmux-web/src/terminal.tsx
@@ -28,16 +28,23 @@ import type { Session } from './types'
 // ── Config ──
 
 const encoder = new TextEncoder()
-const CHECKPOINT_RESET = encoder.encode('\x1b[r\x1b[H\x1b[2J\x1b[3J')
+const CHECKPOINT_WIRE_RESET = encoder.encode('\x1b[r\x1b[H\x1b[2J\x1b[3J')
+// CSI r is part of the shared raw attach frame for legacy CLI consumers, but
+// it is not needed to repaint a browser checkpoint. Omitting it lets a
+// same-session reconnect retain the PTY's current margins; a session switch
+// has already reset xterm and therefore starts with default margins.
+const CHECKPOINT_BROWSER_RESET = encoder.encode('\x1b[H\x1b[2J\x1b[3J')
 const CHECKPOINT_SGR_RESET = encoder.encode('\x1b[0m')
 
 /**
  * Add browser-only buffer authority at the complete checkpoint boundary.
  * The binary frame itself is also consumed by `gmux attach`, so putting
  * ?1049h/l in it would change an operator's terminal. Reordering the known
- * reset prefix means a session switch selects the target buffer before
- * clearing/rendering it, while a reconnect in alternate mode leaves the
- * saved normal buffer intact. Older runners still get the narrow SGR reset.
+ * SGR is emitted before buffer selection and erase, preventing the old
+ * background from coloring blank cells. The browser-specific transform also
+ * omits the shared frame's CSI r: same-session reconnects retain margins,
+ * while session-ID changes perform the full xterm reset. No other terminal
+ * state is claimed to be serialized by this rendered checkpoint.
  */
 function prepareBrowserCheckpoint(chunks: Uint8Array[], activeAlternate: boolean | null): Uint8Array[] {
   if (chunks.length === 0) return chunks
@@ -49,23 +56,26 @@ function prepareBrowserCheckpoint(chunks: Uint8Array[], activeAlternate: boolean
     return [prepared, ...chunks.slice(1)]
   }
 
-  const hasReset = first.length >= BSU.length + CHECKPOINT_RESET.length
-    && startsWith(first.slice(BSU.length), CHECKPOINT_RESET)
-  if (!hasReset || activeAlternate === null) {
+  const hasReset = first.length >= BSU.length + CHECKPOINT_WIRE_RESET.length
+    && startsWith(first.slice(BSU.length), CHECKPOINT_WIRE_RESET)
+  if (!hasReset) {
     const prepared = new Uint8Array(CHECKPOINT_SGR_RESET.length + first.length)
     prepared.set(CHECKPOINT_SGR_RESET)
     prepared.set(first, CHECKPOINT_SGR_RESET.length)
     return [prepared, ...chunks.slice(1)]
   }
 
-  const buffer = encoder.encode(activeAlternate ? '\x1b[?1049h' : '\x1b[?1049l')
-  const body = first.slice(BSU.length + CHECKPOINT_RESET.length)
-  const prepared = new Uint8Array(BSU.length + buffer.length + CHECKPOINT_RESET.length + CHECKPOINT_SGR_RESET.length + body.length)
+  const buffer = activeAlternate === null ? null : encoder.encode(activeAlternate ? '\x1b[?1049h' : '\x1b[?1049l')
+  const body = first.slice(BSU.length + CHECKPOINT_WIRE_RESET.length)
+  const bufferLength = buffer?.length ?? 0
+  const prepared = new Uint8Array(BSU.length + CHECKPOINT_SGR_RESET.length + bufferLength + CHECKPOINT_BROWSER_RESET.length + body.length)
   let offset = 0
   prepared.set(BSU, offset); offset += BSU.length
-  prepared.set(buffer, offset); offset += buffer.length
-  prepared.set(CHECKPOINT_RESET, offset); offset += CHECKPOINT_RESET.length
   prepared.set(CHECKPOINT_SGR_RESET, offset); offset += CHECKPOINT_SGR_RESET.length
+  if (buffer) {
+    prepared.set(buffer, offset); offset += buffer.length
+  }
+  prepared.set(CHECKPOINT_BROWSER_RESET, offset); offset += CHECKPOINT_BROWSER_RESET.length
   prepared.set(body, offset)
   return [prepared, ...chunks.slice(1)]
 }
@@ -286,6 +296,10 @@ export function TerminalView({
   const reconnectTimer = useRef<ReturnType<typeof setTimeout> | null>(null)
   const disposed = useRef(false)
   const currentSessionId = useRef(session.id)
+  // The WebSocket effect can be re-run by unrelated dependency changes. Keep
+  // the reset tied to an observed session-ID transition, not to effect
+  // lifetime or to a reconnect attempt.
+  const terminalSessionId = useRef<string | null>(null)
   const sessionRef = useRef(session)
   const ctrlArmedRef = useRef(ctrlArmed)
   const altArmedRef = useRef(altArmed)
@@ -889,6 +903,15 @@ export function TerminalView({
     const epoch = termEpochRef.current + 1
     termEpochRef.current = epoch
     termIoRef.current.reset(epoch)
+
+    const sessionChanged = terminalSessionId.current !== null
+      && terminalSessionId.current !== session.id
+    terminalSessionId.current = session.id
+    // A session change is the one boundary where the existing xterm state is
+    // not authoritative. Reset only on that actual ID transition:
+    // same-session reconnect attempts retain the last committed screen and
+    // parser state during the outage.
+    if (sessionChanged) termRef.current.reset()
 
     // Reset sizes so stale values from a previous session can't trigger a
     // spurious pill while the loading overlay is visible (before ws.onopen).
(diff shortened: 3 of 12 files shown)


mgabor3141:
**Advisory — MEDIUM — `apps/gmux-web/src/terminal.tsx:875`: reconnect attempts erase the last good terminal before a replacement socket has opened or delivered its checkpoint.**

At exact head `3d72aa74`, `connect()` increments the TerminalIO epoch and calls both `termIo.reset()` and `term.reset()` whenever `isFirstConnect` is false, *before* constructing/opening the replacement WebSocket. `isFirstConnect` becomes false after the first successful `onopen`, so this schedule is deterministic:

1. Connect successfully and render a non-empty terminal.
2. Make subsequent `/ws/<id>` opens fail (offline mode, or abort the WS handshake), then close the current socket.
3. After the 500 ms retry timer invokes `connect()`, xterm is reset to an empty screen.
4. Because no replacement snapshot can arrive, the useful last-known terminal remains blank; rapid retries reset it again. A failed initial open does not do this (the flag is still true), demonstrating that the flag means “this effect has opened once,” not “a reconnect checkpoint is ready.”

Expected: retain the old rendered state while disconnected/while opens fail, and reset only when the current replacement connection’s snapshot is ready to be applied (with the reset serialized against TerminalIO writes). Actual: an attempted reconnect destroys state before success is known. This also makes the disconnected pill less useful because the content beneath it disappears.

The new E2E only closes a socket while the server remains available and waits for the pill to disappear, so it cannot catch the failed-open schedule. The Go ordering assertion itself is non-vacuous: baseline passed, and reversing its `<` ordering predicate made `TestSnapshotFrameRestoresActiveBufferAndOrdersCheckpoint` fail; however, it does not cover browser reset timing.

**Verdict: AMEND.** Move/gate the reconnect reset at successful snapshot application (without allowing fresh snapshot/live bytes to precede or race it), and add a deterministic failed-open/retry test asserting the old screen remains until a checkpoint arrives.

mgabor3141 (commented):
Adversarial anchor review at exact head `3d72aa74`. Independent repro in a disposable workspace with a real daemon, real `gmux attach` under a pty/tmux, and a real browser. Verdict: **AMEND** — the server-side buffer-authority idea is right, but the two new behaviors each destroy something reconnect-without-reset preserved, and the PR is red for reasons I reproduced locally.

## High

**H1 — `\x1b[?1049h` now goes into the operator's real terminal and is never undone.**
`cli/gmux/cmd/gmux/attach.go:66-115` feeds the same WS checkpoint straight to the local tty, and `cli/gmux/internal/localterm/localterm.go:148-165` (`Detach`) emits only `\x1b[?1004l` — no `?1049l`. So `gmux attach` to any alt-screen session leaves the terminal stuck in the alternate screen once the attach ends.

Repro (disposable daemon, real binaries, tmux as the user's terminal):
```
session: bash -c 'printf "\033[?1049h\033[2J\033[H\033[44mALT-EXIT\033[0m"; sleep 20'
tmux 80x30 → gmux attach <id> → session exits → attach returns
during attach:      alt=1
after session exit: alt=1     ← never restored
pane: "ALT-EXITbash-5.3$ echo BACK-IN-SHELL"  (prompt on the alt screen; user's scrollback gone until `tput rmcup`)
```
Raw capture of the frame: `^[[?1004h^[[?2026h^[[r^[[H^[[2J^[[3J^[[?1049h^[[44mALT-CONTENT…`, zero `?1049l` in the entire stream. Impossible pre-PR (frame had no buffer byte).
Same root, normal-buffer case: `\x1b[?1049l` is emitted *after* `\x1b[H\x1b[2J\x1b[3J` and *before* the rendered bytes; DECRST 1049 is a DECRC-style restore (xterm.js `src/common/InputHandler.ts:2257-2272` → `restoreCursor()` restores `savedX/savedY`, `savedCurAttrData`, origin/wraparound, charsets). Harmless in the browser only because `terminal.tsx:853` reset zeroes that state first; the CLI client never resets, and the frame carries no leading `\x1b[0m`/DECSTR.

**H2 — reconnect now blanks the terminal for the whole outage.**
`apps/gmux-web/src/terminal.tsx:875-881` resets xterm at the top of `connect()`, i.e. before the socket opens, and `connect()` re-runs on every backoff attempt (`terminal.tsx:1006-1010`, up to 8 s). `termLoading` is only armed once per session effect (`terminal.tsx:864`), so nothing covers the wiped screen.

Real-browser probe (WebSocket ctor redirected to a dead port after the first drop, everything else real):
| build | buffer text during outage |
|---|---|
| 3d72aa74 | length **0** — fully blank, no overlay |
| same build, only the 4-line reset block removed | length **486** (`A-BLUE / A-SCROLL-01 …` still on screen) |

Screenshot: black terminal + "Connection lost, reconnecting…". This is precisely the dominance loss: without the reset, the last known frame survived a WS-only outage (tunnel/proxy flap, phone sleep-wake, backend socket briefly gone) and the checkpoint swapped in atomically; now every reconnect also costs a ≥1 RTT blank flash. Suggest resetting at *frame application* time (in the replay path immediately before the checkpoint chunk, or a real `\x1bc` prefixed by the server inside BSU) rather than at connect time.

## Medium

**M1 — alt-buffer checkpoints discard scrollback and can destroy the alt content.**
`renderScreen` (`cli/gmux/internal/ptyserver/ptyserver.go:394-421`) always prepends `e.Scrollback()`, which is the *normal* screen's scrollback. With `\x1b[?1049h` those lines are now pushed through xterm's alternate buffer (no scrollback) and destroyed, and the client's normal buffer is left empty. Probe: while attached to the alt session, `term.buffer.normal` text is `""`; after the session leaves alt (`printf "\033[?1049l"`), visible text length is **0** — a fully blank terminal.
Worse when `client rows < emulator height` (mobile / pre-claim attach): applying the captured real alt frame in tmux, emulator height 25 → `rows=24: ALT-CONTENT visible 0` (destroyed), `rows=25/26: visible 1`. Pre-PR the overflow landed in normal-buffer scrollback and was recoverable. Suggest skipping the scrollback section when `altScreen`.

**M2 — PR is red; the harness change breaks an unrelated spec (reproduced locally, 2 failed / 48 passed).**
1. `e2e/tests/terminal-copy-tui.spec.ts:146` — clipboard is `"A-SCR hello world…"`. `e2e/global-setup.ts:195` repurposed the *shared* session A to emit `A-BLUE`, 40 `A-SCROLL-*` lines and a permanent scroll region `\033[2;4r`; every other spec driving session A inherits that. Spawn a third session for this spec instead of mutating the shared one.
2. `e2e/tests/terminal-switch-state.spec.ts:22` — the PR's own spec times out in the full suite and passes only in isolation (6.0 s here). Its precondition (`cursorX===5 && cursorY===2`) is order-dependent on no earlier spec typing into or resizing session A.

Also green here: `go test ./internal/ptyserver` ✅, `go test -race ./internal/ptyserver` ✅, web unit tests 541 pass (1 file fails on the pre-existing `@gmux/protocol` resolve error).

## Low / nits
- `ptyserver.go:339-357`: `altCb ...func(bool)` variadic + mirrored `s.altScreen` is redundant — `s.screen.IsAltScreen()` already exists and is used by the new test itself (`ptyserver_test.go:935`); `s.screen` is read under `s.mu` at the only call site. Variadic-as-optional silently allows forgetting the callback.
- `ptyserver_test.go:928-949`: passes by early `return` inside an `if`, with the `t.Fatalf` after it — reads as inverted and is easy to mis-edit. It never covers the `?1049l` branch, nor the scrollback-before-`?1049h` interaction that actually misbehaves.
- `apps/gmux-web/src/terminal-io.ts` `reset()` clears `writeInFlight` while an xterm write callback can still be pending; that stale callback re-enters `pump()` and can issue `term.resize()` mid-write, the exact invariant the module documents. Pre-existing shape, now hit once per reconnect attempt.
- `apps/gmux-web/src/terminal.tsx:1-29`: 30 lines of import reordering in a 4-line behavior fix — hides the real hunk.
- `Validate PR title` / `Validate PR commits` fail: none of the three commits is conventional, and the third is empty.

## What I could not break
Checkpoint/live-output ordering is sound (`ptyserver.go:1599-1611`: drain → frame → register, all under `s.mu`). Probe with a live `TICK-%03d` stream and 3 forced drops: `{count:163, first:22, last:184, dupes:[], sorted:true}` — no duplicate, lost or reordered bytes. `?2026h/l` still wraps the whole frame including the new buffer byte. A→B→A isolation, buffer type, SGR bg, cursor and geometry behave as claimed in isolation. `s.altScreen` stays in sync across 1047/1048/1049 and RIS (`vt/csi_mode.go:37-56`, `vt/mode.go:6-38`). Selection loss and scroll-to-bottom on reconnect are identical before/after this PR, so not charged here.

Advisory only — not an approval or a formal change request.

mgabor3141:
**Advisory — MEDIUM — `apps/gmux-web/src/terminal.tsx:54`: the no-metadata legacy path no longer establishes buffer authority, so switching from an alternate-buffer session can paint a normal-buffer session into the old alternate buffer.**

Exact head: `a393fcbfce3922ea17e6af9dd943116d408c909f`.

Deterministic mixed-version schedule:

1. Use the new web client with an already-running older gmux-run, or pass through an older gmuxd proxy that does not forward `?client=browser`.
2. View session B while xterm's active buffer is alternate.
3. Switch to session A whose authoritative emulator buffer is normal.
4. The old backend sends the legacy `BSU → reset → rendered A → ESU` binary frame but no `terminal_checkpoint` metadata. `checkpointAlt` therefore remains `null`; `prepareBrowserCheckpoint` takes lines 54–58 and adds only SGR reset.
5. Since this delta removed the prior `term.reset()` on session change, xterm remains in B's alternate buffer. A's normal snapshot is cleared/rendered into that alternate buffer and `buffer.active.type` remains `alternate`.

Expected legacy behavior is at least the pre-delta session isolation: A becomes a clean normal-buffer terminal. Actual behavior carries B's buffer selection into A. This is observable during rolling upgrade/restart because existing runner processes can outlive the newly served frontend, and the code explicitly claims an older-runner fallback.

The new metadata ordering is otherwise sound: ptyserver writes metadata then the complete binary frame under `s.mu`, `wsClient.write`/WebSocket ordering preserves that order, and registration occurs only afterward. Browser replay stages through ESU before queueing the checkpoint, with later live frames queued behind it. Failed opens do not mutate xterm. The TerminalIO generation fence also held in the focused test: an old in-flight callback must complete before the new write and resize, while its stale `onWritten` is suppressed.

**Verdict: AMEND.** Preserve buffer isolation for metadata-less checkpoints (for example, retain a session-change-only RIS, distinct from reconnect handling), and add a deterministic old-runner/old-proxy A/B test where B is alternate and A is normal.

mgabor3141 (commented):
Delta re-review at exact head `a393fcbf`. Re-ran every finding from my first pass against a fresh disposable daemon, real `gmux attach` under a pty/tmux, and a real browser. Verdict: **AMEND** — strictly dominant over the prior patch, **not** dominant over base: one deleted line reintroduces three verified switch-time defects that `main` does not have.

## Every prior finding is fixed (re-verified, not taken on trust)

| prior | re-test | result |
|---|---|---|
| H1 `?1049h/l` in the shared attach stream | `script -q -c "gmux attach <id>"`, alt + normal sessions | `?1049h: 0  ?1049l: 0  any ?1049: 0`; frame head `\x1b[?2026h\x1b[r\x1b[H\x1b[2J\x1b[3J…`. tmux replay at rows 24/25/26 → `alternate_on=0` in all three; undersize case lands in scrollback (base behaviour), nothing destroyed |
| H2 reconnect blanks the screen | real browser, every reconnect redirected to a dead port | before / during outage identical: `len 527, 40 scroll lines, marker present`; recovery restores `baseY 17` |
| M1 alt checkpoint discards scrollback | `snapshotFrameWithScreen(..., !browserClient || !IsAltScreen())` + `?1049h` keeping the saved normal buffer | fixed for reconnect (see N2 for switch) |
| M2 harness pollution / order-dependent spec | full local suite | **50 passed, exit 0** (incl. `terminal-copy-tui`, `terminal-switch-state`) |
| L3 `reset()` cleared `writeInFlight` | `vitest run src/terminal-io.test.ts` | 33 passed; new fence test is the right pin. Residual: sticky `writeInFlight` has no watchdog — only reachable via `term.dispose()`, where `termIoRef` is dropped anyway |
| L1/L2 variadic `altCb`, inverted Go test | gone; new test asserts *absence* of `?1049` + BSU/ESU atomicity | good |
| duplicate/lost bytes | 3 forced drops during a live `TICK-%03d` stream | `{count:184, first:1, last:184, dupes:[], sorted:true}` |
| race | `go test -race ./internal/ptyserver` | ok, 21.7 s |

The redesign is the right shape: buffer authority moved out of the stream shared with a real terminal and behind a versioned `?client=browser` handshake, applied only once the complete BSU/ESU checkpoint is staged.

## New failure modes — all from deleting the session-switch RIS (`terminal.tsx:884-891`)

**N1 (High, new vs base) — input modes leak across a session switch and inject bytes into the next session's PTY.** Session A (live) turns on `?1h ?1000h ?1006h ?2004h ?1004h` like any TUI; switch to plain session B; press ArrowUp and click once:
```
head a393fcbf: appCursorKeys=true mouse=vt200 bracketedPaste=true focus=true
               B screen: "E-B-ALT^[OA^[[<0;15;8M^[[<0;15;8m"
```
B's shell received `ESC O A` plus two SGR mouse reports it never enabled. A rendered checkpoint cannot repair this: `\x1b[r`, `\x1b[0m` and `?1049l` clear none of DECCKM / mouse / bracketed paste / focus / charsets. This is precisely the case the deleted comment named.

**N2 (High, new vs base) — the previous session's screen becomes the saved normal buffer of an alt-mode session.** `?1049h` intentionally leaves the normal buffer intact (right for reconnect), but on a *switch* that buffer still holds the previous session. Probe A(30 lines)→B(in alt):
```
saved normal buffer while B is in alt: "P2-A-SECRET\nP2-A-LINE-01…P2-A-LINE-14…"   (A, not B)
after B's TUI exits alt: A's screen + A's scrollback is what the user sees
```
Quitting a TUI in B shows another session's output as if it were B's. Base showed B's own stale alt render — wrong, but same-session. The new spec only covers reconnect-in-alt, never switch-into-alt, which is why it stays green.

**N3 (Medium, new vs base) — switching into an alt session paints the whole screen with the previous session's background.** Order is `?1049h → 2J/3J → \x1b[0m`, so `activateAltBuffer(this._eraseAttrData())` and the ED both run with the stale SGR (xterm.js implements BCE). A leaves `\x1b[44m` active → every sampled right-edge cell in B's alt screen has `bg=16777220` (44/44): a full-screen blue terminal. Note the metadata-less fallback path already emits `\x1b[0m` *before* the frame, so the two paths disagree — moving the SGR reset ahead of the buffer-selection/reset bytes fixes it.

**N4 (Medium) — the fix silently does not apply to remote/peer sessions.** `services/gmuxd/internal/apiclient/client.go:328` builds `spokeURL = fmt.Sprintf("%s/ws/%s", base, sessionID)` and `ProxyWS` (`client.go:359-395`) never forwards `r.URL.RawQuery`, so the hub→spoke hop drops `?client=browser`; `checkpointAlt` stays `null` and cross-host sessions get no buffer authority at all. `wsproxy` was patched, this hop was not, and nothing covers it.

**Low:** `wsproxy` now forwards the whole client query string into the runner socket and has no test files (`go test ./internal/wsproxy/...` → "no test files"); `prepareBrowserCheckpoint` copies the (multi-MB) first chunk ~3× per attach; the new spec resolves ids by scraping `<a href>`+`split('~')` with two 1.5 s sleeps and covers neither N1 nor N2/N3; the unrelated import reordering is still in the diff.

## Dominance

- **vs prior patch `3d72aa74`: strictly dominant** — H1, H2, M1, M2, L3 all resolved, nothing reintroduced.
- **vs base `main`: not dominant** — base has none of N1/N2/N3, and all three trace to the one deleted `termRef.current.reset()`, which is *not* on the reconnect path (`connect()` re-runs without re-running the effect).

Experiment (temporary local restore of that single call, rebuilt, same probe, reverted afterwards):
```
                       retention during outage  appCursorKeys  mouse  bracketedPaste  savedNormal has A  bg bleed
head a393fcbf          preserved (519 chars)     true          vt200  true            true               44/44 cells
+ restored switch RIS  preserved (519 chars)     false         none   false           false              0/44
```
The amend is purely additive: restore the RIS in the session-switch effect (keep the reconnect path reset-free), move `\x1b[0m` ahead of the buffer-selection/reset bytes, forward the query on the peer hop, and add a switch-into-alt case to the spec. With those four, the change is dominant over both base and the prior patch and I'd read it as MERGE.

Advisory only — not an approval or a formal change request.

mgabor3141:
**Advisory — MERGE — exact head `42ed4d25269bc525d70e5aa5c66a8cde85e68005`**

Final delta re-review found no blocking ordering/concurrency defect.

Verified mechanisms and schedules:

- `apps/gmux-web/src/terminal.tsx:899-911` resets xterm only on an actual session-ID transition. Same-session failed/retry reconnects retain the last committed screen; successful reconnects stage BSU…ESU completely before enqueueing the transformed checkpoint. Legacy no-metadata reconnects retain the already-active buffer, while legacy session switches are isolated by the ID-transition reset.
- Browser metadata is emitted before the binary checkpoint while ptyserver holds `s.mu`; the client is registered only after both writes. PTY output therefore linearizes either into the drained snapshot or into a later live broadcast, not both/neither. WebSocket ordering preserves metadata → binary through both local and peer copy loops.
- Local `wsproxy` and peer `apiclient.ProxyWS` forward only the exact single `client=browser` capability. Duplicate values and extra/arbitrary query keys are rejected rather than crossing the Unix-socket or authenticated peer boundary.
- Rapid failed→successful replacement schedules cannot reset after fresh bytes: stale sockets are identity-gated, each attempt owns a separate replay buffer, and only the current socket can enqueue. The replacement checkpoint and later live messages share TerminalIO FIFO ordering.
- TerminalIO keeps `writeInFlight` set across epoch reset. The old xterm callback must complete before the new write/resize fence advances, while its stale completion callback is suppressed.

Focused validation:

- `go test ./cli/gmux/internal/ptyserver ./services/gmuxd/internal/wsproxy ./services/gmuxd/internal/apiclient -count=1` — pass.
- `pnpm --dir apps/gmux-web exec vitest run src/terminal-io.test.ts src/replay.test.ts --passWithNoTests` — 53/53 pass.
- PR E2E and full lint/build/test checks are green, including real session switch, metadata-less reconnect, margin retention, and failed-open preservation schedules.

Residual scope: I did not rerun the full Playwright environment locally; I inspected its deterministic schedules and relied on the green exact-head CI execution.

mgabor3141 (commented):
Final delta re-review at exact head `42ed4d25`. Re-ran N1–N4 and every earlier H/M finding against a fresh disposable daemon, real `gmux attach` under a pty/tmux, and a real browser. Verdict: **AMEND** — one reproducible regression versus merge base remains; everything else is clean and dominant. The gap is one byte plus one metadata field.

## N1–N4 and prior findings: all fixed (re-verified)

| item | evidence | result |
|---|---|---|
| N1 mode/input isolation on switch | A (live) sets `?1h ?1000h ?1006h ?2004h ?1004h` + blue bg → switch to B(alt), read `term.modes`, then ArrowUp + click | `{appCursor:false, mouse:'none', bracketed:false, focus:false}`; bytes reaching B's PTY = `\x1b[A` only (no `\x1bOA`, no `\x1b[<…M`) |
| N2 saved normal buffer | same probe | `savedNormal has A content: false`; after B exits alt, still `false` |
| N3 stale erase-attr bg | right-edge cells of B's alt screen | `0/44` non-default bg (was 44/44) |
| N4 peer metadata propagation | `go test ./internal/wsproxy ./internal/apiclient` ok; live probe through gmuxd: `?client=browser` → `TEXT:{"active_buffer":"normal","type":"terminal_checkpoint"}` then `BIN:98`; `?client=browser&evil=1` → `BIN:98` only; no query → `BIN:98` only | allowlist + peer hop behave exactly as documented |
| H1 raw CLI alt cleanup | `script -q -c "gmux attach <alt id>"` + tmux replay | `?1049h: 0  ?1049l: 0`; `alternate_on=0` at rows 24 and 25; undersize case lands in scrollback, nothing destroyed |
| H2 same-session outage retention | all reconnects redirected to a dead port | during outage `hasLine30=true len=398`, identical after recovery |
| M2 full Playwright | full suite locally | **50 passed, exit 0** |
| L3 TerminalIO fence | `vitest run src/terminal-io.test.ts` | 33 passed |
| race | `go test -race ./internal/ptyserver` | ok, 21.9 s |

Gating the RIS on an observed session-ID transition (`terminalSessionId` ref) is the right seam: it fires on real switches, not on mount, reconnect attempts or unrelated effect re-runs. Moving `\x1b[0m` ahead of buffer selection and erase fixes the BCE bleed at the source, and the two proxy hops now agree on one strict capability marker.

## O1 (Medium-high, new vs base) — dropping `\x1b[r` wipes the screen when a scroll region is live

`terminal.tsx:33-35` removes DECSTBM reset from the browser checkpoint so a reconnect keeps the live margins. The checkpoint body is then painted *inside* that region: `\x1b[H` lands above the top margin and each `\r\n` scrolls the 3-row region instead of advancing down the screen.

Probe (margins injected exactly as this PR's own test does):
```
before reconnect: 30 visible F2-LINE rows, margins {top:1,bottom:3}
after  reconnect:  0 visible F2-LINE rows, margins {top:1,bottom:3}
                   first 8 rows = ["F2-SHELL","","","","","","",""]   identical screen: false
```
Screenshot: a single line survives, the rest of the terminal is empty.

Causation proven by restoring the byte (temporary local edit, rebuilt, same probe, reverted; `jj diff` clean):
```
head 42ed4d25 (no CSI r):  0/30 rows after reconnect, margins {1,3},  identical=false
+ CSI r restored:         30/30 rows after reconnect, margins {0,43}, identical=true
```
Base sends `\x1b[r` to every client, so base repaints correctly and loses margins — which the post-reconnect resize/SIGWINCH redraw restores for any real TUI. Head keeps margins and destroys the frame, i.e. worse than base in a supported path: any app that sets DECSTBM while attached (`less`, `vim`, `htop`, or a shell left with a region after an abnormal exit). It self-heals only if the app redraws; a non-redrawing program stays wiped.

Note the new e2e asserts `{top:1,bottom:3}` survives the reconnect but never asserts the *content* survives, so it is green over a wiped screen.

Suggested amend: emit `\x1b[r` before the paint and re-apply the region afterwards — the emulator already knows it, so extend the checkpoint metadata (e.g. `scroll_top`/`scroll_bottom`) and have `prepareBrowserCheckpoint` append the DECSTBM after the body; then assert content-retention next to the margin assertion.

## Low / nits
- The capability is silently disabled by *any* extra query parameter (`?client=browser&evil=1` → no metadata). Right security posture, brittle contract: nothing warns if the frontend ever adds a second `/ws` parameter, and the fallback is unobservable.
- `DialWS(ctx, sessionID, browser ...bool)` reintroduces the variadic-optional style this same PR removed from `newScreen`.
- The new spec is one ~15-phase 90 s test reaching into `_core.buffers.active` and depending on a `sleep 5` inside the seeded PTY command.
- The unrelated import reordering in `terminal.tsx` is still in the diff.

## Dominance versus merge base
Dominant on: reconnect buffer authority (normal/alternate), saved-normal-buffer preservation, alt-mode scrollback, outage retention, CLI attach safety, cross-session mode/bg/content isolation, TerminalIO write/resize fencing, peer capability propagation behind a strict allowlist. Not dominant on exactly one axis: repaint fidelity with a live scroll region (O1). Close that and this reads as MERGE.

Advisory only — not an approval or a formal change request.

mgabor3141 (commented):
O1-only re-review at exact head `8097815c`. Re-ran the 30-line DECSTBM reproduction and every prior finding against a fresh disposable daemon, real `gmux attach` under a pty/tmux, and a real browser. Verdict: **AMEND** — O1 is genuinely fixed, but the placement of the margin restore breaks the checkpoint's cursor authority. One-line ordering change, verified below.

## O1 is fixed

Real PTY-side `\x1b[2;4r\x1b[3;6H` after 30 numbered lines, then a socket drop:
```
before reconnect: 30 lines, margins {top:1,bottom:3} (0-based), cursor 5,2
after  reconnect: 30 lines, missing [], margins {top:1,bottom:3}
metadata:         {"active_buffer":"normal","scroll_top":2,"scroll_bottom":4,"rows":44}
6 further PTY lines → rows 5+ unchanged = true (scrolling confined to rows 2-4)
```
Reset-before-paint and exact restore-after-paint both hold, and per-buffer semantics are right: an alt reconnect restored the alt region `{2,6}` (metadata `alternate, 3..7`) while the client's normal-buffer region stayed `{1,4}`. Library-level checks agree with the mirroring: `RegisterCsiHandler` keys on the full `ansi.Cmd` so `CSI ? … r` (XTRESTORE) can't be confused with DECSTBM; `Screen.Resize`/`Screen.Reset` both do `s.scroll = s.buf.Bounds()`, matching "reset on resize/RIS"; vt has no DECSTR path.

## O2 (Medium-high, new vs base) — the trailing DECSTBM homes the cursor

`xterm.js` `setScrollRegion()` ends with `this._setCursor(0, 0)` (DEC behaviour). The client appends the restore *after* `body`, and `body` contains the frame's `CUP` + `DECTCEM` + `ESU`, so the last thing every browser checkpoint does is home the cursor — including when the region is the full screen (`\x1b[1;44r` is still emitted).

```
PTY cursor parked with CUP 9;20 (expect 19,8), default margins:
  before reconnect: cursor 19,8
  after  reconnect: cursor 0,0
  next PTY output ('G2-MARK') lands on visible row 0 and overwrites it:
     ["printf 'G2-MARK'", "G2-MARK'\033[9;20H'", "", ""]
  after A→B→A switch: cursor 0,0
region case: first post-reconnect output overwrote row 1 (row1 unchanged: false)
```
Base and heads `a393fcbf`/`42ed4d25` end the frame with `CUP`+`?25h` and preserve the cursor; the legacy/no-metadata path here still does, so only the supported path regresses. Your suite can't see it: the cursor assertions became `expect(state.cursorX).toBeGreaterThanOrEqual(0)` in the previous round.

Verified fix (throwaway experiment: emit the restore inside `snapshotFrameWithMargins` immediately before `cursorPos`, drop the client append; rebuilt, then reverted):
```
                       content  margins  cursor(region case)  cursor(CUP 9;20)  row1 kept  MARK row
head 8097815c          30/30    {1,3}✓   5,2 → 0,0            19,8 → 0,0        no         0
+ restore before CUP   30/30    {1,3}✓   5,2 → 5,2            19,8 → 19,8       yes        8
```
Bonus: doing it in the shared frame also restores margins for `gmux attach` (today the raw frame resets and never restores — same as base, so not a regression, but the same line fixes both).

## No regression in previously fixed findings

| finding | evidence |
|---|---|
| N1 modes/input isolation on switch | `{appCursor:false, mouse:'none', bracketed:false, focus:false}`; only `\x1b[A` reaches B's PTY |
| N2 saved normal buffer | `savedNormalHasA: false` when switching into B's alt session |
| N3 stale erase-attr bg | `0/44` non-default bg cells |
| N4 capability/metadata | `?client=browser` → `TEXT:{…terminal_checkpoint…scroll_top…}` then `BIN`; `?client=browser&evil=1` → `BIN` only; no query → `BIN` only; `go test ./internal/wsproxy ./internal/apiclient` ok |
| H1 raw CLI attach | `?1049h: 0  ?1049l: 0`, no DECSTBM added, frame shape identical to base; tmux `alternate_on=0` at rows 24/25 |
| H2 outage retention | during outage `hasLine30=true len=389`, after recovery `len=398` |
| M2 full Playwright | 50 passed, exit 0 |
| L3 TerminalIO fence | 33 vitest tests pass |
| race | `go test -race ./internal/ptyserver` ok (22.8 s) |

## Mutation sensitivity
- Server mutation "ignore the tracker, always report full-screen margins" → your `terminal-switch-state.spec.ts` **fails** on the C-session margin assertion. The margin path is properly pinned.
- Nothing pins the checkpoint cursor, which is why O2 shipped green. `TestTerminalCheckpointMetadataCarriesMargins` pins JSON shape only; `TestMarginsFollowVTPerBufferAndResetSemantics` pins the tracker but not the `handleWS` buffer selection (an "always report the normal buffer's region" mutation would only be caught for the normal buffer, by e2e).

Suggested amend: move the DECSTBM restore ahead of the frame's cursor sequence (ideally into `snapshotFrame` so both clients benefit) and restore a cursor equality assertion — `cursorX/cursorY` must equal the PTY's position after reconnect and after a switch.

Dominance vs merge base: dominant on reconnect buffer authority, saved-normal-buffer preservation, alt scrollback, outage retention, CLI safety, cross-session isolation, TerminalIO fencing, peer capability, and now scroll-region preservation with a correct repaint. One axis short: post-checkpoint cursor position. Close that and this reads as MERGE.

Advisory only — not an approval or a formal change request.

mgabor3141 (commented):
Final O2 verification at exact head `46b51fb3`. Re-ran the cursor-home reproduction, the O1 40-line/margin case, normal/alternate, raw CLI, prior findings, and a mutation check — all against a fresh disposable daemon, real `gmux attach` under a pty/tmux, and a real browser. Verdict: **MERGE**.

## O2 fixed at the right seam

`cursorTailOffset()` anchors on the frame's trailing `CUP + DECTCEM + ESU` (ESU at end → exact `\x1b[?25h|l` → `…H`-terminated CSI of digits with exactly one `;` within 24 bytes) and splices the DECSTBM between repaint and tail; a non-matching tail falls back to the legacy path, which keeps cursor and content and merely forgoes buffer/margin authority. Anchoring at the tail is what makes this safe — escape sequences inside rendered content can't be mistaken for the authoritative CUP.

```
reconnect (PTY parked at CUP 9;20):  19,8 → 19,8            (was 0,0)
   marker lands visible row 8, column 19                     (was row 0)
   row 0 preserved: true                                     (was overwritten)
A→B→A switch:                        cursor 26,8 = A's real PTY position   (was 0,0)
alternate reconnect:                 cursor 11,5 → 11,5; alt margins {3,8} restored;
                                     client's normal-buffer margins untouched {0,24};
                                     metadata {"active_buffer":"alternate","scroll_top":4,"scroll_bottom":9}
```

## O1 40-line / margin scrolling still fixed

```
before: 40 lines emitted inside region 2..4, margins {1,3}, cursor 19,8, row0 "H1-LINE-01"
after reconnect: all 40 present (missing []), margins {1,3}, cursor 19,8, identical visible screen: true
marker: visible row 8 at column 19; row 0 preserved
6 further lines: margins {1,3}, row 0 unchanged, rows 4+ unchanged → scrolling confined to rows 2-4
```

## Prior findings — no regression

| finding | evidence |
|---|---|
| N1 modes/input isolation on switch | `{appCursor:false, mouse:'none', bracketed:false, focus:false}`; only `\x1b[A` reaches B's PTY |
| N2 saved normal buffer | `savedNormalHasA: false` |
| N3 erase-attr bg | `0/44` non-default bg cells |
| N4 capability/metadata | `?client=browser` → `TEXT:{…terminal_checkpoint…}` then `BIN`; `?client=browser&evil=1` → `BIN` only |
| H1 raw CLI attach | frame `\x1b[?2026h\x1b[r\x1b[H\x1b[2J\x1b[3J…\x1b[5;3H\x1b[?25h\x1b[?2026l` — no `?1049`, no added DECSTBM, identical to base; tmux `alternate_on=0` at rows 24/25, undersize case lands in scrollback |
| H2 outage retention | `hasLine20=true len=273` while every reconnect fails |
| M2 full Playwright | 50 passed, exit 0 |
| L3 TerminalIO fence | 33 vitest tests pass |
| race / go | `go test -race ./internal/ptyserver` ok; wsproxy + apiclient ok |

## Mutation check: restore-after-CUP

Reverting only the ordering (DECSTBM spliced after the cursor tail = previous head's behaviour), rebuilt, then reverted:
```
e2e/tests/terminal-switch-state.spec.ts → FAILED
  - "cursorX": 19,  - "cursorY": 8,
  + "cursorX": 0,   + "cursorY": 0,
```
The new C-session assertions (cursor after reconnect, marker at the parked column, `row0` unchanged) and the restored `cursorX===5 / cursorY===2` check on A pin exactly the defect that shipped last round; `stty -echo` removes the echo noise that made those assertions brittle before.

Dominance versus merge base holds on every axis examined across this review: reconnect buffer authority, saved-normal-buffer preservation, alt-mode scrollback, outage retention, CLI attach safety, cross-session isolation, TerminalIO fencing, peer capability, scroll-region preservation with a correct repaint, and post-checkpoint cursor authority. No open regression found — from my side this is ready to merge.

Advisory only — not a formal approval.

PR #475, opened 2026-08-04
feat: session promote, demote, and reparent

mgabor3141:
## Summary
- split mutable organizational `parent_session_id` from write-once `launched_from_session_id` launch history in schema v3
- add transactional reparenting with existence, self-parent, deep-cycle, version, dismissal, and dense sibling-order guarantees
- expose promote, demote, and reparent through additive HTTP endpoints and `gmux session ...`
- amend ADR 0026 and CLI/family documentation for ownership-tree orchestration

No UI behavior or notification-path logic is added. Existing family and notification projections consume the organizational parent field.

## Tests
- `pnpm exec moon run :build :test`
- `pnpm lint`
- `pnpm exec moon run gmuxd:check-centralstore-generated gmuxd:check-centralstore-cross-build`
- targeted centralstore, statetool, sessioncoord, gmuxd HTTP, CLI, and web family projection tests

PR #486, opened 2026-08-16
feat(web): promote a family child to its own root row, and return it, from the session menu

mgabor3141:
Frontend UX for the presentation-promotion state that #475 shipped on the backend: promote a family child to its own root row, and return it to its family, from the `⋮` session menu.

## Design decision

Three candidate affordances were assessed against the existing code (full assessment in the workspace's `.memory/assessment-promote-root-frontend.md`); one was implemented:

- **`⋮` session menu action — implemented.** The menu exists for every session view (desktop and mobile, alive and dead/replay), gating lives in a pure, unit-testable function beside `lifecycleAction`, and — decisively — a *promoted* session is no longer a family member, so the header family panel doesn't exist for it: the menu is the only surface where "Return to family" can live at all.
- **Drag a child out of / back into its family — rejected.** Family children are not sidebar rows (a family renders as one root row + one slot/member row + a counts line), so there is no draggable object; every pixel of a folder list is already an index-based *reorder* drop target with a one-PATCH invariant (`family-entry-interaction.spec.ts`); touch has no HTML5 drag at all (`canDrag` gates on `hover: hover`); and overloading drops would conflate exactly the three mutations ADR 0026 §8 keeps distinct (reorder / promote / reparent). A parallel drag system with its own drop zones would be needed — cost and ambiguity outweigh a second entry point for the same two verbs.
- **Family-panel action — rejected (this iteration).** For the selected row it duplicates the `⋮` menu ~40px away; for other rows it would add the panel's first per-row control to rows that are whole-surface navigation anchors, and would have to defend the #485 tally=filter=verb equivalence. Deep/folded members stay reachable: drawer row → navigate (the selected spine is pinned past folds) → `⋮` → promote.

## Behavior

- **Eligibility** (`promotionAction` in `family.ts`, sharing the projection's own edge rule): a family child (parent local + `semantic_agent`, cycle-safe, not promoted) gets **Promote to root**; a `promoted_to_root` session gets **Return to family** only while its *current* organizational parent resolves locally and is a semantic agent — a deleted or non-agent parent hides the action (demote would visibly do nothing). Peer-projected sessions (network peers and Local/devcontainer peers alike) get nothing: the daemon refuses the mutation with `local_only`, so offering it would lie.
- **Copy carries the ownership model**: "Shows as its own top-level session — *‹parent›* still owns it" / "Groups back under *‹parent›*". Promotion never severs `parent_session_id`.
- **No optimistic overlay**, deliberately (same rationale as `reorderSessions`): the URL derives from the presentation root, and the snapshot commit path (`commitWithSlugRewrite`) rewrites the selected session's URL atomically with the session data. An optimistic flag flip without that coupled rewrite would leave the old URL unresolvable for a beat and bounce the router home. The mutation is local-only, so the authoritative SSE snapshot echoes immediately; failures surface through the existing toast seam (`Promote failed: …` / `Return to family failed: …`), network rejects stay with the reconnecting pill.
- **Routing preserved**: promotion/demotion of the selected session rewrites the address bar in place (replace), including the worst case where the child's own `project_slug` differs from the root's — pinned by new store tests.
- A11y/mobile: plain menu buttons (note is part of the accessible name), focus returns to the `⋮` trigger on activation, tap-safe, no new motion.

No backend changes: the #475 endpoints, error semantics, and wire flag covered everything.

## Evidence

- Unit: `family.test.ts` eligibility matrix (child/process child/root/promoted/deleted‑parent/non‑agent‑parent/self‑parent/peer/cycle/dead‑retained + copy pinning); `store.test.ts` endpoint wiring, failure toasts, silence on network reject, URL rewrite across promote/demote snapshots for selected and non-selected sessions, sidebar folder reprojection. Web suite: **34 files / 753 tests pass**.
- e2e (`pnpm test:e2e`, **68 passed** including the new specs):
  - `promote-demote-menu.spec.ts` (mock fixtures, incl. a new promoted member in `demo-family.ts`): item presence + exact copy for child/promoted/plain-root, one POST per activation, focus return, #485 panel untouched, phone portrait + landscape tap path with geometry bounds.
  - `promote-demote-daemon.spec.ts` (disposable real daemon): builds a real family — stub `claude` parent (adapter resolution is basename-based ⇒ `semantic_agent: true`) + shell child via `GMUX_SESSION_ID` — then promotes and returns it through the menu at desktop/portrait/landscape, asserting the SSE round trip: own sidebar row appears/disappears, family slot row disappears/reappears, router stays on the session. Cleans up after itself.
- Screenshot validation (before / menu / promoted / demoted × desktop / portrait / landscape) captured by the daemon spec under the workspace's `.memory/screenshots/` and inspected: family entry (root row + member row + `$ 1` counts) ⇄ standalone selected row with plain header, menu note wraps within every viewport.
- `pnpm lint` (biome + tsc), `moon run :build :test` (web + Go) all green. No Go files touched.

---

## Follow-up round (pre-external-review)

**`docs:` commit** — the feature in the canonical user docs:
- **[Using the UI → Session families](apps/website/src/content/docs/using-the-ui.md)**: where the action lives (`⋮` menu), promotion = presentation + root notification semantics while the parent **still owns it** (× on the parent still stops/removes the promoted child; launch provenance preserved), Return to family rejoins the *current* parent, local-host-only, CLI equivalents linked.
- **[CLI reference → `gmux session`](apps/website/src/content/docs/reference/cli.md)**: the whole #475 verb group was missing from the website reference; added mirroring `gmux help session`, cross-linked with the UI section, noting reparent has no UI equivalent.
- `docs/task-family-ui.md`: seam doc now names the web `⋮` surface and its gating alongside the CLI override.
- Verified against the binary (`gmux help session` grammar matches) and the built site (`astro build` clean; `#gmux-session` / `#session-families` anchors and both cross-links present in dist HTML).

**`fix(web):` commit** — adjacent polish, each item backed by a reproduced deficiency or explicit repo convention (full candidate/rejection log in the workspace assessment):
1. **Double-submit guard** (reproduced: activate Promote, reopen the menu before the snapshot lands, click → two POSTs). The item now goes disabled with a busy label (`Promoting…`/`Returning…`, the `lifecycleAction` resuming convention via `promotionCopy(action, pending)`) until the authoritative snapshot flips the offered action, the session changes, or the request fails — failure re-arms it beside its toast.
2. **Escape focus restoration** (reproduced: Tab into the dropdown, Escape → focus stranded on `<body>`). SessionMenu now hands focus back to the `⋮` trigger on Escape, matching FamilyDrawer's existing convention.
3. **Accessible name** for the `⋮` trigger: its computed name was the literal "⋮" glyph; added `aria-label="Session actions"` per the app-wide icon-button convention.

Rejected after measurement (details in assessment): dropdown max-height/scroll (fits with slack even at 320px-tall landscape), note line-clamping (wraps within bounds at 390px; clamping would hide the ownership sentence), safe-area insets (menu never reaches those regions), stale-parent SSE handling (already reactive-correct). Drag/drop, family-popup controls, and reparent UI remain out of scope.

**Evidence**: +1 unit test (pending copy), +4 e2e menu tests (in-flight reopen disabled + forced click inert with exactly one POST; failure re-arms; Tab→Escape refocuses trigger; aria-label pinned), screenshot of the pending item captured and inspected. Re-validated: web vitest **754 passed**, `tsc`/biome/moon lint clean, builds green (incl. website), full Playwright **72 passed** including the three real-daemon round trips.

---

## Remediation after external review

The pushed delta at `14f952228a231ce8c32bd459b78ec5aa4f726189` addresses all three review reports:

- **No stranded promotions:** repository semantics do not support parent-project fallback: project stamps/match rules are authoritative and an unmatched promoted root has neither a sidebar row nor a routable URL. The menu therefore keeps **Promote to root** visible but disabled with an explicit “Needs a project…” explanation. Known local stamped projects and matching projects remain eligible; unknown stamped projects are gated too. Real-daemon coverage exercises unmatched cwd recovery, cross-project movement, and slug collisions.
- **Truthful docs/copy:** website, CLI help, and the task-family seam now distinguish presentation from ownership, document the active-subagent budget re-root, and state that completion suppression follows the immutable launch parent. The UI keeps the parent-ownership and current-parent demotion wording.
- **Race-safe pending state:** pending mutations are centralized and keyed by session ID plus request sequence and target flag. A late failure from session A cannot clear session B (or a newer request for A); authoritative success is announced only after the target flag arrives.
- **Keyboard/a11y:** Escape handling is registered synchronously with `useLayoutEffect` and the regression test has no timing sleep. A persistent polite atomic status region announces “Promoting to root…” / “Returning to family…” and the completed outcome; pending button labels and notes are pinned exactly.

## Delta verification

- `pnpm --filter @gmux/web exec vitest run`: **762 passed / 34 files**
- `pnpm test:e2e`: **79 passed**, including real disposable-daemon desktop/portrait/landscape, unmatched cwd, cross-project, delayed response/session-switch races, #485 filters, and immediate Escape/focus
- `pnpm lint`, `pnpm build`, `pnpm --filter @gmux/website build`: passed
- All Go module tests (including `cli/gmux` help and `services/gmuxd`): passed

Commits are separated into frontend remediation (`cdc9c2d3`) and its focused coverage (`14f95222`), with the docs corrections folded into the prior docs commit. The branch remains open, unmerged, and not installed.

---

## Second delta remediation (head `7f12e6ba32908fc03ee0f5b4d08389c371d7f952`)

Addressed N1–N5 from the latest delta review:

- Demote now proves the **post-demotion family root** has a real stamp-backed local sidebar placement. A real-daemon outside-parent/placed-child schedule leaves Return to family focusable but `aria-disabled`, prevents the POST, and preserves the promoted route.
- Promotion and demotion use the same `sidebarProjectForSession` predicate as `buildProjectFolders`; matching rules without stamps, including retained-dead sessions, are blocked. Local stamped routing uses the same predicate.
- Pending ownership/reconciliation is fully store-side: full snapshots, upserts, deletion, target-then-reversal, terminal reparenting, request tokens, and a documented 30-second final hung-request safety valve are covered.
- A stable app-level live region announces both authoritative Promote and Return to family success once; failures remain toast-only. Blocked actions are focusable native buttons with `aria-disabled`, `aria-describedby`, prevented activation, and discoverable reasons.

Delta verification: **766** web unit tests, focused promotion/daemon Playwright **23 passed**, full Playwright **81 passed**, lint/typecheck/build, website build, and CLI/gmuxd Go tests passed.

Remediation commits: `66e814b33ecae2a52998f90920598473c535843f`, `6b565b4e6f7776566ea0dcfb0590fc3b98b6cbe9`, `7f12e6ba32908fc03ee0f5b4d08389c371d7f952`.


mgabor3141 (commented):
Adversarial review of d9201a5 (frontend-only; #475 endpoints already on main). Advisory comment, not a formal review. Full notes: `.memory/review-486-opus.md`.

**Verdict: AMEND** — one reproduced defect that strands a live session out of the UI, plus newly published docs that misstate notification semantics. The rest of the mechanism held up under probing.

---

### F1 (High, reproduced against a real daemon) — "Promote to root" strands a child whose cwd matches no project

`apps/gmux-web/src/family.ts:182-195` gates eligibility on peer/edge/parent facts only; it never asks whether the *promoted* projection can be routed. After promotion the presentation root is the session itself, so `viewToPath` (`apps/gmux-web/src/routing.ts:253-278`) returns `null` when it has neither `project_slug` nor a `matchSession` hit → `selectedSlugRewrite` (`store.ts:1314-1324`) does nothing → the old URL stops resolving. The daemon can't help: a session whose cwd matches no catalog entry is left **unplaced** (`centralstore/register.go:689-691`, `placeunplaced.go:71-74`, so `project_slug` is empty on the wire), and promotion only rewrites the *scope*, never the project entry (`domain.go:1161-1179,1195-1210`).

Repro (disposable e2e daemon + Chromium, HEAD build, throwaway spec since deleted):

1. `gmux -- claude` (stub, `semantic_agent: true`) in the workspace → project `test-project`.
2. `GMUX_SESSION_ID=<parent> gmux -- sh -c 'echo child ready; sleep 600'` with **cwd outside every project entry**. API: `parent_session_id=<parent>`, no `project_slug`.
3. Browser: navigable at `/test-project/shell/~<child>`, terminal live, menu offers `Promote to root — … claude still owns it`.
4. Click →   url `/`, `.xterm` gone, `__gmuxNavigateToSession(child)` → **false**, no `.session-item` and the title absent from the page, while `/v1/sessions` still shows it alive with `promoted_to_root: true`.
5. Only `POST /demote` (i.e. the CLI) restores reachability.

Expected: the row the menu promised. Actual: no row, no route, **no ⋮ menu to demote from** — one click, no warning, no in-UI undo. This falsifies the eligibility comment at `family.ts:158-160` ("the menu can never offer a mutation whose result the sidebar wouldn't show") and the doc line "placed by its own directory, like any root". Trigger is ordinary: an agent-launched build in `/tmp`, a worktree not added as a project, a session started in `$HOME`.

Neither new test can catch it: `promote-demote-daemon.spec.ts:186` asserts `url … toContain('/test-project/')` with parent and child in the *same and only* project (the URL cannot move), and `store.test.ts:735-800` pins only the different-but-existing-project case. Suggested fix: make eligibility routable-aware (`null`/disabled when `!session.project_slug && !matchSession(session, projects)`).

### F2 (Medium) — the new user-facing docs misstate notification semantics

- `reference/cli.md:782-785`: promotion gives "full root visibility and **notification behavior**".
- `using-the-ui.md:34`: the promoted session "**notifies you like any top-level session**".
- `reference/cli.md:791-793`: "Reparenting changes … **child-notification suppression**".

The notify router keys on **`LaunchedFromSessionID`**, not `parent_session_id`, and never reads `promoted_to_root`: `services/gmuxd/cmd/gmuxd/notify_central.go:124-139` and `:162-176`. `notify_central_parent_test.go:112-119` ("promoted presentation still has launcher contract") asserts promoted child + active parent ⇒ **suppressed** (`wantPending: false`), versus `root` ⇒ `true`. And since `launched_from_session_id` is write-once, reparenting cannot move suppression either. The wording is inherited from `cli/gmux/cmd/gmux/help.go:244-246` / ADR 0026 §8 (which also claims that field has "no code consumer" — `notify_central.go:135` is one), but this PR is what ships it to end users, and its own in-app copy ("… still owns it") points the other way.

The rest of those paragraphs checks out: recursive dismissal really does reach a promoted child (`centralstore/dismiss.go:34-70,114-128`, `dismiss_test.go:281-320`), and peer refusal matches `serve_central.go:1063-1067`.

### F3 (Low) — "breaks only the presentation edge" understates it: budget ownership moves

`sessioncoord/active_subagents.go:126-129` (`resolveRoot` returns the node itself when promoted), `:265-275`, `:336-350`: promoting a live semantic child frees a slot in the parent family's active-subagent budget and gives the promoted subtree a fresh limit — `serve_central.go:1097-1098` calls it "updating budget ownership". Pre-existing from #475, but the comment at `family.ts:150-153` and the docs frame promotion as presentation-only while the UI now makes the ceiling change one click away.

### F4 (Low, reproduced) — the double-submit guard is per-mount

`main.tsx:427-430` resets `promotionPending` on `[session.id, promotion?.kind]`. Probe (`?mock`, POST held open): promote → 1 POST; navigate to another session and back; reopen the menu → item enabled, label back to "Promote to root"; clicking fires a **second POST while the first is in flight**. Harmless (`SetPromotion` is idempotent, `domain.go:1685,1728`), so the guard is UX-only — but the comment's "pending sticks until the snapshot flips … or the request fails" isn't true across navigation. Rapid triple-click without navigating correctly yields exactly one POST.

### F5 (Low, tests) — one vacuity gap

Mutating the **pending** demote note (`family.ts:207`, `Returning…` branch) to a constant leaves the whole unit suite green; every other copy string is pinned. Mutation results for the rest, all detected: drop `session.peer` guard, drop the demote parent `semantic_agent`/`peer` guard, pending promote label, demote note, promote "still owns it" (unit); drop `disabled=`, never clear pending on failure, drop the Escape `triggerRef.focus()` (e2e). Good assertions.

### F6 (Nit, a11y)

`aria-label` on the trigger is a real improvement, but there's still no `aria-haspopup`/`aria-controls`, no `role="menu"`/`menuitem`, and no focus trap — probing Tab order gives `session-menu-action ×2 → session-menu-promotion → xterm-helper-textarea`, i.e. focus leaks into the terminal past the last item. Consistent with the pre-existing menu; not a blocker.

---

**What held up under probing** (vitest + Playwright throwaways, all deleted): peer gating matches the daemon's `local_only` for network *and* Local/devcontainer peers; demote correctly hides when the parent was deletion-repaired away; no optimistic overlay + `commitWithSlugRewrite` keeps the selected session across a cross-project promote/demote *and* across promotions that newly create a slug collision (URL switches to the `~id` form, for the selected and the non-selected row alike); failures toast and re-arm, network rejects stay silent; the #485 tally/filter surface is unchanged and the panel-exclusion assertion is non-vacuous (2 real rows); `family-entry-interaction` and `popover-touch-dismiss` still pass with the new mock member; the menu path works by tap in both phone layouts. Menu-only is usable, so I'm not flagging the drag/drop or family-popup alternatives.

Commands: `vitest run` (574 pass, after building `@gmux/protocol`), `playwright test` on both new specs (14 pass) plus the two family/popover specs, `biome lint`, `tsc --noEmit`, `./scripts/build.sh` between mutations. Not exercised: a real network-peer pair, Go suites (no Go changes), actual screen-reader output.

mgabor3141 (commented):
Delta re-review of 14f95222 (was d9201a5). Advisory comment, not a formal review. Full notes appended to `.memory/review-486-opus.md`.

**Verdict: AMEND** — every prior finding is genuinely fixed and well tested, but the *same* stranding defect is still open in the demote direction, reproduced against a real daemon, and the new code, its unit test and the new docs all assert that it cannot happen.

### Prior findings — re-run

| # | Prior finding | Status |
|---|---|---|
| F1 | promote strands an unmatched-cwd child | **Fixed.** `family.ts:181-197` gates on `hasRootPlacement` and shows the item *blocked with a reason* rather than hidden. My original real-daemon repro is now the passing `promote-demote-daemon.spec.ts:245-291`: item disabled, forced click emits no POST, flag stays false, URL + terminal intact. Mutating the gate to `return true` fails it. |
| F2 | docs misstate notification semantics | **Fixed.** `cli.md:782-802`, `using-the-ui.md:34`, `help.go:244-251`, `docs/task-family-ui.md:13-26` now say suppression follows the immutable launch parent and does *not* move on reparent — matches `notify_central.go:124-139,162-176`. The `/reference/host-toml/` link is real (`host-toml.md:62`, "per local behavioral root"). `go test ./cli/gmux/...` passes with the new help text. |
| F3 | budget ownership moves, undocumented | **Fixed** in all four surfaces and in the `family.ts:157-165` comment; matches `active_subagents.go:126-129,265-275,336-350`. |
| F4 | per-mount double-submit guard | **Fixed** by the generational module-scope map (`store.ts:1854-1885`); my old repro is now `promote-demote-menu.spec.ts:240-263`. |
| F5 | pending-demote note unasserted | **Fixed** (`family.test.ts:187-199`), and copy assertions tightened `toContain` → `toBe`. |
| F6 | menu a11y | **Improved**: `aria-controls`, sr-only `role="status"`, `useLayoutEffect` for the Escape listener (mutating it back fails the immediate-Escape test). |

### N1 (Medium, reproduced on a real daemon) — demote strands the mirror image of F1

`family.ts:192-197` never applies the gate to demote, on the stated grounds that demote "restores the presentation that made the session reachable in the first place" (`family.ts:178-180`), with `family.test.ts:112-117` pinning that as correct. That is false whenever the *parent* is the unplaced one:

1. `gmux -- claude` with cwd **outside every project** → semantic-agent parent, no stamp.
2. `GMUX_SESSION_ID=<parent> gmux -- sh -c 'sleep 600'` inside `test-project` → stamped child. While unpromoted it is already unroutable (`__gmuxNavigateToSession` → `false`), since its presentation root is unplaced — pre-existing.
3. `POST /promote` → routable at `/test-project/shell/~<id>`, terminal live, healthy state.
4. The ⋮ menu offers **"Return to family — Groups back under claude", enabled**. One click → url `/`, terminal gone, `__gmuxNavigateToSession` → `false`. Recovery only via CLI/API promote.

Unit confirmation: after the demote, `viewToPath` is `null` for both child *and* parent, while `foldersFrom`'s `temporaryPlacements` (`store.ts:636-649`) still renders a `claude` row — a family row that cannot be opened. Fix: ask `hasRootPlacement` about the presentation root the demote would rejoin (`familyRoot` with the flag cleared), exactly as promote already does.

### N2 (Low, reproduced on a real daemon) — the gate's disclaimed branch is weaker than the sidebar's rule

`hasRootPlacement` (`family.ts:170-176`) accepts an **unstamped** session when a viewer match rule matches, but `buildProjectFolders` buckets by stamp only (`projects.ts:365-376`), so such a session gets no folder row — the very outcome the block exists to prevent. Repro: a child in `workspace-c` that exits (retained, `resumable: true`), then `project-c` added afterwards; the daemon re-places only *live* ids (`sessioncoord/catalog.go:74-80`), so it stays unstamped while `matchSession` matches. Promote is offered **enabled**; afterwards the URL correctly rewrites to `/project-c/shell/~<id>` and the view survives, but `project-c` is empty and the session has no row anywhere. Milder than F1 (no stranding; demote still undoes it), but "gets its own sidebar row" isn't kept. Requiring the stamp would make the gate match the sidebar.

### N3 (Low, reproduced) — a stale pending entry can wedge the item disabled until reload

The entry is settled only when the *offered kind changes* (`main.tsx:445-456`) or the request fails; if the world returns to the original kind while no menu is mounted, it is never spent. Repro: hold the promote POST → navigate away → release (promote lands) → `POST /demote` from the CLI while away → return: the item reads **"Promoting…", disabled, permanently** (unchanged after a 1.5 s wait); only a reload clears it. Same mechanism yields a stale "Promoted to root." announcement on arrival. Settling on `session.promoted_to_root === entry.targetPromoted` regardless of kind (or a TTL like `_pendingMutations`' `PENDING_TTL_MS`) would close it.

### N4 (Low, tests) — the announce guard is vacuous

Mutating `if (sameSession && reachedTarget)` → `if (sameSession)` (`main.tsx:452`) leaves the suite green: nothing asserts that a failed/superseded request stays silent. The `!project.peer` filter and the `onClick` early return are likewise unasserted (the e2e force-clicks an already-`disabled` button, which emits no event).

### N5 (Nit) — the blocked item is `disabled`, so its reason is sighted-only

`disabled` removes it from the tab order, so keyboard/screen-reader users can't reach the "Needs a project…" explanation the comment calls discoverable. `aria-disabled` plus the inert click handler already at `main.tsx:525` would keep it announceable.

### Dominance

- **vs merge base `c7dbcc36`**: dominant. No regression found in the surfaces it touches — `family-entry-interaction` + `popover-touch-dismiss` (12) still pass with the two new mock members, 762 unit tests pass, `tsc --noEmit` and `biome lint` clean, `go test ./cli/gmux/...` clean after the help edit; `family.ts`'s new runtime import is the pure `projects.ts` (no cycle, no store).
- **vs prior head `d9201a58`**: dominant on every prior finding, with one non-strict trade — F4's benign duplicate POST is gone, but the now-persistent guard introduces N3 (rare wedged item needing a reload). Net improvement: the guard finally does its job, and N3 needs a concurrent external mutation during a navigation window.
- **Residual**: N1 is the same defect class the delta set out to close, still open in the other direction and now contradicted by a passing test and a docs sentence.

Delta mutation testing: gate→always-true (✅ 2 unit + daemon test), `settlePromotion` seq check dropped (✅ 2 unit), `promotionInFlight` ignoring the map (✅ e2e), `useLayoutEffect`→`useEffect` (✅ e2e), pending demote note (✅, prior gap closed), announce `reachedTarget` removed (❌ N4).

Commands: `vitest run` (762), `playwright test` on both promote specs (21 pass) plus the two neighbouring family/popover specs, `tsc --noEmit`, `biome lint`, `go build`/`go test` in `cli/gmux`, `./scripts/build.sh` between mutations, four throwaway probes (placement-gate matrix, real-daemon demote strand, real-daemon dead-retained unstamped promote, real-daemon stale-pending wedge) — all deleted, `jj diff` clean. Not exercised: real peer pair, gmuxd Go suites (no daemon changes), actual screen-reader output.

mgabor3141 (commented):
Final delta re-review of 7f12e6ba. Advisory comment, not a formal review. Full notes appended to `.memory/review-486-opus.md`.

**Verdict: MERGE.** Every finding from both prior rounds is closed, each closure is protected by a test that fails when I break the mechanism, and the full Playwright suite (81) plus the web unit suite (766) pass. Three low-severity residuals below are follow-up material, not blockers.

### Re-run of every prior finding

| # | Finding | Status |
|---|---|---|
| F1 | promote strands an unmatched-cwd child | **Closed.** Real-daemon repro re-run: item visible, `aria-disabled="true"`, keyboard-focusable, `Enter` inert, no POST, flag false, URL + terminal intact. |
| F2/F3 | notification + budget docs | **Closed**, re-verified against `notify_central.go:124-139,162-176`; unchanged by this delta. |
| F4/F5 | double-submit guard, pending copy | **Closed.** |
| F6/N5 | blocked reason sighted-only | **Closed.** `aria-disabled` replaces `disabled` (`main.tsx:486-495`) so the reason is reachable by keyboard/AT — and mutating away the click-handler early return now fails the e2e, which matters because `aria-disabled` alone does not stop activation. |
| N1 | demote strands via an unplaced parent | **Closed.** `family.ts:219-229` recomputes `familyRoot` over a snapshot with the flag cleared, and blocks with its own copy. My real-daemon repro now blocks; deep cases check out (grandparent unplaced ⇒ blocked; parent itself a placed promoted root ⇒ allowed). |
| N2 | gate looser than the sidebar predicate | **Closed.** `sidebarProjectForSession` (`projects.ts:62-74`) is now the exact stamp-backed bucketing rule, shared by `family.ts` and `routing.ts`; my dead-retained-unstamped repro now blocks. |
| N3 | stale pending entry wedged the item | **Closed.** `reconcilePromotionPending` (`store.ts:1929-1982`) runs at the store boundary for every authoritative update. Re-ran the exact schedule (promote → navigate away → target lands → external CLI demote → return): item enabled and correct. |
| N4 | announce guard vacuous | **Closed.** Announcements now derive centrally from the observed target; mutating the announcement, the terminal-transition branch, the deletion branch or the TTL each fails a test. |

### Residual (low) findings

**R1 (Low, A/B verified) — the routing fall-through can serialize one session's URL as another's.** `routing.ts:268-278` resolves the stamp through `sidebarProjectForSession` and, when the stamp is *not* in the catalog, falls through to the disclaimed `matchSession` branch — whose URL cannot resolve back to the same session (`adoptedHere` requires *no* stamp), so it can resolve to a different one. Same probe, two heads:

- `14f95222`: `viewToPath(stale-stamped s)` → `/gone/pi/worker` → resolves to `home`.
- `7f12e6ba`: → `/alpha/pi/worker` → resolves to **session `t`**, a twin sharing adapter+slug in the matched project (`hasSessionSlugCollision` doesn't force `~id` because the namespaces differ, `local:gone:` vs `local:alpha:`).

Window: `project_slug` not present in `projects` — transient sessions/world skew after a project rename/removal, or a `PUT /v1/projects` from another client. Both heads misbehave there; the new one silently opens someone else's terminal instead of bouncing home. Returning `null` for an unknown stamp instead of falling through would close it.

**R2 (Low) — the 30 s valve re-arms a request that is still in flight.** `PROMOTION_PENDING_TTL_MS` (`store.ts:1889`) drops the entry unconditionally. Probed: after expiry the item re-arms (a second POST is possible while the first is open — harmless, `SetPromotion` is idempotent) and a later success has nothing pending to reconcile, so it is never announced to assistive tech. Fine as a stated safety valve; worth a sentence in the comment.

**R3 (Low, reproduced) — an announcement can outlive the state it describes.** A result produced while the session was unmounted is spoken on arrival even if it has since been reversed: in the N3 re-run, returning to the session yields `[data-promotion-status]` = "Promoted to root." while the session is no longer promoted. Consider dropping an undelivered announcement that a later snapshot contradicts.

**R4 (Nits).** `aria-describedby="session-promotion-note"` points at a span *inside* the button (`main.tsx:488,512`), so most screen readers read the note twice (name + description). `main.tsx:62` ships `window.__store` in `?mock` mode (used by the new e2e). The demote branch materializes a modified snapshot and rebuilds the family index every call, defeating the `WeakMap` cache — measured 1.2 ms @2 000 sessions, 7.4 ms @10 000, on every `SessionMenu` render for a promoted selection and per pending entry per snapshot; `index.rootById.get(parent.id)` answers the same question in O(1) for the non-cyclic case.

### Docs truth

Re-verified: "using them on an unplaced root can make the session temporarily unreachable" (`cli.md:794-798`) matches my observation that `__gmuxNavigateToSession` returns false after an API promote of an unplaced child; the demote precondition in `using-the-ui.md:34` and `docs/task-family-ui.md:19-25` matches `family.ts:219-229`; the notification/budget/dismissal statements are unchanged and still true.

### Dominance

- **vs prior head `14f95222`**: dominant on N1–N5; R1 is the only new risk and it replaces a home-bounce with a wrong-session open inside a window that was already broken. No earlier guarantee was withdrawn — the guard still survives navigation, the seq check still prevents cross-session settles, and every earlier assertion still passes.
- **vs merge base `c7dbcc36`**: dominant. Full Playwright suite 81/81, unit 766/766, `tsc --noEmit`, `biome lint`, `go build`/`go test` in `cli/gmux` all clean.

Delta mutation testing, all detected: demote gate always allowed (unit); `sidebarProjectForSession` loosened to `matchSession` (2 unit); reconcile never announces (2 unit); reconcile skips the terminal branch (unit); TTL never fires (unit); deletion branch doesn't clear pending (unit); click-handler early return removed (e2e).

Commands: `vitest run` (766), `playwright test` full suite (81) plus targeted runs per mutation, `tsc --noEmit`, `biome lint`, `./scripts/build.sh` between mutations, throwaway probes (gate matrix incl. dead non-resumable rows, routing A/B against the previous head, TTL + announcement lifecycle, real-daemon N1/N3 re-runs, perf micro-benchmark) — all deleted, `jj diff` clean. Not exercised: real peer pair, gmuxd Go suites (no daemon changes), actual screen-reader output.

PR #501, opened 2026-08-25
fix(cli): refuse pending stdin before launching a session

mgabor3141:
## Problem

Managed commands deliberately do not inherit launcher stdin. A prefilled pipe or heredoc therefore launched the child on an unattended PTY, printed a warning only after the session existed, and left stdin-consuming commands such as `cat`, `python -`, and `sh -s` waiting forever. Non-streaming agent harnesses did not surface the warning until their external timeout.

## Change

- inspect launcher stdin non-consumingly before any session side effects
- refuse definitely pending pipe data or unread regular-file data with exit code 1
- create no ID, socket, child, daemon request, registration, or detached re-exec on refusal
- preserve empty harness pipes and `/dev/null`
- remove the post-launch warning and amend ADR/help/reference/skill documentation

This is the immediate safety fix. Transparent streamed stdin and independent input/output attachment remain follow-up work.

## Verification

- `cd cli/gmux && go test ./...`
- focused stdin regressions under `-race`
- `pnpm --dir apps/website build`
- real-binary regressions assert early latency, stable exit code, no child marker, no daemon request, no socket artifacts, and byte-for-byte unconsumed input for foreground, detached, pipe, and regular-file cases


mgabor3141:
@greptile please review.

PR #512, opened 2026-08-31
perf(web): index sidebar projections so large session lists stay responsive

mgabor3141:
Slice S0 of the frontend sync architecture assessment: a pure client-side algorithmic fix, no wire/protocol change, no new dependencies.

## What

`sidebarSessions`, `foldersFrom`, and `unreadCount` in `store.ts` ran `Array.find` / `Array.some` / `projects.some` inside per-session loops, making every derived recompute quadratic — and protocol 3 delivers every one-session mutation as a full snapshot, so a single unread flip cost the same as a cold load (~1.5 s main-thread stall at 10k retained sessions, ~16 s at 30k).

This replaces those scans with the already-existing WeakMap-cached `familyIndex` (`byId` / `rootById` / `childIds` Maps+Sets, one build per snapshot shared by all computeds) plus a new memoized `projectKeys` set, and a `Set`-based dedup in `sidebarSessions`. `selected` also uses the index. Output is bit-identical; ordering, folder semantics, family ownership/temporary placements, filtering, peer ownership, and unread/error attention are all preserved.

## Numbers (i7-9700K, Node 24, fixture `makeFixtureWorld(seed=1234, n)`, `--expose-gc`)

One-session mutation → all projections (`sidebarSessions`, `folders`, `unreadCount`, `sidebarActivity`, `familyDotById`, `familyActivityById`, `homePartition`):

| n | before | after | cold before | cold after | heap Δ |
|---|---|---|---|---|---|
| 1,000 | 22.2 ms | 10.6 ms | 30.6 ms | 19.8 ms | 0.4 MB (unchanged) |
| 10,000 | 1,513.7 ms | 103.9 ms | 1,104.7 ms | 113.2 ms | 1.5 MB (unchanged) |
| 30,000 | 16,026.0 ms | 378.8 ms | 16,502.7 ms | 360.6 ms | −2.4 MB (unchanged) |

Scaling is now ~linear (10k→30k ≈ 3.6×).

## Tests

- **`store-projections.test.ts`** — differential/property suite: the pre-optimization algorithms are kept verbatim as reference implementations and every randomized seeded world (multi-host, family edges incl. malformed, unstamped strays, transitional families, unresolvable stamps; crossed with tab filters, alive-only, root/child selection, 20 random single-session mutations, removals) must project identically through both. Mutation-tested: deleting one line of the new code fails 8 assertions.
- Perf regression test: best-of-3 one-mutation derived cost at 10k sessions < 800 ms (old code ≈ 1.5 s, new ≈ 100 ms — wide margins on both sides for CI stability) + a family-index sharing/memoization identity test.
- **`store.bench.ts`** — `vitest bench` at 1k/10k/30k (cold + one-session mutation) over the same fixture.
- Full web suite (786 tests), `tsc --noEmit`, biome lint, vite build, and Playwright `harness` / `sidebar-view-toggle` / `family-entry-interaction` / `promote-demote-menu` (35 passed) all green.

Ref: `.memory/report-frontend-sync-architecture.md` §7 slice S0 / addendum §B (assessment measured ~1,800× at 30k on the two projections alone; end-to-end here is ~42× because the remaining linear work — view resolution, day partitioning, folder sort — now dominates).

### DIFF
--- a/apps/gmux-web/src/store-perf-fixture.ts
+++ b/apps/gmux-web/src/store-perf-fixture.ts
@@ -1,0 +1,155 @@
+/** Deterministic synthetic worlds for the store-projection differential and
+ * performance tests (`store-projections.test.ts`, `store.bench.ts`).
+ *
+ * Mirrors the corpus shape from the frontend-sync assessment probes:
+ * multiple hosts (one Local devcontainer peer), ~50 projects plus peer
+ * references, ~10% task-family children (with grandchildren), retained-dead
+ * sessions, unread flags, unstamped strays, unresolvable stamps, and
+ * transitional families (stamped child under an unstamped root — the
+ * temporary-placement path in `foldersFrom`).
+ *
+ * Everything is seeded (Mulberry32) so a fixture is exactly reproducible:
+ * the same (seed, n) pair always yields byte-identical worlds, which lets
+ * perf numbers in reports/benches reference "fixture seed=S n=N" precisely.
+ */
+import type { PeerInfo, ProjectItem, Session } from './types'
+
+export interface FixtureWorld {
+  sessions: Session[]
+  projects: ProjectItem[]
+  peers: PeerInfo[]
+  peerProjects: Record<string, { slug: string; launch_cwd?: string }[]>
+  /** A mid-list family child id, for selection-dependent projections. */
+  selectedChildId: string
+  /** A root session id whose unread bit the mutation benches flip. */
+  mutationTargetId: string
+}
+
+/** Mulberry32 — tiny, high-quality-enough seeded PRNG. */
+export function mulberry32(seed: number): () => number {
+  let a = seed >>> 0
+  return () => {
+    a = (a + 0x6D2B79F5) | 0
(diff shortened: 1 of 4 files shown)


mgabor3141:
Addressed the two review advisories (pushed as an amended commit, b33d1e6):

- **Load-bearing `selected` test**: `expectProjectionsMatchReference` now asserts `selected.value` is the exact snapshot object (or null with no selection) on every property-world step, plus a deterministic root/child URL-selection test. Mutant-verified: both a `selected → null` mutant and a `selected → {…copy}` mutant fail 9/12 tests.
- **Perf threshold 800 → 400 ms**, calibrated over 24 repeated best-of-3 local runs: optimized 29–37 ms, pre-optimization O(N²) floor 1,200–1,487 ms — so 400 ms keeps >10× CI headroom while sitting 3× below the old implementation's best case.
- **Object-identity check corrected**: the sidebar row comparison previously said "identity" but used deep `toEqual`; it now asserts `toBe` per element.

Full web suite (787), tsc, biome all green; perf test passed 3/3 consecutive runs at the new threshold.

PR #513, opened 2026-08-31
fix(peering): surface spoke session-stream omissions in the hub world projection

mgabor3141:
## Problem

Protocol 3 is fail-open by design: rows a sender cannot ship (`row_too_large` > 48 KiB, `transaction_limit` past the 100k-row / 64 MiB bounds, encode failures) are quarantined with bounded diagnostics and `ready` still commits the survivors. Browsers surface this via the persistent omitted-sessions banner — but on the **peer hop** the diagnostics only reached `log.Printf` and were discarded at `ready`. A hub therefore presented a merged projection that silently missed remote rows. Worse, the peer-side 256-entry diagnostic detail cap could swallow the sender's `diagnostics_suppressed` counted summary (it arrives as the 257th error event), so even the total was lost.

Verified against the frontend-sync assessment's reproductions: aggregate overflow (63,713 committed / 6,287 omitted with exact 256+6031 accounting), the >48 KiB single-row exile (reachable today via unbounded `command`/`title`/`remotes`), and the peer-hop loss.

## Fix (narrow, protocol-3 compatible)

- Peer staging now accumulates an **exact omitted total** and a bounded per-code breakdown separately from the capped detail list, so counted summaries past the cap are never lost. Counts clamp to `MaxStagedRows`; code cardinality (8 + `other`) and length (64) clamp against hostile senders.
- On `ready`, after the surviving rows commit atomically, the summary is published on the `Peer` and — when it changes — marks the world projection dirty. It surfaces as additive optional fields `peers[].sessions_omitted` / `sessions_omitted_codes` in `snapshot.world` and `/v1/health` (`WorldProjection` + `PeerStatus`).
- A clean `ready` or a legacy single-frame snapshot clears the marker; an abandoned transaction leaks nothing; the marker persists across disconnects, matching the retained rows.
- Browser folds peer omissions into the existing sidebar omitted-sessions banner with per-host detail lines (`host: N sessions omitted upstream`), defensively re-validating counts.

## Decisions

- **Quarantine, not truncation**: oversized rows stay omitted-but-counted. Truncating summary fields at projection time would silently alter authoritative session data consumers rely on (resume commands, cwd); schema-bounded summary rows are protocol 4's job (see sync report addendum §C.1). Diagnostics continue to carry only bounded identities (SHA-256 for long IDs), never row contents.
- **Compatibility**: purely additive `omitempty` JSON fields; old peers never see them (peers don't receive `snapshot.world`), old browsers ignore them; sessionstream wire format and 48 KiB envelope untouched.

## Tests

- Go (`internal/peering`, race-clean): exact end-to-end accounting through `WorldProjection`/`PeerStatus` incl. suppressed summary; hostile-sender clamps; marker lifecycle (abandon / persist / legacy clear); exactly-once world notification.
- Web (vitest): `peerStreamOmissions`/`peerOmittedTotal` projection, rejection of absent/zero/negative/non-integer/unsafe counts, clearing on later snapshots.
- Mutation checks: dropping counted totals, suppressing the world notification, skipping the legacy clear, and weakening the browser-side validation each fail a test.

`GOTOOLCHAIN=go1.26.1 go test -race ./...` (gmuxd), full vitest suite (778), `tsc --noEmit`, biome — all green.

### DIFF
--- a/apps/gmux-web/src/sidebar.tsx
+++ b/apps/gmux-web/src/sidebar.tsx
@@ -24,7 +24,7 @@ import {
   unreadCount, localHostLabel, unresolvedHosts, duplicateConversationFiles,
   sidebarActivity, sidebarMode, setSidebarMode,
   activeSelectors, removeSelector, setHostFilter,
-  aliveOnly, setAliveOnly, tabHref, navigate, sessionStreamWarnings, sessionStreamOmittedTotal,
+  aliveOnly, setAliveOnly, tabHref, navigate, sessionStreamWarnings, sessionStreamOmittedTotal, peerStreamOmissions, peerOmittedTotal,
   type DotState,
 } from './store'
 import { HostSuffix } from './host-suffix'
@@ -913,9 +913,11 @@ export function Sidebar({
   const totalVisible = foldersVal.reduce((n, f) => n + f.sessions.length, 0)
   const connected = connState.value === 'connected'
   const streamWarnings = sessionStreamWarnings.value
-  const omittedSessionCount = sessionStreamOmittedTotal.value
+  const localOmittedCount = sessionStreamOmittedTotal.value
+  const peerOmissions = peerStreamOmissions.value
+  const omittedSessionCount = localOmittedCount + peerOmittedTotal.value
   const detailedOmittedCount = streamWarnings.reduce((sum, warning) => sum + warning.count, 0)
-  const suppressedOmittedCount = Math.max(0, omittedSessionCount - detailedOmittedCount)
+  const suppressedOmittedCount = Math.max(0, localOmittedCount - detailedOmittedCount)
   const hasProjects = projectsVal.length > 0
   const isOnlyHomeProject = projectsVal.length === 1
     && projectsVal[0].slug === 'home'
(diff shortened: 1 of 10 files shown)


mgabor3141:
Amended per adversarial review (f452a46):

**Finding 1 (HIGH)** — `managerProjectionSink.PeerWorldChanged` now mirrors `onStatus`: it fires `EventHooks.PeerWorldDirty` in addition to forwarding to a non-nil sink, so the production central path (`sink == nil`) recomposes `snapshot.world` on omission marker set/clear. No duplicate-recompose risk: the only caller (`setSessionOmissions`) is change-gated.

**Finding 2 (LOW)** — `recordOmission` now saturates the total and the per-code entry against shared headroom (`MaxStagedRows − omittedTotal`) with the same effective increment, so `sessions_omitted_codes` always sums exactly to `sessions_omitted` and every entry is bounded. The old per-event clamp became dead code (the review's surviving mutation) and was removed.

New tests: `TestPeerOmissionChangeFiresPeerWorldDirtyHook` (production-shaped `sink=nil` + hooks; SET and CLEAR with byte-identical survivors so `ReplacePeerSessions` no-op suppression can't mask it; unchanged marker stays silent), and the hostile-sender test now streams repeated `1<<60` counts plus a truncated long code and asserts per-code/total sum consistency and exact saturation values.

Mutation-kill evidence: removing the hooks firing → FAIL; reintroducing unclamped per-code accumulation → FAIL; prior mutations (dropped counted totals, suppressed notification, skipped legacy clear) still caught. `GOTOOLCHAIN=go1.26.1 go test -race ./...`, 778 vitest tests, tsc, biome — all green.

PR #517, opened 2026-08-31
perf(gmuxd): run startup conversation indexing in the background

mgabor3141:
## Problem

Startup conversation discovery (`convIndex.Snapshot()`) synchronously walked and fully parsed every adapter conversation before any listener bound. On a real 2786-conversation corpus that blocked first `/v1/health`, SSE bootstrap, CLI `gmux start` readiness, and every REST route for **33 s** (system resource profile §1, risk #5). The cost scales with the user's conversation history, not gmux state.

## Change

Bind listeners promptly; index in the background with bounded concurrency:

- **`Index.StartSnapshot(ctx, workers, onAdapterIndexed, onComplete)`** scans adapter `ConversationSource`s on background goroutines, at most 2 adapter scans in flight. It starts *after* `StartOwnedTriggers`, so the source watchers are already running — a conversation created/changed during the scan is observed rather than lost (strictly narrower than the old snapshot-then-watch gap). Ownership settlement (`bootstrapOwnership`) still precedes discovery, so autostart contenders that yield to a healthy incumbent pay nothing (#460/#461 semantics unchanged).
- **Coherent intermediate state:** sessions serve immediately from centralstore — titles stay runner-reported last-known-good (#508; this index never writes the store) — and gain resume commands progressively. Each adapter completion marks the composer dirty, so SSE subscribers see the enrichment without reconnecting.
- **Explicit signals:** `/v1/health` gains `conversation_index: {status: indexing|ready, indexed}`; `/v1/conversations/{adapter}/{slug}` answers `503 indexing` + `Retry-After` instead of a false 404 while the scan hasn't established authority over absence; found entries serve immediately.
- **No races:** index mutations remain mutex-serialized and `Upsert` is keyed by adapter-native conversation ID, so watcher events racing the scan neither duplicate nor lose entries. A snapshot interrupted by shutdown never reports complete. The index is memory-only, so a restart mid-indexing simply rescans.

## Proof (3000×200 KB synthetic pi corpus, isolated HOME/state, warm cache)

| metric | blocking (main) | this PR |
|---|---|---|
| spawn → first `/v1/health` 200 | 4.15 s | **0.06 s** |
| SSE `/v1/events` first frame after bind | 8 ms | 8 ms |
| background index complete | — | 4.1 s (`indexed 3000 in 4.07s`) |

- **Differential:** after both daemons finish indexing, all 3000 `/v1/conversations/pi/<slug>` responses are byte-identical between blocking and async binaries (0 mismatches).
- **Mid-indexing restart:** `kill -9` at ~1 s, restart → converges to `{status: ready, indexed: 3000}`.
- **Intermediate state observed live:** health during scan reported `{indexing, indexed: 17}`; unknown slug returned `503 indexing`, then `404` post-completion; known slug `200`.
- **Race coverage:** new tests for bounded concurrency, blocking/async differential state (incl. slug-collision suffixes), concurrent watcher-style upserts during the scan (no dup/lost of 44 IDs), and cancelled-context incompleteness. `go test -race -count=3 ./services/gmuxd/internal/conversations/` and `go test -race ./services/gmuxd/cmd/gmuxd/` pass; full `./services/gmuxd/...` green.

Follows the phase-1 rollout ("A: serve-first, async full index, explicit status") recommended by the adapter-reindex startup assessment; the durable adapter-owned index (phase C) remains future work.

### DIFF
--- a/services/gmuxd/cmd/gmuxd/conversations_http.go
+++ b/services/gmuxd/cmd/gmuxd/conversations_http.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"net/http"
+
+	"github.com/gmuxapp/gmux/services/gmuxd/internal/conversations"
+)
+
+// handleConversationLookup serves GET /v1/conversations/{adapter}/{slug}.
+//
+// While the startup snapshot is still running, a lookup miss is ambiguous —
+// the conversation may simply not have been scanned yet — so the handler
+// answers 503 "indexing" with Retry-After instead of a false 404. Completeness
+// is read BEFORE the lookup: only a miss against an index that was already
+// complete when the request started is an authoritative 404. Reading it after
+// the miss could race the final adapter's upsert + completeness flip and
+// 404 a conversation that is in the index by the time the response is
+// written; with this order, the worst case is one extra retryable 503.
+func handleConversationLookup(convIndex *conversations.Index) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		adapterName := r.PathValue("adapter")
+		slug := r.PathValue("slug")
+		complete := convIndex.SnapshotComplete()
+		info, ok := convIndex.Lookup(adapterName, slug)
+		if !ok {
+			if !complete {
+				w.Header().Set("Retry-After", "2")
+				writeError(w, http.StatusServiceUnavailable, "indexing", "conversation index is still loading; retry shortly")
+				return
+			}
+			writeError(w, http.StatusNotFound, "not_found", "conversation not found")
+			return
+		}
+		writeJSON(w, map[string]any{"ok": true, "data": map[string]any{"slug": info.Slug, "adapter": info.Adapter, "title": info.Title, "cwd": info.Cwd, "resume_command": info.ResumeCommand, "created": info.Created}})
+	}
+}
+
+// conversationIndexHealth is the /v1/health "conversation_index" payload:
+// transport readiness must not imply complete conversation discovery, so the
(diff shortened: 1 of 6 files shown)


mgabor3141:
Amended per review (`446b7a0a`):

- **F1 (zombie):** `Scan` snapshots a per-`(adapter, ref)` removal generation before its unlocked `DescribeConversation` and commits `resumeByRef` + index entry under one lock hold only if unchanged; `Remove`/`RemoveByRef` bump the generation — including for refs not yet indexed (the mid-scan case). Deterministic regression test (`TestScanRemoveRaceDeletionWins`) reproduces the reviewer's schedule via a gated describe for both unindexed and pre-indexed refs, plus asserts a post-removal rescan of a re-created file still indexes. Mutation checks: neutering the generation check fails both subtests with the reviewer's exact zombie symptom; dropping the unconditional bump in `RemoveByRef` also fails both.
- **F2:** `handleConversationLookup` reads `SnapshotComplete` *before* the lookup, so a miss can only 404 against an index already complete at request start; the ready-flip race now costs at most one extra retryable 503.
- **Test gap:** extracted `handleConversationLookup`/`conversationIndexHealth` and added an httptest covering 503+`Retry-After` during indexing, 200 for already-indexed entries mid-scan, 404/200 after completion, and both `conversation_index` health states.

`go test -race` on conversations (×2) and cmd/gmuxd green; full `./services/gmuxd/...` suite green. F3 (peering flake in the race-detector job) acknowledged as pre-existing — will rerun the job if it trips.

PR #526, opened 2026-09-08
fix: rerun dead shell sessions instead of refusing the relaunch the UI offered

mgabor3141:
## The bug

A dead shell session in the sidebar offers a relaunch; clicking it fails:

```
Resume failed: sessioncoord: resume spawn for b75zx03g: runner spawn: session b75zx03g is not resumable
```

Two predicates disagreed:

| Side | Predicate (before) |
|---|---|
| Presentation → `resumable` → the button | dead ∧ recorded `command` ∧ verdict ≠ Gone (`snapshot/wire/convert.go:161-171`) |
| Execution → `POST /resume` → runner spawn | `ResolveResumeCommandFor(adapter, conversationRef)` non-empty (`serve_central.go:301`, `bootstrap_runner_adapters.go:437`) |

They agree only when the row has a conversation ref. A shell never has one —
the ref comes solely from the agent hook's `conversation_file` event — so
**every dead shell was advertised as relaunchable and refused at spawn**. Same
for `editor` and for any agent that died before its hook bound a conversation.
The pre-existing agreement test only covered rows that *do* have a ref.

Reproduced on disposable daemons, locally and across a peer. Two extras found:

- **Restart destroyed live shells**: restart is stop + the same spawn, so the
  stop succeeded and the respawn refused.
- The peer path was *not* broken: the request is forwarded and the owning
  daemon produced the error. The peer projection just faithfully relayed the
  owner's own wrong `resumable`.

## The fix

**One policy, both sides.** New dependency-free leaf
`services/gmuxd/internal/relaunch`: a conversation ref that resolves →
`resume`; a ref that resolves to nothing → refuse (the adapter is
authoritative; the user asked for *that* conversation); no ref but a recorded
command → `rerun`. The wire converter and the runner spawner both call it.

**A truthful affordance on the wire.** Additive `relaunch: "resume" | "rerun"`
(omitted = the daemon will refuse) on `wire.Session` and on the peer
projection, so a peer row carries the *owning* daemon's verdict. `resumable`
keeps exactly its old value — sidebar visibility is unchanged.

**The UI reads the verdict instead of guessing it** from an adapter-name
allowlist (kept only as the fallback for peers on older daemons), and routes
the verb into the toast: a failed rerun says "Rerun failed", not "Resume
failed".

**Restart refuses before stopping**, so it can no longer be a kill with extra
steps, and the remaining refusals are actionable:
`… cannot be resumed: its pi conversation (/path.jsonl) is missing, empty, or
no longer resumable by the adapter`.

## Verification

- `go test -race ./...` green across gmuxd/cli/adapter/paths. New: relaunch
  policy table; `TestRelaunchVerbOnTheWire`;
  `TestRelaunchDerivationPresentationMatchesSpawn` (presentation ⇔ spawn for
  ref-less rows across shell/editor/claude/codex/pi/unknown — the systemic
  guard); guard-message and restart-guard tests.
- Web: vitest 870 green, `tsc --noEmit` clean, biome clean.
  `session-actions.test.ts` rewritten around the wire verdict + legacy
  fallback.
- e2e: new `relaunch-affordance.spec.ts` covers a dead **local** shell and a
  dead **peer** shell (the spec starts and tears down a second isolated
  gmuxd): button shown ⇔ says "Rerun" ⇔ click brings the session back alive on
  the *owning* daemon. Full Playwright suite 112 passed. Reverting just the
  spawner wiring makes the new spec fail, so it guards the reported bug.
- Docs: `develop/session-schema.md`, `architecture.md`, `using-the-ui.md`.

### DIFF
--- a/apps/gmux-web/src/main.tsx
+++ b/apps/gmux-web/src/main.tsx
@@ -16,7 +16,7 @@ import { MenuButton } from './menu-button'
 import { installCopySession } from './mock-data/export-session'
 import { ReplayView } from './replay-view'
 import { viewToPath } from './routing'
-import { lifecycleAction } from './session-actions'
+import { lifecycleAction, type RelaunchVerb } from './session-actions'
 import { SettingsModal } from './settings'
 import { Sidebar } from './sidebar'
 import {
@@ -182,7 +182,7 @@ class ErrorBoundary extends Component<
 function MainHeader({ session, onRestart, onResume, resuming }: {
   session: Session | null
   onRestart?: () => void
-  onResume?: (id: string) => void
+  onResume?: (id: string, verb?: RelaunchVerb) => void
   resuming?: boolean
 }) {
   const familyTriggerRef = useRef<HTMLButtonElement>(null)
@@ -328,7 +328,7 @@ function sessionHref(session: Session): string | undefined {
 function SessionMenu({ session, onRestart, onResume, resuming }: {
   session: Session
   onRestart?: () => void
-  onResume?: (id: string) => void
+  onResume?: (id: string, verb?: RelaunchVerb) => void
   resuming?: boolean
 }) {
   const [open, setOpen] = useState(false)
@@ -398,7 +398,7 @@ function SessionMenu({ session, onRestart, onResume, resuming }: {
   // button in ReplayView's action bar.
   const action = lifecycleAction(session, resuming)
   const actionHandler = action?.id === 'restart' ? onRestart
-    : action?.id === 'resume' && onResume ? () => onResume(session.id)
+    : action?.id === 'relaunch' && onResume ? () => onResume(session.id, action.verb)
     : undefined
   const showStale = action?.id === 'restart' && !!staleKind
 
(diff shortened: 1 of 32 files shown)


mgabor3141:
Amended after adversarial review (rebased onto `de1904f9`; #525/#529 merged cleanly under this change).

- **F1** — the production `ResolveCommand` lambda is now the named `productionResolveRelaunchCommand`, injected at the call site, used as the spawner's default branch and by the HTTP guards; the agreement tests import *that value*. Reverting it to `ResolveResumeCommandFor` now fails `go test ./cmd/gmuxd/` (5 subtests + `TestAgentResumeGuard`), not just Playwright.
- **F2** — new `relaunch_restart_route_test.go` drives `POST /restart` through the real route against a live registered runner and asserts 400 `not_resumable` **and** zero `Terminate`/`Spawn` and a surviving registry entry. Moving the guard after `Coordinator.Restart` fails it.
- **F3** — closed, not just documented: `Coordinator.Restart` reads the row under its claim, asks the spawner `CanSpawn` (new optional `RunnerSpawnChecker`, same predicate `Spawn` applies) and refuses *before* the stop, then spawns the row it authorized instead of the post-stop re-read. The reviewer's probe is now `TestRestartSpawnsTheRowItAuthorized` (durable read returns a freshly-bound unparseable `conversation_file` on the second call). Both directions mutation-checked.
- **F4** — chose "surface": `resumeSession` turns the daemon's `original_cwd`/`fallback_cwd` into an info toast ("Rerunning in /home/user instead (/gone no longer exists)"), via a pure helper with unit + store-seam tests; `using-the-ui.md` and `session-schema.md` no longer promise "the same directory".
- **F5** — the exit-less refusal now reads "this session has no recorded exit: if it is still running, restart it instead; if it just stopped, retry in a moment" (true for the convergence window too).
- **F6** — `orchestrating-agents.md` / `reference/cli.md` now state that prompting a session that never bound a conversation reruns the recorded command into a fresh one.
- **F7** — `relaunch` parses as a plain string; an unknown future verb degrades to the `resumable` fallback instead of throwing.
- **F8** — the check order in `relaunchVerb` is documented as safe (both fields come from one `kind`).
- e2e failure signal: the liveness poll now surfaces the daemon's toast text (verified against the M1 mutant).

Verification: `go test -race ./...` green (gmuxd, cli, adapter, paths); vitest 874; tsc + biome clean; full Playwright suite 115 passed under a unique TMPDIR.

PR #527, opened 2026-09-08
fix(web): patch-release polish — resize refusal, day labels, wake retry, dep trim

mgabor3141:
Four small, independent web fixes for v2.1.1. No rewrites; each commit stands alone.

## 1. Refuse non-viable terminal measurements instead of clamping to 1 row
`measureTerminalFit` clamped a sub-cell measurement to the minimum grid
(`Math.max(1, …)`), so a transiently tiny container — the mobile keyboard
animating in, a collapsed pane, a flex child mid-relayout — published a real
1-row resize to the PTY and reflowed the running program for a layout the user
never saw.

Restores the refusal from #463 that #466 dropped (#466 fixed the *horizontal*
case with a CSS `min-width`, which left the vertical one unguarded). `null`
already means "not measurable yet" everywhere downstream: the first-claim path
retries on later frames and falls back to checkpoint geometry on its deadline,
so the claim/checkpoint invariants from #484/#502/#509/#514 are untouched.

## 2. Refresh day-relative sidebar labels when the local day turns
"Today"/"Yesterday" were only re-evaluated at a snapshot commit. Since #518 an
identical snapshot commits as a no-op, and a quiet tab commits nothing at all,
so a sidebar left open across midnight kept calling yesterday "Today".

Adds the missing clock trigger: **one** timer armed for the next local midnight
(computed from calendar parts, so DST still lands on the boundary), re-armed
after firing, cleared with the rest of `initStore`'s cleanups. One signal write
per day — not polling.

## 3. Drop the unused `@trpc/client` dependency
Nothing in the repo imports tRPC; the frontend talks REST + SSE + WebSocket.

## 4. Let a wake pull a deep pending SSE retry forward
(From the #522 review.) A wake arriving while a backoff timer was pending was
absorbed entirely, so a tab that woke one second into an 8 s (up to 9.6 s with
jitter) wait stayed dark for the rest of it.

A pending retry more than a second away is now rescheduled to now; timers about
to fire are left alone so a burst of wakes cannot churn sources. `attempts` is
deliberately **not** reset — the backoff is pulled forward, not collapsed, so a
wake still cannot drag the retry rate back to the 500 ms floor (only a user's
tap does). No other supervisor semantics change.

## Verification
- `pnpm exec vitest run` — 46 files, 878 tests pass (new: `terminal-resize`
  grid-refusal cases, `day-boundary.test.ts`, two supervisor wake cases)
- `pnpm exec tsc --noEmit`, Biome check on touched files
- Playwright (isolated daemon, `TMPDIR` scoped): `terminal-resize`,
  `terminal-reattach-geometry`, `initial-checkpoint-geometry`,
  `terminal-claim-retry`, `reconnect-stability` — all pass
- `./scripts/build.sh`
- Mutation checks: dropping the grid guard fails 2 resize tests; a no-op
  midnight timer fails 2 day-boundary tests; absorbing the wake, collapsing the
  backoff, or rescheduling unconditionally each fail a supervisor test

### DIFF
--- a/apps/gmux-web/package.json
+++ b/apps/gmux-web/package.json
@@ -18,7 +18,6 @@
     "@gmux/protocol": "workspace:*",
     "@gmux/scroll-anchor": "workspace:*",
     "@preact/signals": "^2.9.0",
-    "@trpc/client": "^11.6.0",
     "@xterm/addon-fit": "0.12.0-beta.288",
     "@xterm/addon-search": "0.17.0-beta.288",
     "@xterm/addon-image": "github:gmuxapp/xterm.js#v6.1.0-beta.288-gmux.5&path:addons/addon-image",
--- a/apps/gmux-web/src/day-boundary.test.ts
+++ b/apps/gmux-web/src/day-boundary.test.ts
@@ -1,0 +1,121 @@
+/** Day-relative labels ("Today", "Yesterday", "Last <weekday>") are a
+ * function of the wall clock, but every other trigger in the store is a
+ * function of *data*. #518's incremental reconciliation made that gap
+ * observable: an identical snapshot commits as a no-op, and a quiet tab
+ * commits nothing at all, so a sidebar left open across midnight kept
+ * calling yesterday "Today" until some unrelated update arrived.
+ *
+ * watchDayBoundary is the clock trigger: one timer per app lifetime, armed
+ * for the next local midnight, re-armed after it fires, cleared on unmount.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import {
+  _rawSessions, _setRawWorld, homePartition, sessionsLoaded, urlHash, urlPath,
+  urlSearch, watchDayBoundary, worldLoaded,
+} from './store'
+import { makeSession } from './test-helpers'
+
+;(globalThis as { window?: unknown }).window ??= globalThis
+
+// Local (not UTC) parts: the partition computes calendar days in local time.
+const NOW = new Date(2026, 0, 12, 15, 0, 0).getTime()
+const HOUR = 60 * 60 * 1000
+
+beforeEach(() => {
+  vi.useFakeTimers({ now: NOW })
+  _setRawWorld({ projects: [], peers: [], peerProjects: {} })
+  _rawSessions.value = [makeSession({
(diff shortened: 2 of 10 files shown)


mgabor3141:
Amended for the review (rebased onto `main` @ `de1904f9`, head now `cf0ebcc10490`).

**F1 (HIGH) — fixed as suggested, and pinned.** The refusal is no longer
committed as the viewport: `processViewportResize` returns early on `null`, and
`fitAndResize`/`announceViewportClaim` got the same treatment (both also wrote
`null` before returning — `fitAndResize` is the pill's own tap handler, so
committing a refusal there hid the pill and removed the only manual recovery).
Two regression tests:
- e2e `terminal resize › a shell that collapses below one cell and settles
  smaller self-heals` — the reviewer's probe shape (collapse to 8 px, settle at
  300 px), asserting both halves: the PTY never takes the 1-row resize while
  collapsed, and afterwards the terminal drives to the new size with no pill.
  Mutation-checked in a real browser: reverting the `if (!newVp) return` guard
  fails it (`terminal refits after the shell settles` times out), and removing
  the viability guard fails it the other way (`Received: 1` rows).
- unit `decideViewportResize › a forgotten viewport turns the next real
  measurement passive` — documents *why* a refusal must not be committed.

**F2 (MEDIUM) — pull-forward is now rate-limited** to one per `maxDelayMs`
(`lastWakePullAt`, cleared on `start`/`retry`/a stable open, so a user's tap and
a credited success both restore the privilege). Measured in the reviewer's
harness (connects/60 s, jitter ×1.0): no wakes 11, wakes @5 s 15, wakes @1 s
**16** (was 62). New test pins ≤1.5× the unwoken rate under a 1 Hz wake stream;
mutants "no rate limit" and "no pull at all" each fail a test.

**F3 (LOW) — DST is now asserted on purpose:** a TZ-pinned case
(`America/New_York`, armed 2026-03-07 22:00 EST) checks the first arm is 2 h and
the re-arm across the spring-forward is **23 h**, not 24. The `+24h` mutant fails
it.

Re-verified: vitest 46 files / **880 passed**, `tsc --noEmit`, Biome lint on the
touched files, `./scripts/build.sh`, and Playwright on an isolated daemon
(unique `TMPDIR`): `terminal-resize`, `terminal-reattach-geometry`,
`initial-checkpoint-geometry`, `terminal-claim-retry`, `reconnect-stability` —
19/19 pass.

PR #528, opened 2026-09-08
fix(gmuxd,build): patch-release polish — bounded scan guard, diagnostic dev version

mgabor3141:
Two small daemon/scripts fixes for v2.1.1, plus the disposition of a third
reported item that does not reproduce on `main`.

## 1. Bound the conversation scan guard to in-flight work
`conversations.Index` guards a `Scan`'s late commit with a per-`(adapter, ref)`
removal generation (#517's zombie guarantee). The map was only ever written:
every `Remove`/`RemoveByRef` — manual `rm`, log rotation, a tool deleting a
conversation — left an entry behind for the daemon's lifetime.

It is now a *window* rather than a ledger: created when a `Scan` of that ref
starts, carrying the removals seen while that scan runs, dropped when the last
in-flight scan settles (including the describe-failure path). The staleness
guarantee is unchanged, because that window is exactly when a removal has
something to invalidate — a removal with no scan pending needs no record, since
any scan starting afterwards re-reads the world.

Tests: prune-after-settle, prune-after-invalidating-an-in-flight-scan,
concurrent scans share one entry, describe failures leak nothing.

## 2. Stamp source builds with a diagnostic dev version
`scripts/build.sh` defaulted `VERSION=dev` and `scripts/install.sh` never set
it, so a locally installed gmux/gmuxd answered `dev` to `--version` and
`/v1/health` — the one question those surfaces get asked on a source install.

Now: `dev+<short hash>` (`-dirty` on the git path), falling back to plain `dev`
outside a checkout. The jj lookup runs `--ignore-working-copy`, so a build stays
read-only: no repo lock, no snapshot. `VERSION` is exported *before* the vite
build so bundle and daemon agree — otherwise `version-watch` would think the tab
is permanently stale.

GoReleaser's real-version path is untouched (it never runs this script), and the
health/version wire surface is unchanged: still a free-form string.

`dev+…` is still a development build, so the two policies keyed off `"dev"` now
recognise the prefix: `gmux` does not start replacing a healthy daemon because a
rebuild moved the hash, and the update checker stays offline and silent.

## 3. Reported "reparent visibility leak" — not present on `main`
The report was that `case "reparent"` in `serve_central.go` computes
`visibleSession(...)` without consulting `ok`, letting a reparent event reach a
client that cannot see the session. Verified against `main`:

- `case "reparent"` emits no event at all; it validates the body and calls
  `Coordinator.SetSessionParent`, which returns `ErrSessionNotFound` → 404 for a
  session that is not in the store.
- Every per-session emission on `/v1/events` is already filtered for the
  subscriber's class: `snapshot.sessions` encodes through
  `SessionsPayload.FilterOwned` (`sessionEncodeMemo.payloadForLocked`), and
  `session-activity` is gated by `shouldForwardActivity`. There is no third
  per-session event.
- The one `ok` consumer in that switch is `scrollback`, and the reparent
  contract is deliberately snapshot-blind (ADR 0026 §2a): the existing
  `TestSessionReparentHTTPValidation` drives the handler with an *empty* fanout
  and requires 200. Gating reparent on `visibleSession` would regress that.

No code change; recorded here so the item is closed with its evidence.

## Verification
- `go test -race` — `services/gmuxd/internal/conversations`,
  `services/gmuxd/internal/update`, `services/gmuxd/cmd/gmuxd`,
  `cli/gmux/cmd/gmux`: all pass
- Mutation checks: never-pruning, pruning before the guard check, and a removal
  that does not invalidate an in-flight scan each fail the new/existing tests;
  narrowing `isDevVersion`/`IsDevVersion` back to `== "dev"` fails both the
  table test and the daemon-replacement test
- `./scripts/build.sh` (stamped `dev+<hash>`, bundle and daemon agree),
  `VERSION=v9.9.9` override still wins, VCS-less and git-only fallbacks probed
- Playwright on an isolated daemon: `no-external-network`, `harness`,
  `conversation-discovery`
- Drift guard: `TestPiSubcommandsMatchHelp` and
  `TestPiCommandsWithExtensionInjected` against the installed `pi`

### DIFF
--- a/cli/gmux/cmd/gmux/daemon.go
+++ b/cli/gmux/cmd/gmux/daemon.go
@@ -16,6 +16,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/gmuxapp/gmux/packages/buildversion"
 	"github.com/gmuxapp/gmux/packages/paths"
 	"github.com/gmuxapp/gmux/packages/sessionenv"
 )
@@ -91,9 +92,11 @@ func gmuxdNeedsStartContext(ctx context.Context) bool {
 		return true
 	}
 
-	// "dev" builds never replace — avoids churn during development and needs
-	// no identity response once socket ownership is established.
-	if version == "dev" {
+	// Development builds never replace — avoids churn during development and
+	// needs no identity response once socket ownership is established. A source
+	// install stamps "dev+<hash>", which is still a development build: rebuilding
+	// must not start replacing the running daemon on every hash change.
+	if buildversion.IsDev(version) {
 		return false
 	}
 
--- a/cli/gmux/cmd/gmux/main.go
+++ b/cli/gmux/cmd/gmux/main.go
@@ -16,6 +16,8 @@ import (
 // Falls back to "dev" for local builds.
 var version = "dev"
 
+
+
 func main() {
 	if len(os.Args) > 2 && os.Args[1] == "__detached-target" {
 		os.Exit(runDetachedTarget(os.Args[2:]))
--- a/cli/gmux/cmd/gmux/main_test.go
+++ b/cli/gmux/cmd/gmux/main_test.go
@@ -286,3 +286,21 @@ func TestParseHealthField(t *testing.T) {
 		t.Errorf("nonexistent = %q, want empty", got)
 	}
 }
+
+// A source install stamps "dev+<hash>" (see scripts/build.sh). It is still a
+// development build: rebuilding must not make every gmux invocation replace
+// the running daemon because the hash moved.
+func TestGmuxdNeedsStart_StampedDevNeverReplaces(t *testing.T) {
+	old := version
+	version = "dev+1a2b3c4d5e6f-dirty"
+	defer func() { version = old }()
+
+	stateDir, cleanup := startTestSocketDaemon(t, "0.4.3")
+	defer cleanup()
+	t.Setenv("XDG_STATE_HOME", stateDir)
+
+	if gmuxdNeedsStart() {
+		t.Error("a stamped dev build must not replace a healthy daemon")
+	}
+}
+
--- a/go.work
+++ b/go.work
@@ -5,6 +5,7 @@ toolchain go1.26.1
 use (
 	./cli/gmux
 	./packages/adapter
+	./packages/buildversion
 	./packages/paths
 	./packages/scrollback
 	./packages/socklease
--- a/packages/buildversion/buildversion.go
+++ b/packages/buildversion/buildversion.go
@@ -1,0 +1,34 @@
+// Package buildversion answers one question about the version string that
+// -ldflags stamps into gmux and gmuxd: did this binary come from a working
+// tree, or from a release?
+//
+// It lives in a shared package because the answer drives policy in three
+// separate places — the CLI's daemon autostart, gmuxd's incumbent/takeover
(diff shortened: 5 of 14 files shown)


mgabor3141:
Amended for the review (rebased onto `main` @ `de1904f9`, head now `35d3d090a1b4`).

**F1 (MEDIUM) — fixed with one shared predicate.** New tiny module
`packages/buildversion` (`IsDev`, `SameBuildClass`) is now the single source of
truth for the three policies that key off "is this a release?": the CLI's
autostart replacement, the update checker, and — the missed one — gmuxd's own
`implicitIncumbentCheck`/`takeover`, which now use `sameDaemonBuild(...)`
instead of `ident.Version == version`. Two dev stamps are one version, so a bare
`gmuxd run` yields again.
Tests: `TestImplicitIncumbentCheckTreatsDevStampsAsOneVersion` (7 cases: both
stamp directions, stamped-vs-unstamped, `-dirty`, same release, and the two
cases that must *still* replace — real version difference, release-vs-dev),
plus the `buildversion` tables. Mutants: exact equality → 3 subtests fail;
always-same → the two replace-expected subtests fail; `HasPrefix(v,"dev")` →
table fails. Reviewer's two-binary repro re-run against the amendment:
`dev+bbbb…` against a healthy `dev+aaaa…` incumbent now prints
`already running: healthy daemon dev+aaaaaaaaaaaa (pid …)` and exits 0 with the
incumbent alive, while a `v9.9.9` build still replaces it.

**F2 (LOW/MED) — git first, jj marked approximate.** `dev_version()` now tries
git when the tree has a `.git` of *its own* (so a grove workspace cannot answer
with the parent repo's HEAD) — that path is read-only *and* dirty-aware. The
jj-only fallback keeps `--ignore-working-copy` (no lock, no snapshot) but no
longer claims precision it lacks: its stamps end in `~` and the comment says
plainly that they name the tree as of the last jj command and can never say
`-dirty`. Probed: colocated clean → `dev+<hash>`, dirty → `dev+<hash>-dirty`,
jj-only workspace → `dev+<hash>~`, no VCS → `dev`.

**F3 (LOW) — removal-only test added.** `TestRemovalsWithoutScansRecordNothing`
(500 `RemoveByRef`/`Remove` of distinct refs, no scans) kills exactly the mutant
the reviewer found surviving: reinstating the ledger write in
`noteRemovalLocked` now fails with `removal-only workload left N guard entr(ies)`.

**F4 (LOW) — fixed rather than documented.** A full build records the bundle's
stamp in `bin/.web-version` (outside the `all:web` embed dir), and a
`--skip-frontend` build reuses it, so a daemon-only rebuild no longer disagrees
with the bundle it serves and no tab is forced to reload. Verified:
full build → `dev+…`, then `--skip-frontend` → the same string.

**F5 (INFO)** — noted, out of scope here; it is a property of `FilterOwned` and
becomes actionable with #525's peer-forwarding work.

Re-verified: `go test -race` for `cmd/gmuxd` (64 s), `internal/update`,
`cli/gmux/...`, `packages/buildversion` (`-count=3`), and
`internal/conversations` **`-race -count=3`**; drift guard
(`TestPiSubcommandsMatchHelp`, `TestPiCommandsWithExtensionInjected`);
`./scripts/build.sh`; Playwright `no-external-network`, `harness`,
`conversation-discovery` on an isolated daemon.

PR #535, opened 2026-09-16
proto(3.0): roots-only world state, paginated live children, deltas

mgabor3141:
**Prototype — not for merge.** Stacked on #534 (`spike/wire-deltas`). Full write-up with protocol surface, rejected designs, measurement tables, cuts, risks and open questions: `.memory/report-proto-world-split.md` (local, gitignored — summary below).

## What it does

Bounds the world state by **shape**, not pruning:

1. **World state = sidebar = roots only.** Each root carries additive `descendant_counts {total, alive, unread, error, waiting, active, running, children}`; unread on children becomes a count on the root; user-started shells (and their editors) stay roots. Requested with `view=roots` on `GET /v1/events`.
2. **`GET /v1/sessions/{id}/children?cursor=&limit=`** — newest-first, stable keyset cursor `(created_at,id)`, forwarded to the owning peer for `id@peer` and re-namespaced; 2.1 spokes (404) fall back to the hub's own projection. Plus **`GET /v1/sessions/{id}?ancestors=1`**.
3. **Children stay live while viewed** via client-registered scopes on the one SSE connection (`POST /v1/events/scopes {conn, add, remove}` → `snapshot.scope.delta` / `.reset`), emitted at the *same epoch* as the world delta so counts and drawer never disagree. Drawer closed ⇒ zero bytes for the subtree.
4. **Incremental world-state updates** via the R5 ring generalised to demanded view classes (browser|peer × full|roots) + refcounted scopes; `since=&boot=` resume on the roots frame. Hub↔spoke link asks `view=roots&delta=1`; 2.1 spokes degrade to full rows + hub-side counts (exercised live against a 2.1.1 spoke).

## Measured on the live corpus (2,850 rows; disposable daemon on a DB copy; real Chromium tab)

| | before | after |
|---|---|---|
| world-state bootstrap | 3,059,127 B (464 KB gz) | **90,229 B (17 KB gz)** |
| subagent spawn under DRIVER, drawer closed | 3.2 MB full tx / 8,475 B first cut | **879 B** (+6.4 KB world frame when `local_alive` changes) |
| same, drawer open | | + 762 B scope delta → drawer row appears |
| reconnect after 8 h (250 mutations) | 3.2 MB | 10,139 B |
| page DRIVER's 113 direct children @50 | | 3 pages, 93 KB |
| page club-3090@gmux-hs's 690 direct children @50 | | 14 pages, 888 KB (via hub fallback) |
| daemon CPU / broadcast | 0.7 ms (today) | 3.0 ms roots class; 15.6 ms full-view deltas (spike) |

Playwright **117/117** (unique TMPDIR); `peer-promote` adapted (hub list no longer carries a 3.0 spoke's children — removed semantic, not a bug). Web vitest 910 ✓, tsc ✓, Go touched packages ✓.

## Found & fixed while measuring

- A child spawn renumbered `project_index` on every root after it (flat FD-1 index counts children): 8.5 KB → 879 B by re-ranking over roots in the roots view (sort key only).
- Roots connections re-sent an unchanged 6.4 KB `snapshot.world` on every `sessions[]` change → skipped when byte-identical.
- Every drawer open fetched page 1 twice: scope chain start was `f.epoch` (bumped by Subscribe) instead of the baselined state's epoch.
- Per-epoch row-hash memo across views (hub + 3.0 spoke: 2× → +0.3 ms). The second marshal (wire encode) is left; needs `sessionstream.Encode` to take pre-encoded rows.

## Rejected for real-time (details in the report)

Polling `/children?since=` (no atomicity with counts, N requests per mutation); one SSE per subtree (connection cap, two epoch chains); inferring scope from `client-state` (drawer ≠ selection).

## Open questions

Scopes-on-one-connection as the 3.0 design? `project_index` roots-only at FD-1 vs re-ranked in the view? Split health counts out of `snapshot.world`? Relay scopes to 3.0 spokes or keep peer drawers pull-only? `gmux ls` on a hub with a 3.0 spoke?


mgabor3141:
## Round 2 — after the design review ("sound with changes")

Head `9ef0d7e9`, four commits on top of `90302eae`. Still a draft prototype; not to merge.

**Structural**
- **D1 atomic epochs**: scope payloads now ride *inside* `snapshot.sessions.delta` (`scopes:{key:{from_epoch,total,upsert,remove}|{reset}}`, sorted keys); `snapshot.scope.*` events are gone. Client applies world rows + every scope payload in one signals `batch` (unit test observes `1/1 → 2/2`, never a split). Verified at paint level in a real tab with MutationObserver + rAF across 3 runs: one mutation batch, one frame, zero refetches. That probe also caught a client false-reset (scope `from_epoch` = subscribe epoch > page epoch) — fixed by reporting the baseline state epoch.
- **D2 viewport scopes**: `add:[{scope, cursor}]` names the pages held; the ring diffs `Window(listing, cursor, 200)`. 690-child burst with page 1 held: 50 rows / 16.5 KB / fits (was 628 rows → reset per broadcast). Reparent-into-a-held-range is delivered by the window (M2). `total` is the daemon's (M1).
- Ring: scope changesets have their own id budget and base (M5 — `TestScopeChurnDoesNotEvictTheWorldChain`); 64 scopes/connection, release on close/disconnect (M3); scope rows/index/hashes memoized per epoch; allocation-free reflective row fingerprint replaces `json.Marshal` (oracle-tested).
- M6: `as=peer` roots computed against the audience's own family index.

**Defects**: H1 (reload held pages, dehydrate what's gone), M1, M3 (generation counter, release after close), L1 (idempotent annotate), L3 (no empty deltas), L4, L5, L7 (one re-registration POST). `gmuxd run --no-peers` / `GMUXD_NO_PEERS=1`.

**Re-measured** (2,851-row corpus): roots class 2.7 ms; club-3090 scope 7.7 → **3.4 ms**; **121 viewports (every parent) × 8 connections 8.8 ms** (review: 150×4 = 14.6 ms; full-view was 15.6 ms). Connections add ~0.1 ms. Remaining cost is O(rows covered by open windows); the end state needs a composer change hint or store revision (report §R2.3).

Gates: tsc ✓, vitest 916 ✓, Go `-race` ✓, Playwright 117/117 ✓. Full write-up: `.memory/report-proto-world-split.md` §Round 2.

